PR #3, opened 2024-12-04
Add quotes around DXC path

damyanp:


llvm-beanz:
I want to look at this a bit, because I feel like this shouldn't be needed.

llvm-beanz (approved):

PR #4, opened 2024-12-04
Fix setting root descriptor table

damyanp:


damyanp:
I haven't tried the new test on Vulkan or Metal.

llvm-beanz (approved):

PR #10, opened 2024-12-06
Fix setting root descriptor table

llvm-beanz:


damyanp:
What was the fix?  And any ideas on how to get github to show the diffs between this and the original change?

llvm-beanz:
> What was the fix? And any ideas on how to get github to show the diffs between this and the original change?

I haven't actually fixed it... I just created this PR to make sure all the actions runners grabbed it and failed...

damyanp:
> I haven't actually fixed it... I just created this PR to make sure all the actions runners grabbed it and failed...

Well, that explains why I couldn't find any differences!!

damyanp:
If the Vulkan implementation has a similar issue to the DX one, then I'd propose just removing the concept of descriptor sets entirely for now.

llvm-beanz:
The issues on Vulkan were all related to the test case, I've fixed the test case in https://github.com/llvm-beanz/HLSLTest/pull/10/commits/0ad46db5c4412dd0eb74a917a90ff2d8c7849778.

I'm not sure the abstraction I have in the pipeline is really the best abstraction, but the idea was to basically have it match the way descriptor tables (in Vulkan descriptor sets) are laid out. The only added complication is in generating the root signature which adds an additional level of indirection.

PR #19, opened 2024-12-28
Enable passing clang builders for prs

llvm-beanz:
This PR adds the currently passing clang build configurations to the PR checks, and marks the StructuredBuffer test as unsupoorted on Clang because it is currently crashing Clang.


llvm-beanz:
@bogner, I merged #17 because it passed the pre-merge checks, but it didn't occur to me that the pre-merge checks didn't include clang builds. It looks like clang is currently crashing compiling that shader, so this PR marks it as unsupported until we can get a fix into llvm/main.

bogner (approved):

bogner:
Makes sense to me. This needs a few patches from my llvm “next” branch yet to pass.

PR #26, opened 2025-01-21
Support for SRVs

bogner:
This is mostly cargo-culting the logic from UAVs and adjusting for the simpler read-only case of SRVs.

Note that I don't have a setup locally for Vulkan right now, so I've just XFAIL'd the test in Vulkan configs for now.


llvm-beanz (approved):
This looks right to me. I'll work on the VK implementation this afternoon if I have time.

PR #34, opened 2025-03-01
Add WSL support for D3D12 tests (excluding WAVE)

Icohedron:
This PR adds Windows Subsystem for Linux (WSL) support to run the D3D12 offload tests for both clang and dxc (excluding WAVE tests).

Additional prerequisites to use the offload-test-suite with WSL: 
- A Linux operating system running on WSL2 for access to the  `libd3d12.so` and `libdxcore.so` libraries located under `/usr/lib/wsl/lib/`
- The `libDirectX-Guids.so` and `libd3dx12-format-properties.so` libraries from [`DirectX-Headers`](https://github.com/microsoft/DirectX-Headers)

This PR introduces the following changes:
- Modifies CMakeLists files and adds `cmake/modules/FindD3D12_WSL.cmake` to find and link the required libraries and header files for using D3D12 in WSL
  - Two of the required libraries (d3d12 and dxcore) are included in WSL2 distributions under `/usr/lib/wsl/lib/`
  - Additionally requires two libraries built from DirectX-Headers (DirectX-Guids and d3dx12-format-properties) 
  - In addition to the usual D3D12 headers, the headers from `DirectX-Headers/include/wsl` are also required
- Edits `lib/API/DX/Device.cpp` to use libraries from the Windows SDK that are also supported by WSL
    - Use `Microsoft::WRL::ComPtr` from `<wrl/client.h>` instead of `CComPtr` from `<atlbase.h>`
    - Use eventfd file descriptors in place of the Windows event API when not compiling for Windows
    - The DXCore API is used to acquire devices instead of DXGI
    - The debug layer is excluded from WSL, since it doesn't appear to work in WSL
- Edits `lib/API/Device.cpp`, `include/API/Device.h`, `tools/api-query/api-query.cpp`, and `tools/offloader/offloader.cpp` to implement a workaround for a WSL-specific issue that is causing `api-query` and `offloader` to stall instead of exiting gracefully when multiple DirectX-capable devices are available 

Potential areas for improvement:
- Find a way to build the required WSL libraries from the existing DirectX-Headers in `third-party/DirectX-Headers` so that the libraries don't need to externally be provided (e.g., from a system package install of DirectX-Headers)
  - Adding `add_subdirectory(third-party/DirectX-Headers)` to the top-level CMakeLists.txt and then attempting to use the `DirectX-Guid` and `DirectX-Headers` library targets from it results in CMake errors along the lines of "export called with target 'OffloadTestAPI' which requires target ['DirectX-Guids'/'DirectX-Headers'] that is not in any export set."

Known Issues and fixes/workarounds:
- The D3D12 Debug Layer does not appear to work within WSL. (`D3D12GetDebugInterface` causes a segfault when called.)
  - Good luck debugging!
- PSOs can fail to be created due to unsigned shaders
  - Fix: Use a newer version of DXC that includes the validator binary `dxv`
- `double free or corruption (!prev)` may occur when running `api-query` and/or `offloader`, which follows with an infinite stall
  - A workaround is implemented which avoids this issue on normal program exit (i.e., `return 0;` from main)
  - This issue appears to occur whenever `DeviceContext::Devices` from `lib/API/Device.cpp` contains more than one device. 
    When `api-query` or `offloader` exit, some memory resources are failing to automatically free themselves. 
    Valgrind output suggests the core issue stems from the implementation of `Microsoft::WRL::ComPtr::InternalRelease` in WSL. 
    Manually clearing `DeviceContext::Devices` before a program exit appears to eliminate the issue for the exit
  - An additional workaround that would cover all possible program exit points is to only select one `DXDevice` by editing `lib/API/DX/Device.cpp` to `break;` at the end of the first iteration of the for-loop in the function `InitializeDXDevices()`
    - Before implementing this workaround, you can run `api-query` to list all GPU devices available. Afterwards, edit the `AdapterIndex` of the for-loop to select the index of the desired GPU
 
This PR was tested on a machine with the following (relevant) specifications and software versions:
- Operating System: Windows 11 24H2 
- `wsl --version` output:
```
WSL version: 2.3.26.0
Kernel version: 5.15.167.4-1
WSLg version: 1.0.65
MSRDC version: 1.2.5620
Direct3D version: 1.611.1-81528511
DXCore version: 10.0.26100.1-240331-1435.ge-release
Windows version: 10.0.26100.3323
```
- WSL Distribution: Ubuntu 24.04.2 LTS
- GPU: Intel(R) Arc(TM) Graphics -- the iGPU of the Intel(R) Core(TM) Ultra 7 165H processor
- DXC version: [v1.8.2502](https://github.com/microsoft/DirectXShaderCompiler/releases/tag/v1.8.2502) (GitHub Release) 

Tests ran and succeeded: `check-hlsl-unit` `check-hlsl-clang-d3d12` `check-hlsl-d3d12`


llvm-beanz:
This is great, but it looks like it breaks all the plain Windows builds.

Can you also add a markdown file under a `docs` folder to describe the WSL machine setup? I need to write some other docs too.

Icohedron:
I set up a new Windows development machine to debug and push fixes for the plain Windows builds.
I also wrote `docs/WSL.md` to describe the requirements and setup for running the experimental runtime test suite for HLSL in WSL.

Upon testing the WSL setup on the new Windows development machine, I also discovered a double free or memory corruption issue that occurs when multiple DXDevices are available. This prompted me to document all known issues regarding WSL support in `docs/WSL.md` along with possible fixes and workarounds.

cmake/modules/FindD3D12_WSL.cmake
```diff
@@ -0,0 +1,29 @@
+ # Check if in WSL and if has DirectX driver and runtime
+if (EXISTS "/usr/lib/wsl/lib/")
```
llvm-beanz, line 2: Not indenting the if block is a bit confusing, but to avoid indenting the whole file you can invert the condition and have the block just call return() to stop processing the file.

lib/API/DX/Device.cpp
```diff
@@ -633,16 +649,33 @@ class DXDevice : public offloadtest::Device {
...
       if (auto Err = HR::toError(
-              IS.Fence->SetEventOnCompletion(CurrentCounter, IS.Event),
+              IS.Fence->SetEventOnCompletion(CurrentCounter, Event),
               "Failed to register end event."))
         return Err;
+
+#ifdef _WIN32
       WaitForSingleObject(IS.Event, INFINITE);
+#else // WSL
+      pollfd pfd;
```
llvm-beanz, line 670: Please make variable names match LLVM coding standards: https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly

lib/API/DX/Device.cpp
```diff
@@ -746,32 +779,43 @@ class DXDevice : public offloadtest::Device {
...
   }
-  for (unsigned AdapterIndex = 0;;) {
-    CComPtr<IDXGIAdapter1> Adapter;
 
-    HRESULT HR = Factory->EnumAdapters1(AdapterIndex++, &Adapter);
-
-    if (DXGI_ERROR_NOT_FOUND == HR)
-      return llvm::Error::success();
+  ComPtr<IDXCoreAdapterList> AdapterList;
+  GUID attributes[]{DXCORE_ADAPTER_ATTRIBUTE_D3D12_CORE_COMPUTE,
```
llvm-beanz, line 802: Here too: https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly

llvm-beanz (commented):
Thank you for this! This mostly looks good. I want to think a little bit about the uninitialization and event handling code, but those are probably worth evolving in subsequent PRs.

PR #35, opened 2025-03-01
Refactor defining buffers to use named buffers

llvm-beanz:
This change separates the definition of buffers of data from the resource bindings that connect them to shaders.

I believe this is a simplification that will make it easier to write tests because we'll have user-defined names that we can match on in the output.

### DIFF
--- a/include/Image/Image.h
+++ b/include/Image/Image.h
@@ -85,12 +85,12 @@ public:
     assert(Data.size() == size() && "Data size does not match properties");
   }
 
-  ImageRef(const Resource &R)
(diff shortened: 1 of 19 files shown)


llvm-beanz:
@damyanp & @bogner curious for your feedback on this. I think this makes the tests easier to write and maintain.

I think I'm still going to need to redo how the descriptor sets and bindings are expressed in the YAML, but I think this is a good first step. Thoughts?

damyanp (approved):
Separating them like this does seem to make the tests clearer and have more meaningful output.

bogner (approved):
This looks like a good direction

PR #38, opened 2025-03-05
Simplify cbuffer test

bogner:
These bindings match between SPIR-V and DirectX, we don't need special handling.


bogner:
Looks like we need some more work with how bindings are specified before we can clean this up.

PR #40, opened 2025-03-10
[DX] Use the root signature from a shader if present

llvm-beanz:
This updates the DX device code to inspect the shader for a precompiled root signature and use that instead of building a root signature from scratch.

Still missing support for root descriptors, root constants and lots of tests, but this is a good starting base.

### DIFF
--- a/test/M3/RootSignatures/DescriptorTables.test
+++ b/test/M3/RootSignatures/DescriptorTables.test
@@ -1,0 +1,67 @@
+#--- source.hlsl
+
+struct Input {
+  float4 A;
+  float4 B;
+};
+
+StructuredBuffer<Input> In : register(t2, space0);
+RWBuffer<float4> Out1 : register(u1, space4);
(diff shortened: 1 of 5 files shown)


test/M3/RootSignatures/DescriptorTables.test
bogner, line 1: Using an “M3” folder makes sense right now but it will lose meaning over time. I think we should just have a “Feature” folder here and correlate to milestones externally.

bogner (approved):
Looks pretty reasonable. Should we add a few tests that specify root signatures in different ways (like the root signature define flag to dxc)? These should all work since we’re inspecting the dxil container.

llvm-beanz:
> Should we add a few tests that specify root signatures in different ways (like the root signature define flag to dxc)?

I think this should come out of the root signatures workstream documenting and authoring tests. I want to make sure some basic tests are written to verify that it "works" but exhaustive testing shouldn't be all on one person 😄.

PR #41, opened 2025-03-10
Update README to reflect YAML changes in #35 and #36

bogner:


llvm-beanz (approved):
Yea... I need to write some docs too.

PR #44, opened 2025-03-18
[HLSL] Adding tests for branch/flatten attributes

joaosaffran:
Adding tests to document DXC and clang support of branch/flatten attributes


llvm-beanz (approved):

llvm-beanz:
@joaosaffran can you move these tests to a folder `Features/Attributes`?

test/Basic/IfBranchAttr.test
```diff
@@ -0,0 +1,59 @@
...
+#endif
+
+RWBuffer<int> In REGISTER(u0);
+RWBuffer<int> Out REGISTER(u1);
+
+
+[numthreads(8,1,1)]
+void main(uint3 TID : SV_GroupThreadID) {
+    [flatten] 
+    if (TID.x) {
```
inbelic, line 15: Are we trying to test that for `TID.x == 0` the `Out` is not negated?

IIUC, when `TID.x == 0` then `int X = In[0] = 0` and so we aren't actually checking that it wasn't modified since `-0 = 0`.

llvm-beanz: Good catch.

@joaosaffran, can you update the buffer values so that they are values that are more obvious?

inbelic: Just as an fyi, this doesn't seem to be addressed in the latest commit, at least not for the `if` tests.

PR #45, opened 2025-03-18
[RootSignature] Add frontend RootSignature testing

inbelic:
adds testing demonstrating the various behaviours of specifying a root signature through a function attribute in HLSL for non-sampler descriptor types

- `Defaults.test`: all default parameter values
- `Flags.test`: various root/descriptor flags for M3
- `NumberParameters.test`: integer parameter edge-cases
- `ManualDescriptors.test`: shows manual user specified `offset` and `numDescriptors` parameters and the edge-cases
- `ParameterInsensitivity.test`: shows that parameters have no specified order, as well as, the case-insensitivity of keywords and enums

Noted discrepancies with the [root signature specification](https://github.com/llvm/wg-hlsl/blob/main/proposals/0002-root-signature-in-clang.md) are tracked to be resolved here: https://github.com/llvm/wg-hlsl/issues/192. 

Resolves: https://github.com/llvm/llvm-project/issues/130826

### DIFF
--- a/test/M3/RootSignatures/Defaults.test
+++ b/test/M3/RootSignatures/Defaults.test
@@ -1,0 +1,93 @@
+#--- source.hlsl
+
+cbuffer RootConstants : register(b0) {
+  float4 C;
+};
+
+struct Input {
+  float4 A;
+  float4 B;
+};
+
+struct Output {
+  float4 A;
+};
+
+StructuredBuffer<Input> In : register(t0);
+RWStructuredBuffer<Output> Out1 : register(u1);
+RWStructuredBuffer<Output> Out2 : register(u2);
+
+// Root signature to sanity test default values optional values
+
+#define RootSig                                                                \
+  "RootConstants(num32BitConstants = 4, b0), "                                 \
+  "DescriptorTable( "                                                          \
+  "  SRV(t0), "                                                                \
+  "  UAV(u1) "                                                                 \
+  "), "                                                                        \
+  "UAV(u2) "
+
+[RootSignature(RootSig)]
+[numthreads(1,1,1)]
+void main(uint GI : SV_GroupIndex) {
+  Out1[GI].A = In[GI].A * In[GI].B * C;
+  Out2[GI].A = In[GI].A * In[GI].B * C * 2;
+}
+
+//--- pipeline.yaml
+---
+Shaders:
+  - Stage: Compute
+    Entry: main
+    DispatchSize: [1, 1, 1]
+RuntimeSettings:
+  DirectX:
+    RootParameters:
+      - Kind: Constant
+        Name: Root
+      - Kind: DescriptorTable
+      - Kind: RootDescriptor
+        Resource:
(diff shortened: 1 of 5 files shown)


llvm-beanz:
@inbelic the new M3/RootSignatures/Flags.test seems to be crashing the test runner. If you have time to debug please do, otherwise I'll put it on my queue.

joaosaffran (approved):
LGTM

test/M3/RootSignatures/Defaults.test
```diff
@@ -0,0 +1,93 @@
...
+
+struct Output {
+  float4 A;
+};
+
+StructuredBuffer<Input> In : register(t0);
+RWStructuredBuffer<Output> Out1 : register(u1);
+RWStructuredBuffer<Output> Out2 : register(u2);
+
+// Root signature to sanity test default values optional values
```
inbelic, line 20:
```suggestion
// Root signature to sanity test the default values that are given to optional
// parameters (`space = 0`, `offset = DESCRIPTOR_RANGE_OFFSET_APPEND`, etc)
```

test/M3/RootSignatures/NumberParameters.test
```diff
@@ -0,0 +1,99 @@
...
+  float4 A;
+};
+
+StructuredBuffer<Input> In : register(t4294967294);
+RWStructuredBuffer<Output> Out1 : register(u1, space4294967279);
+RWStructuredBuffer<Output> Out2 : register(u2);
+
+// Root signature to test edge-cases of specify number arguments:
+//  - Maximum valid register value (0xfffffffe = 4294967294)
+//  - Maximum valid register space value (0x7fffffff = 4294967279)
```
inbelic, line 22:
```suggestion
//  - Maximum valid register space value (0x7fffffff = 2147483647)
```

llvm-beanz (approved):

test/M3/RootSignatures/NumberParameters.test
```diff
@@ -0,0 +1,99 @@
...
+//  - Maximum valid num32BitConstants value
+//    (61 = 64 - # of used DWORDS for other params)
+//  - Using (un)signed integer parameter values
+
+
+#define RootSig                                                                \
+  "RootConstants(num32BitConstants = +61, b0), "                               \
+  "DescriptorTable( "                                                          \
+  "  SRV(t4294967294), "                                                       \
+  "  UAV(u1, space = 4294967279) "                                             \
```
bogner, line 32: Should we be concerned that the space specified here is much larger than the "maximum valid register space value" mentioned above? In any case, the comment and the test don't match about what we're doing here.

inbelic: Ugh sorry. It should have been `0xffffffef` as the max value. I did the great math of `0xfffffff0 - 1 = 0x7fffffff` and didn't put together that is what your comment was pointing out... So it is functionally correct

PR #48, opened 2025-03-21
[RootSignature] Add test to demonstrate `offset = append` behaviour

inbelic:
- as a way to document the correct usage and behaviour of specifying `OffsetInDescriptorsFromTableStart = D3D12_DESCRIPTOR_RANGE_OFFSET_APPEND` we add a test that explicitly sets this value in Device.cpp through default behaviour

- this matches the behaviour of not specifying an offset in a given root signature (`DescriptorTables.test`)

Resolves https://github.com/llvm/llvm-project/issues/132955

### DIFF
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -231,7 +231,6 @@ public:
 
     uint32_t RangeIdx = 0;
     for (const auto &D : P.Sets) {
-      uint32_t DescriptorIdx = 0;
       uint32_t StartRangeIdx = RangeIdx;
       for (const auto &R : D.Resources) {
         switch (getDXKind(R.Kind)) {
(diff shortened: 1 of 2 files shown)


lib/API/DX/Device.cpp
```diff
@@ -248,10 +247,11 @@ class DXDevice : public offloadtest::Device {
         Ranges.get()[RangeIdx].NumDescriptors = 1;
         Ranges.get()[RangeIdx].BaseShaderRegister = R.DXBinding.Register;
         Ranges.get()[RangeIdx].RegisterSpace = R.DXBinding.Space;
-        Ranges.get()[RangeIdx].OffsetInDescriptorsFromTableStart =
-            DescriptorIdx;
+        // D3D12_DESCRIPTOR_RANGE_OFFSET_APPEND = -1.
+        // TODO: include proper enum when it is merged to main
```
llvm-beanz, line 251: Seems like this should be fixed before merging.

test/M3/RootSignatures/OffsetAppend.test
llvm-beanz, line 1: This test doesn't actually seem to add anything over existing tests. Am I missing something?

inbelic: My understanding is that since all the other test-cases (in `RootSignature/`) define a root signature, the default root signature construction path above is no longer run in any of the tests.

inbelic:
From offline discussion, we would like to test that the root signatures bytes are set correctly using another method such as `obj2yaml`, rather than relying on this test that isn't so directly related to the code setting it (default behaviour).

PR #49, opened 2025-03-24
Remove 'Unsupported: Clang' from tests that pass with clang

bogner:
However, do mark tests that are unsupported for `Clang-Vulkan` as such.

### DIFF
--- a/test/M3/RootSignatures/DescriptorTables.test
+++ b/test/M3/RootSignatures/DescriptorTables.test
@@ -54,8 +54,6 @@ DescriptorSets:
 ...
 #--- end
 
-# UNSUPPORTED: Clang
-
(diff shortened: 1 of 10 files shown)


llvm-beanz (approved):

test/M3/RootSignatures/DescriptorTables.test
llvm-beanz, line 1: Oh... the RootSignature tests only "work" because I made the DX bindings so that they generate the same root signature. We probably shouldn't remove the unsupported marker here, and should instead make changes so that these fail with Clang since they aren't actually supported.

bogner: Removed those two from this change for now.

PR #51, opened 2025-03-26
Add simple test for implicit resource bindings

hekota:
Test for implicit bindings of resource arrays will need to be added later.


test/Basic/ImplicitBinding1.test
```diff
@@ -0,0 +1,58 @@
...
+      DirectXBinding:
+        Register: 2
+        Space: 0
+...
+#--- end
+
+# UNSUPPORTED: Clang
+# RUN: split-file %s %t
+# RUN: %if !Vulkan %{ %dxc_target -T cs_6_0 -Fo %t.o %t/source.hlsl %}
+# RUN: %if Vulkan %{ %dxc_target -T cs_6_0 -fspv-target-env=vulkan1.3 -fvk-use-scalar-layout -Fo %t.o %t/source.hlsl %}
```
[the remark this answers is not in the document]
hekota: They're not ignored: `dxc failed : SPIR-V CodeGen not available. Please recompile with -DENABLE_SPIRV_CODEGEN=ON.`

llvm-beanz (approved):

test/Feature/ImplicitBindings/simple-gaps.test
```diff
@@ -0,0 +1,75 @@
...
+
+cbuffer CB { // gets b0
+  int a;
+}
+
+// TEST GAPS:
+// Need to add tests for:
+// - arrays of resources (array descriptors are not yet supported by the offload test suite)
+// - resources in user-defined structs
+// - combination of the above
```
bogner, line 16: I think there was some confusion over my use of the word "gaps" as a suggestion for the test file name. I was suggesting that this test is testing gaps in implicit bindings (ie, we have explicit bindings for A and C, but an implicit binding in between them with B).

The tests that fill the gaps in testing that you mention here would go in their own files (arrays.test, in-structs.test, etc), and so a comment in this file about them doesn't really make a lot of sense. Better to file issues saying we should write those tests instead.

bogner (approved):

test/Feature/ImplicitBindings/simple-gaps.test
hekota: Got it :) In a way implicit bindings are always trying to fill gaps, even if the gap is 0 to UINT_MAX.

The tasks to add tests for arrays and structs are llvm/wg-hlsl#175 and llvm/wg-hlsl#205.

test/Feature/ImplicitBindings/simple-resources.test
```diff
@@ -0,0 +1,69 @@
...
+        Space: 0
+    - Name: CB
+      Kind: ConstantBuffer
+      DirectXBinding:
+        Register: 0
+        Space: 0
+...
+#--- end
+
+# UNSUPPORTED: Clang
```
llvm-beanz, line 61: Vulkan bindings are very different and don't have the same implicit behavior.

```suggestion
# UNSUPPORTED: Clang, Vulkan
```

llvm-beanz: I'm going to commit this suggestion to see if it resolves the failing bot, then we can merge this.

bogner: > Vulkan bindings are very different and don't have the same implicit behavior.

Whiile "It's expected that this test will fail on DXC under Vulkan" is true, I don't think this comment explains why very accurately. There are a couple of things here:
- DXC doesn't seem to strip unused explicit bindings when targetting Vulcan
- DXC doesn't fill gaps in bindings compatibly between DX and Vulkan (see also how it handles packoffset)

These are arguably both bugs in DXC, rather than different behaviours of Vulkan. SPIR-V can easily represent the bindings in a matching way to what we emit for DirectX. We can and should make the HLSL behaviour here match between the two targets going forward.

Because of all of this, I think we should probably XFAIL this on DXC-Vulkan rather than call it unsupported, and comment as to why.

llvm-beanz: My argument against using `XFAIL` here even though this is pretty clearly a bug is that `XFAIL` tests always run in case the result changes: we're never going to fix this in DXC, so the result will never change running it just slows down the tests.

I can make it more specific to DXC and add a comment because hopefully we'll do this correctly in Clang.

test/Feature/ImplicitBindings/simple-resources.test
```diff
@@ -0,0 +1,69 @@
...
+        Space: 0
+    - Name: CB
+      Kind: ConstantBuffer
+      DirectXBinding:
+        Register: 0
+        Space: 0
+...
+#--- end
+
+# UNSUPPORTED: Clang, Vulkan
```
llvm-beanz, line 61:
```suggestion
# UNSUPPORTED: Clang

# This is a bug in DXC's SPIRV that we're almost certainly never going to fix.
# I'm marking this as unsupported rather than xfail because it isn't worth
# running the test just to verify it is still failing.
# UNSUPPORTED: DXC-Vulkan
```

PR #65, opened 2025-04-04
Add support and tests for `ByteAddressBuffer` for DirectX

hekota:
Adds test and support for `ByteAddressBuffer` and `RWByteAddressBuffer` (DirectX only).

Fixes llvm/wg-hlsl#201

### DIFF
--- a/include/Support/Pipeline.h
+++ b/include/Support/Pipeline.h
@@ -43,8 +43,10 @@ enum class DataFormat {
 enum class ResourceKind {
   Buffer,
   StructuredBuffer,
+  ByteAddressBuffer,
   RWBuffer,
   RWStructuredBuffer,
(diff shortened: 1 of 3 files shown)


llvm-beanz (commented):
This looks great. Do you have a test to go with it? We can put `UNSUPPORTED: Vulkan, Metal` in the test to have it not run there until I get around to finishing out the support.

llvm-beanz:
> This looks great. Do you have a test to go with it? We can put `UNSUPPORTED: Vulkan, Metal` in the test to have it not run there until I get around to finishing out the support.

Oh, I now see the test in #66. Can we instead add the test with this PR?

inbelic (approved):
LGTM, my comments from other pr look addressed.

PR #66, opened 2025-04-04
Add test for `ByteAddressBuffer`

hekota:
Depends on #65

Closes llvm/wg-hlsl#201


inbelic (commented):
LGTM, although I agree that we could move this to the pr with the implementation changes as well.

Also, can we add duplicate tests with different bit-width types (uint16, and hex8)? That way we test all the different code paths in `getRawDXFormat`

hekota:
Merged into #65

PR #68, opened 2025-04-08
Use more consistent folder names

bogner:
- The `M3` folder was from before we settled on putting these tests under `Feature/`
- The `Features/` folder appears to be a typo


inbelic (approved):
Makes sense to me

spall (approved):
lgtm

PR #69, opened 2025-04-08
Introduce syntax to specify vulkan bindings

bogner:
This adds a "VKBinding" part for resource descriptors, similar to the already existing DirectXBinding. These are a bit different from DX in that we only specify a binding index - the binding can also have a descriptor set, but that's represented structurally.

There are a few details to be aware of here:

- DXC's behaviour with `register()` annotations is not to differentiate between u/t/b types, so I've adjusted quite a few tests to simply avoid bindings that would have overlap in vulkan.
- Clang does not support the `vk::binding` attribute or the `-vk-*-shift` flag so tricks like the above are the easiest way to make compatible shaders.
- For tests that were buggy on DXC, I've added new tests to cover the shared functionality
- I've defaulted the spirv dxc invocations to specify vulkan1.3 to simplify command lines

### DIFF
--- a/lib/Support/Pipeline.cpp
+++ b/lib/Support/Pipeline.cpp
@@ -124,12 +124,18 @@ void MappingTraits<offloadtest::Resource>::mapping(IO &I,
   I.mapRequired("Name", R.Name);
   I.mapRequired("Kind", R.Kind);
   I.mapRequired("DirectXBinding", R.DXBinding);
+  I.mapOptional("VulkanBinding", R.VKBinding);
 }
 
 void MappingTraits<offloadtest::DirectXBinding>::mapping(
     IO &I, offloadtest::DirectXBinding &B) {
   I.mapRequired("Register", B.Register);
   I.mapRequired("Space", B.Space);
+}
+
+void MappingTraits<offloadtest::VulkanBinding>::mapping(
+    IO &I, offloadtest::VulkanBinding &B) {
+  I.mapRequired("Binding", B.Binding);
 }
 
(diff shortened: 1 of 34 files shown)


lib/Support/Pipeline.cpp
```diff
@@ -123,6 +123,7 @@ void MappingTraits<offloadtest::Resource>::mapping(IO &I,
   I.mapRequired("Name", R.Name);
   I.mapRequired("Kind", R.Kind);
   I.mapRequired("DirectXBinding", R.DXBinding);
+  I.mapRequired("VulkanBinding", R.VKBinding);
```
llvm-beanz, line 126: I kinda feel like it was a mistake to make the DirectXBinding required, since we only need it if you don't have an embedded root signature.

Should we make the Vulkan binding required or optional? I think there are a few good arguments for optional, but curious about your thoughts?

bogner: I mostly made it required to match the DirectX binding, and don't have a strong opinion. It is odd that we need to specify this in places like the root signature tests where it isn't actually used. However, making it required definitely simplifies some things - if we don't have a vulkan binding, what should we do? Assuming the bindings increase in source order like this used to do tends to fall apart and lead to rather confusing error conditions where we write to the wrong buffers and such.

bogner: I've updated this to make the vulkan binding optional, but we'll error if we try to run on vulkan and it's missing. I think this is a good middle ground and we should consider something similar for the DX binding.

llvm-beanz (approved):

PR #75, opened 2025-04-14
Test for clamp intrinsic, tests float, int, and uint versions

spall:
Shows how to test multiple types in the same test file. I plan to use this intrinsic to prototype and test improvements to how the offload test suite verifies test results.

Due to this issue, https://github.com/llvm/llvm-project/issues/135453, there is a lot of code reuse instead of abstraction.


spall:
Closing this PR; will be implemented later

PR #77, opened 2025-04-21
Enhance offload test suite to compare test results with two new rules 'BufferExact' and 'BufferFuzzy'

### LINKED ISSUE #76: Enhance offload test suite to verify test results
Add support to the offload test suite to support checking test results in the test suite rather than with FileCheck.

Add support to specify Result(s) in the YAML like below:
```
Results:
- Result: 
   Rule: ExactCompare
   Actual: Out
   Expected: ExpectedOut
- Result:
   Rule: FuzzyCompare
   Epsilon: 2
   # Other extra fields required by rules would go here as well
   Actual: Out2
   Expected: ExpectedOut2
```

Implement rules, such as 'ExactCompare', necessary to verify the test results of HLSL intrinsics, which produce scalars and vectors of integer and floating point types.

spall:
Add support for testing results in the offload test suite.
Add two testing functions, 'BufferExact' and 'BufferFuzzy'
BufferExact tests if two buffers contain exactly the same content, using == on each element.
BufferFuzzy works on only floating point values and performs a ULP comparison on each floating point value in two buffers being compared.
The BufferFuzzy rule requires a ULPT field in the YAML which specifies the allowed ULP difference. It also takes an optional DenormMode field, which defaults to Any.

Add a test each for BufferExact and BufferFuzzy which show what error output looks like.
Closes #76


test/Unit/Check/BufferExact.test
llvm-beanz, line 1: The `Unit` folder in all the lit test suites is a special folder. It has no test suffixes (https://github.com/llvm-beanz/offload-test-suite/blob/main/test/Unit/lit.site.cfg.py.in#L18), it exists to configure the GoogleTest unit tests.

Can you instead create a `tools/offloader` folder under `test` and move these there?

lib/Support/Check.cpp
```diff
@@ -0,0 +1,123 @@
...
+  int64_t diff = *((const uint64_t *)&fsrc) - *((const uint64_t *)&fref);
+  uint64_t uDiff = diff < 0 ? -diff : diff;
+  return uDiff <= (uint64_t)ULPTolerance;
+}
+
+bool getResult(offloadtest::Result R) {
+  switch (R.Rule) {
+	  case offloadtest::Rule::BufferExact: {
+	    return testBufferExact(R.ActualPtr, R.ExpectedPtr);
+	  }
```
llvm-beanz, line 69: Looks like some tabs crept in...

Please clang-format.

lib/Support/Check.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+inline bool isdenorm(float f) {
+  return (std::numeric_limits<float>::denorm_min() <= f &&
+          f < std::numeric_limits<float>::min()) ||
+         (-std::numeric_limits<float>::min() < f &&
+          f <= -std::numeric_limits<float>::denorm_min());
+}
+
+bool compareFloatULP(
+    const float &fsrc, const float &fref, unsigned ULPTolerance,
```
llvm-beanz, line 22: Please follow LLVM conventions for naming of things across this file:
https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly

lib/Support/Check.cpp
```diff
@@ -0,0 +1,123 @@
...
+  unsigned int uDiff = diff < 0 ? -diff : diff;
+  return uDiff <= ULPTolerance;
+}
+
+bool compareDoubleULP(
+    const double &fsrc, const double &fref, unsigned ULPTolerance,
+    offloadtest::DenormMode DM) {
+  if (fsrc == fref) {
+    return true;
+  }
```
llvm-beanz, line 48:
```suggestion
  if (fsrc == fref)
    return true;
```

lib/Support/Check.cpp
```diff
@@ -0,0 +1,123 @@
...
+  if (std::isnan(fsrc)) {
+    return std::isnan(fref);
+  }
```
llvm-beanz, line 51:
```suggestion
  if (std::isnan(fsrc))
    return std::isnan(fref);
```

include/Support/Check.h
```diff
@@ -0,0 +1,25 @@
+//===- Check.h - Functions for checking test results-------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+//
```
inbelic, line 9: nit: no description given, can delete otherwise

lib/Support/Check.cpp
```diff
@@ -0,0 +1,81 @@
+//===- Check.cpp - Check test results  ------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
```
inbelic, line 8: nit: no description, can delete otherwise

lib/Support/Check.cpp
```diff
@@ -0,0 +1,81 @@
...
+    llvm::MutableArrayRef<float> Arr2(reinterpret_cast<float *>(B2->Data.get()),
+                                      B2->Size / sizeof(float));
+    for (unsigned I = 0; I < Arr1.size(); ++I) {
+      if (!compareFloatULP(Arr1[I], Arr2[I], ULPT, DM))
+        return false;
+    }
+    return true;
+  }
+  default:
+    llvm_unreachable("Only float types are supported by the fuzzy test.");
```
inbelic, line 78: nit: Maybe this is better as an assert.

I think we should only use unreachable if there are no code paths that reach here. But IIUC, there isn't anything that would prevent us from reaching here when specifying an int buffer

test/Tools/Offloader/BufferFuzzy.test
inbelic, line 1: Can we add tests that specify the `FTZ` or `Preserve` enum of `DenormMode`? Maybe for a value that would have returned true for line 30 of Check.cpp:
https://github.com/llvm-beanz/offload-test-suite/blob/7724cf2f056873c14fa5d5d7b6f2b06ccada7f68/lib/Support/Check.cpp#L27-L32

inbelic (commented):
LGTM. Just some small nits. Thanks for commenting the buffers, it makes it easier to review :)

include/Support/Check.h
spall: I was following the style of the other files in the project which also include this blank spot.

lib/Support/Check.cpp
spall: I added a check for the buffer type in Pipeline.cpp

test/Tools/Offloader/BufferFuzzy.test
spall: I added a test for this in a new file BufferFuzzy-error.test. I also added more positive tests to BufferFuzzy.test.

inbelic (approved):

lib/Support/Check.cpp
```diff
@@ -0,0 +1,81 @@
...
+//===----------------------------------------------------------------------===//
+
+#include "Support/Check.h"
+
+static bool isDenorm(float F) {
+  return (std::numeric_limits<float>::denorm_min() <= F &&
+          F < std::numeric_limits<float>::min()) ||
+         (-std::numeric_limits<float>::min() < F &&
+          F <= -std::numeric_limits<float>::denorm_min());
+}
```
bogner, line 19: Is this different from `!std::isnormal()`?

lib/Support/Check.cpp
bogner: > I think we should only use unreachable if there are no code paths that reach here.

FWIW we should also only assert on conditions that are always true - assert and unreachable should generally be used in functionally identical ways.

spall: > > I think we should only use unreachable if there are no code paths that reach here.
> 
> FWIW we should also only assert on conditions that are always true - assert and unreachable should generally be used in functionally identical ways.

This reminds me I have some asserts I should still verify.

lib/Support/Check.cpp
spall: Looks like yes, 0.0, inf, and NaN are not normal.
https://en.cppreference.com/w/cpp/numeric/math/isnormal

include/Support/Check.h
```diff
@@ -0,0 +1,20 @@
...
+//===----------------------------------------------------------------------===//
+
+#ifndef OFFLOADTEST_SUPPORT_CHECK_H
+#define OFFLOADTEST_SUPPORT_CHECK_H
+
+#include "Pipeline.h"
+
+bool getResult(offloadtest::Result R);
```
llvm-beanz, line 18: Can you add a doxygen comment describing what this does?

Since I think this is verifying the correctness of the result a better name might be something like `verifyResult`.

lib/Support/Check.cpp
```diff
@@ -0,0 +1,81 @@
...
+    }
+    return true;
+  }
+  default:
+    llvm_unreachable("Only float types are supported by the fuzzy test.");
+  }
+  return false;
+}
+
+bool getResult(offloadtest::Result R) {
```
llvm-beanz, line 74: Instead of returning bool, if these functions returned `llvm::Error` we could embed semantic information about the failure into the error. That could allow for more informative results. For example in the ULP compare you could capture the actual and expected values.

Also `llvm::Error` has this neat feature `joinError`, which allows you to build arrays of errors, so you can report back not just the first failure but a list of failures.

Even if we don't go to that full complexity in the initial implementation it is probably worth using `llvm::Error` to report success/failure just to have the ability to improve the diagnostics in the future as we extend this to support alternate verification rules.

test/Tools/Offloader/BufferFuzzy-error.test
```diff
@@ -0,0 +1,113 @@
...
+    Stride: 4
+    Data: [ 1.5, 2.5 ]
+  - Name: Out2
+    Format: Float32
+    Stride: 4
+    ZeroInitSize: 4
+  - Name: Expected2
+    Format: Float32
+    Stride: 4
+    Data: [ 0.0000000000000000000000000000000000000112 ] # isDenorm will return true for this value
```
llvm-beanz, line 36: For values like this where you really rely on an explicit representation it may be best to encode this as hex with the numeric value in a comment.

lib/Support/Check.cpp
spall: Ok I've replaced this function's usage of bool with llvm:Error.

include/Support/Check.h
spall: I updated the name and added a comment

lib/Support/Check.cpp
bogner: Looks like `std::fpclassify(F) == FP_SUBNORMAL` might be the right thing to do here then (See https://en.cppreference.com/w/cpp/numeric/math/fpclassify). Note that subnormal and denormal are equivalent in this context.

lib/Support/Check.cpp
```diff
@@ -0,0 +1,103 @@
...
+    if (testBufferExact(R.ActualPtr, R.ExpectedPtr))
+      return llvm::Error::success();
+    std::string str;
+    llvm::raw_string_ostream oss(str);
+    oss << "Test failed: " << R.Name << "\nExpected:\n";
+    llvm::yaml::Output Yoss(oss);
+    Yoss << *R.ExpectedPtr;
+    oss << "Got:\n";
+    Yoss << *R.ActualPtr;
+    return llvm::createStringError(str.c_str());
```
bogner, line 88: `llvm::SmallString` and `raw_svector_ostream` are slightly more efficient here (they avoid an allocation). Also LLVM naming conventions would us Str, OS, and probably YAMLOS here.

lib/Support/Check.cpp
```diff
@@ -0,0 +1,103 @@
...
+    if (testBufferFuzzy(R.ActualPtr, R.ExpectedPtr, R.ULPT, R.DM))
+      return llvm::Error::success();
+    std::string str;
+    llvm::raw_string_ostream oss(str);
+    oss << "Test failed: " << R.Name << "\nExpected:\n";
+    llvm::yaml::Output Yoss(oss);
+    Yoss << *R.ExpectedPtr;
+    oss << "Got:\n";
+    Yoss << *R.ActualPtr;
+    return llvm::createStringError(str.c_str());
```
bogner, line 100: Same as above

bogner (commented):
Generally looks good, a few minor comments

bogner (approved):

PR #80, opened 2025-04-23
Add support for UAV Counters

V-FEXrt:
This PR add support for HLSL code that updates the counter  on a UAV Resource and reading it back to check against.


lib/API/DX/Device.cpp
```diff
@@ -518,6 +528,8 @@ class DXDevice : public offloadtest::Device {
                                "Failed to acquire UAV data pointer."))
       return Err;
     memcpy(ResDataPtr, R.BufferPtr->Data.get(), R.size());
+    if (R.HasCounter)
+      memcpy((char*)ResDataPtr + getUAVBufferCounterOffset(R), &R.BufferPtr->Counter, sizeof(uint32_t));
```
V-FEXrt, line 532: This doesn't actually work right now. The value isn't filled in by the `memcpy` for some reason. I also tried using `memset`  with an arbitrary value which also didn't seem to help. 

I'm hoping that anyone reviewing can suggest ideas for resolution!

test/Feature/StructuredBuffer/counter.test
```diff
@@ -0,0 +1,49 @@
+#--- source.hlsl
+RWStructuredBuffer<int> Out : register(u0);
+
+[numthreads(1,1,1)]
+void main(uint GI : SV_GroupIndex) {
+  Out.IncrementCounter();
+  Out.IncrementCounter();
```
V-FEXrt, line 7: Add a `DecrementCounter` test

lib/API/DX/Device.cpp
V-FEXrt: Discussed offline with a few folks and this is WAI. The counter is automatically zeroed out so I should just delete this and the support from the yaml

test/Feature/StructuredBuffer/dec_counter.test
```diff
@@ -0,0 +1,48 @@
...
+
+# RUN: split-file %s %t
+# RUN: %dxc_target -T cs_6_0 -Fo %t.o %t/source.hlsl
+# RUN: %offloader %t/pipeline.yaml %t.o | FileCheck %s
+
+# CHECK: Creating UAV: { Size = 4100, Register = u0, Space = 0, HasCounter = 1 }
+# CHECK: UAV: HeapIdx = 0 EltSize = 4 NumElts = 1 HasCounter = 1
+
+# CHECK: Name: Out
+# CHECK: Counter: 4294967292
```
hekota, line 45: So this is checking that the DecrementBuffer underflows from 0? I wonder if there is some way to set the counter value to some non-zero initial value.

hekota (approved):
LGTM!

lib/API/DX/Device.cpp
```diff
@@ -450,8 +458,10 @@ class DXDevice : public offloadtest::Device {
   }
 
   llvm::Expected<ResourceSet> createUAV(Resource &R, InvocationState &IS) {
-    llvm::outs() << "Creating UAV: { Size = " << R.size() << ", Register = u"
+    uint32_t BufferSize = getUAVBufferSize(R);
```
alsepkow, line 461: nit: const

lib/API/DX/Device.cpp
```diff
@@ -89,6 +89,14 @@ static DXGI_FORMAT getRawDXFormat(Resource &R) {
   return DXGI_FORMAT_UNKNOWN;
 }
 
+static uint32_t getUAVBufferSize(Resource &R) {
```
alsepkow, line 92: >uint32_t [](http://example.com/codeflow?start=7&length=8)

nit: Use size_t instead?

lib/API/DX/Device.cpp
```diff
@@ -895,6 +907,8 @@ class DXDevice : public offloadtest::Device {
                                  "Failed to map result."))
         return Err;
       memcpy(R.first->BufferPtr->Data.get(), DataPtr, R.first->size());
+      if (R.first->HasCounter)
+        memcpy(&R.first->BufferPtr->Counter, (char*)DataPtr + getUAVBufferCounterOffset(*R.first), sizeof(uint32_t));
```
alsepkow, line 911: >(char*)DataPtr [](http://example.com/codeflow?start=45&length=14)

nit: use static_cast

test/Feature/StructuredBuffer/dec_counter.test
```diff
@@ -0,0 +1,48 @@
+#--- source.hlsl
+RWStructuredBuffer<int> Out : register(u0);
```
alsepkow, line 2: >int [](http://example.com/codeflow?start=19&length=3)

Should this be uint to match the output type of DecrementCounter?

V-FEXrt: Since we are testing the output bit pattern it doesn't really seem to matter but I went ahead and changed it

PR #87, opened 2025-04-28
Add the particle life use case test

bogner:
This is a shader that demonstrates that we can compile and run a non-trivial compute shader. It's a slightly modified version of [particle_life.hlsl](https://github.com/damyanp/dplife/blob/main/src/particle_life/particle_life.hlsl)

### DIFF
--- a/test/UseCase/particle-life.test
+++ b/test/UseCase/particle-life.test
@@ -119,9 +119,7 @@ void main(uint3 dispatch_thread_id : SV_DispatchThreadID) {
 
     float3 color =  particle_type_to_color(particle.type);
 
-    // TODO: keep original...
-    // color = lerp(color, color * 0.1f, 1-saturate(hit));
-    color = lerp(color, color * 0.1f, (float3)1-saturate(hit));
+    color = lerp(color, color * 0.1f, 1-saturate(hit));
 
     Vertices[particle_id].color = float_to_abgr(color);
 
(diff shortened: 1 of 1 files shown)


bogner:
I'm not 100% sure whether it makes sense to put larger use case tests like this here, but this demonstrates that particle life works via the offload suite. It would at least be a decent regression test, so adding it would have some value.

Note: need to get @damyanp to explicitly give permission to contribute this under the LLVM license if we want to merge this.

bogner:
Note: this PR is stacked on a branch based on #77 as it uses the fuzzy matching implemented there for the test.

damyanp:
> Note: need to get @damyanp to explicitly give permission to contribute this under the LLVM license if we want to merge this.

Permission granted.

test/UseCase/particle-life.test
```diff
@@ -0,0 +1,365 @@
...
+
+    particle.velocity = velocity;
+
+    Vertices[particle_id].position = particle.position;
+
+    float3 color =  particle_type_to_color(particle.type);
+
+    // TODO: keep original...
+    // color = lerp(color, color * 0.1f, 1-saturate(hit));
+    color = lerp(color, color * 0.1f, (float3)1-saturate(hit));
```
damyanp, line 124: ?

damyanp (approved):

PR #90, opened 2025-04-30
[NFC] Apply clang-tidy

llvm-beanz:
This PR configures clang-tidy to enforce a swath of rules that comply with and enforce LLVM's coding standards.

Two new CMake options are added:
* `OFFLOADTEST_USE_CLANG_TIDY` - When set to `On` this runs clang-tidy on non-third-party sources in the offload-test-suite as part of the build process.
* `OFFLOADTEST_CLANG_TIDY_APPLY_FIX` - When set to `On` passes `--fix` to `clang-tidy` to automatically apply fix-its for any issues that have fixes available (Requires `OFFLOADTEST_USE_CLANG_TIDY=On`).

The enabled warning set is:

`clang-diagnostic-*` - Clang compiler diagnostics.
`llvm-*` - LLVM coding standards
`misc-*` - All miscellaneous rules except:
* [misc-use-anonymous-namespace](https://clang.llvm.org/extra/clang-tidy/checks/misc/use-anonymous-namespace.html), which violates LLVM coding standards which [prefers `static` for functions](https://llvm.org/docs/CodingStandards.html#restrict-visibility).
* [misc-include-cleaner](https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html), which did some unholy things to a few source files and caused them to fail to compile.
*[misc-non-private-member-variables-in-classes](https://clang.llvm.org/extra/clang-tidy/checks/misc/non-private-member-variables-in-classes.html), which conflicts with common patterns in LLVM code.

`readability-identifier-naming` - Apply [LLVM naming conventions](https://llvm.org/docs/CodingStandards.html#name-types-functions-variables-and-enumerators-properly)

Additionally this PR adds the `QualifierAlignment: Left` option to the `.clang-format` configuration because the clang-tidy [misc-const-correctness](https://clang.llvm.org/extra/clang-tidy/checks/misc/const-correctness.html) rule adds const to the right of the type which is less common in every codebase ever (Thank's Clang!).

### DIFF
--- a/lib/API/MTL/MTLDevice.cpp
+++ b/lib/API/MTL/MTLDevice.cpp
@@ -21,7 +21,7 @@ using namespace offloadtest;
 static llvm::Error toError(NS::Error *Err) {
   if (!Err)
     return llvm::Error::success();
-  std::error_code EC =
+  std::error_code const EC =
       std::error_code(static_cast<int>(Err->code()), std::system_category());
   llvm::SmallString<256> ErrMsg;
   llvm::raw_svector_ostream OS(ErrMsg);
@@ -58,9 +58,9 @@ class MTLDevice : public offloadtest::Device {
   struct InvocationState {
     InvocationState() { Pool = NS::AutoreleasePool::alloc()->init(); }
     ~InvocationState() {
-      for (auto T : Textures)
+      for (auto *T : Textures)
         T->release();
-      for (auto B : Buffers)
+      for (auto *B : Buffers)
         B->release();
       if (Fn)
         Fn->release();
@@ -86,12 +86,12 @@ class MTLDevice : public offloadtest::Device {
 
   llvm::Error loadShaders(InvocationState &IS, const Shader &P) {
     NS::Error *Error = nullptr;
-    llvm::StringRef Program = P.Shader->getBuffer();
-    dispatch_data_t data = dispatch_data_create(Program.data(), Program.size(),
+    llvm::StringRef const Program = P.Shader->getBuffer();
+    dispatch_data_t Data = dispatch_data_create(Program.data(), Program.size(),
                                                 dispatch_get_main_queue(),
                                                 ^{
                                                 });
-    IS.Lib = Device->newLibrary(data, &Error);
+    IS.Lib = Device->newLibrary(Data, &Error);
     if (Error)
       return toError(Error);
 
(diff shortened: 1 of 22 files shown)


lib/Image/Color.cpp
```diff
@@ -53,19 +54,19 @@ static double convertXYZ(double Val) {
...
-  double L = fmax(0.0, 116.0 * Y - 16.0);
-  double A = 500 * (X - Y);
-  double B = 200 * (Y - Z);
+  const double L = fmax(0.0, 116.0 * Y - 16.0);
+  const double A = 500 * (X - Y);
+  const double B = 200 * (Y - Z);
   return Color(L, A, B, ColorSpace::LAB);
 }
 
 static double convertLAB(double Val) {
```
damyanp, line 68: Why didn't this get changed to `convertLab`?

damyanp (approved):

spall (approved):

lib/API/MTL/MTLDevice.cpp
```diff
@@ -58,9 +58,9 @@ class MTLDevice : public offloadtest::Device {
   struct InvocationState {
     InvocationState() { Pool = NS::AutoreleasePool::alloc()->init(); }
     ~InvocationState() {
-      for (auto T : Textures)
+      for (auto *T : Textures)
         T->release();
-      for (auto B : Buffers)
+      for (auto *B : Buffers)
```
bogner, line 63: Better to spell out `MTL::Texture *` and `MTL::Buffer *` here rather than use auto

lib/Image/Color.cpp
```diff
@@ -31,15 +32,15 @@ static Color clampColor(const Color C) {
                std::clamp(C.B, 0.0, 1.0), C.Space);
 }
 
-static Color RGBToXYZ(const Color Old) {
+static Color rgbToXyz(const Color Old) {
```
bogner, line 35: Not sure how clang-tidy decides which things are initialisms vs words, but I'd expect RGB, XYZ, and LAB to stay capitalized. Does it appease the tool if we call these convertRGBToXYZ? If not this is probably fine, but it feels weirdly inconsistent to me.

tools/offloader/offloader.cpp
```diff
@@ -105,10 +105,12 @@ int run() {
       APIToUse = GPUAPI::DirectX;
       outs() << "Using DirectX API\n";
     } else if (*reinterpret_cast<const uint32_t *>(
-                   PipelineDesc.Shaders[0].Shader->getBuffer().data()) == 0x07230203) {
+                   PipelineDesc.Shaders[0].Shader->getBuffer().data()) ==
+               0x07230203) {
```
bogner, line 109: Really? Might be worth a variable for `PipelineDesc.Shaders[0].Shader->getBuffer()` to clean up this whole if-else chain anyway.

bogner (approved):
LG. A few minor suggestions.

lib/Image/Color.cpp
llvm-beanz: The camelCase rules don't perform dictionary lookups of words, so it doesn't know that "lab" is a word, which is great because in this case it is actually the L*a*b color space 😝 .

PR #180, opened 2025-05-05
Add Float16 as a Scalar type option

### LINKED ISSUE #79: Add Half as a scalar type
Add Half to DataFormat. It looks like Half can be represented using uint16_t.
Will be needed for future intrinsic tests.
Reference: https://github.com/llvm-beanz/offload-test-suite/pull/74

kmpeng:
Add Float16 to DataFormat Enum, represented with a uint16_t.
This will be needed for future intrinsic tests.

Closes #79.

### DIFF
--- a/lib/Support/Pipeline.cpp
+++ b/lib/Support/Pipeline.cpp
@@ -127,10 +127,10 @@ void MappingTraits<offloadtest::Buffer>::mapping(IO &I,
     DATA_CASE(Int16, int16_t)
     DATA_CASE(Int32, int32_t)
     DATA_CASE(Int64, int64_t)
+    DATA_CASE(Float16, uint16_t)
     DATA_CASE(Float32, float)
     DATA_CASE(Float64, double)
     DATA_CASE(Bool, uint32_t) // Because sizeof(bool) is 1 but HLSL represents a bool using 4 bytes.
-    DATA_CASE(Half, uint16_t)
   }
 
   I.mapOptional("OutputProps", B.OutputProps);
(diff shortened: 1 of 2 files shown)


kmpeng:
Wondering if people would want this to be called Float16 instead of Half. Please comment if you have thoughts!

spall (commented):
Looks good! I realize seeing this now that we should probably call it Float16 instead of Half, and then, where appropriate, put Float16 above Float32.

spall (approved):

bogner (approved):
Makes sense to me. I think any other options for the storage type would be less portable, and uint16_t makes it clear we just care about the bit pattern.

lib/Support/Pipeline.cpp
```diff
@@ -127,6 +127,7 @@ void MappingTraits<offloadtest::Buffer>::mapping(IO &I,
     DATA_CASE(Int16, int16_t)
     DATA_CASE(Int32, int32_t)
     DATA_CASE(Int64, int64_t)
+    DATA_CASE(Float16, uint16_t)
```
llvm-beanz, line 130: Do we want this to be unit16_t or hex16?

Seems like hex16 is probably more useful.

spall: The reference code from DXC uses uint16_t to represent half, if that matters.
https://github.com/microsoft/DirectXShaderCompiler/blob/624665f3987d379a299aa14dbc53e0cd3b96afea/include/dxc/Test/HlslTestUtils.h#L567

PR #183, opened 2025-05-12
Add ULP comparison support for Float16

### LINKED ISSUE #81: Add ULP comparison support for Half
After support is added for Half (See issue #79), and 'BufferFuzzy' rule is added as part of PR #77, add support to do ULP comparison for Half values.
A reference for this can be taken from: https://github.com/microsoft/DirectXShaderCompiler/blob/624665f3987d379a299aa14dbc53e0cd3b96afea/include/dxc/Test/HlslTestUtils.h#L567

AC: Add a function to do ULP Comparison for Half. Add Tests to show it testing valid and invalid cases.

kmpeng:
- Add the `compareFloat16ULP` function to do ULP Comparison for Float16 ([reference](https://github.com/microsoft/DirectXShaderCompiler/blob/624665f3987d379a299aa14dbc53e0cd3b96afea/include/dxc/Test/HlslTestUtils.h#L567))
- Update `BufferFuzzy` to support Float16 and call the new function to compare Float16 values
- Add tests to show `BufferFuzzy` testing valid and invalid cases

Closes #81.

### DIFF
--- a/lib/Support/Check.cpp
+++ b/lib/Support/Check.cpp
@@ -15,6 +15,10 @@
 
 static bool isDenorm(float F) { return std::fpclassify(F) == FP_SUBNORMAL; }
 
+static bool isnanFloat16(uint16_t val) {
+  return (val & 0x7c00) == 0x7c00 && (val & 0x03ff) != 0;
+}
+
 static bool compareFloatULP(const float &FSrc, const float &FRef,
                             unsigned ULPTolerance, offloadtest::DenormMode DM) {
   if (FSrc == FRef)
@@ -30,6 +34,18 @@ static bool compareFloatULP(const float &FSrc, const float &FRef,
   // For FTZ or Preserve mode, we should get the expected number within
   // ULPTolerance for any operations.
   int Diff = *((const uint32_t *)&FSrc) - *((const uint32_t *)&FRef);
+  unsigned int AbsDiff = Diff < 0 ? -Diff : Diff;
+  return AbsDiff <= ULPTolerance;
+}
+
+static bool compareHalfULP(const uint16_t &FSrc, const uint16_t &FRef,
+                           unsigned ULPTolerance) {
+  if (FSrc == FRef)
+    return true;
+  if (isnanFloat16(FSrc))
+    return isnanFloat16(FRef);
+  // 16-bit floating point numbers must preserve denorms
+  int Diff = FSrc - FRef;
   unsigned int AbsDiff = Diff < 0 ? -Diff : Diff;
   return AbsDiff <= ULPTolerance;
 }
@@ -58,6 +74,23 @@ static bool testBufferFuzzy(offloadtest::Buffer *B1, offloadtest::Buffer *B2,
                                       B2->Size / sizeof(float));
     for (unsigned I = 0; I < Arr1.size(); ++I) {
       if (!compareFloatULP(Arr1[I], Arr2[I], ULPT, DM))
+        return false;
+    }
+    return true;
+  }
+  case offloadtest::DataFormat::Float16: {
+    if (B1->Size != B2->Size)
+      return false;
+    llvm::MutableArrayRef<uint16_t> Arr1(
+        reinterpret_cast<uint16_t *>(B1->Data.get()),
+        B1->Size / sizeof(uint16_t));
+    assert(B2->Format == offloadtest::DataFormat::Float16 &&
+           "Buffer types must be the same");
+    llvm::MutableArrayRef<uint16_t> Arr2(
+        reinterpret_cast<uint16_t *>(B2->Data.get()),
+        B2->Size / sizeof(uint16_t));
+    for (unsigned I = 0; I < Arr1.size(); ++I) {
+      if (!compareHalfULP(Arr1[I], Arr2[I], ULPT))
         return false;
     }
     return true;
--- a/lib/Support/Check.cpp
+++ b/lib/Support/Check.cpp
@@ -38,8 +38,8 @@ static bool compareFloatULP(const float &FSrc, const float &FRef,
   return AbsDiff <= ULPTolerance;
 }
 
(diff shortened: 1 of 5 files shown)


lib/Support/Check.cpp
```diff
@@ -62,6 +78,23 @@ static bool testBufferFuzzy(offloadtest::Buffer *B1, offloadtest::Buffer *B2,
...
+      return false;
+    llvm::MutableArrayRef<uint16_t> Arr1(
+        reinterpret_cast<uint16_t *>(B1->Data.get()),
+        B1->Size / sizeof(uint16_t));
+    assert(B2->Format == offloadtest::DataFormat::Float16 &&
+           "Buffer types must be the same");
+    llvm::MutableArrayRef<uint16_t> Arr2(
+        reinterpret_cast<uint16_t *>(B2->Data.get()),
+        B2->Size / sizeof(uint16_t));
+    for (unsigned I = 0; I < Arr1.size(); ++I) {
```
bogner, line 92: It really doesn't matter much here, but it's a good habit to get in to avoiding calling a function to recalculate the end of a loop on every iteration.
```suggestion
    for (unsigned I = 0, E = Arr1.size(); I < E; ++I) {
```

See https://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop for a longer rationale for this.

test/Tools/Offloader/BufferFuzzy-error-16bit.test
```diff
@@ -0,0 +1,118 @@
...
+# CHECK: Height:          0
+# CHECK: Width:           0
+# CHECK: Depth:           0
+# CHECK:  ...
+# CHECK: Got:
+# CHECK: ---
+# CHECK: Name:            Out1
+# CHECK: Format:          Float16
+# CHECK: Stride:          2
+# CHECK: Data:            [ 19731, 17664 ]
```
bogner, line 91: It's a bit confusing that we print these as integers. I think this lends creedence to the [suggestion on the previous PR](https://github.com/llvm/offload-test-suite/pull/180#discussion_r2075869916) that we should represent halfs using `llvm::yaml::Hex16` rather than `uint16_t`. This way we'll get a hex value printed and it's obvious that the bit pattern is what we're interested in.

lib/Support/Check.cpp
```diff
@@ -34,6 +38,18 @@ static bool compareFloatULP(const float &FSrc, const float &FRef,
   return AbsDiff <= ULPTolerance;
 }
 
+static bool compareFloat16ULP(const uint16_t &FSrc, const uint16_t &FRef,
+                              unsigned ULPTolerance) {
+  if (FSrc == FRef)
+    return true;
+  if (isnanFloat16(FSrc))
+    return isnanFloat16(FRef);
```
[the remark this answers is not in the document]
kmpeng: Hm yeah. I can change both this and `compareFloatULP` to check the inverse as well.

Wonder what the reason behind only checking it this way in the DXC implementation was. I guess in practice the function would almost always return false when `FRef` is NAN and `FSrc` isn't because the diff would likely result in a very large value? And it would very likely exceed ULPTolerance?

lib/Support/Pipeline.cpp
```diff
@@ -42,8 +42,10 @@ void MappingTraits<offloadtest::Pipeline>::mapping(IO &I,
       if (!R.ExpectedPtr)
         I.setError(Twine("Reference buffer ") + R.Expected + " not found!");
       if (R.Rule == offloadtest::Rule::BufferFuzzy) {
-        if (R.ActualPtr->Format != offloadtest::DataFormat::Float32 ||
-            R.ExpectedPtr->Format != offloadtest::DataFormat::Float32)
+        if ((R.ActualPtr->Format != offloadtest::DataFormat::Float32 ||
```
spall, line 45: I think it would be good to write a function like 'isFloatingPointFormat' which checks if the format is Float16 or Float32. This would be useful when we need to add Float64 to this check.

test/Tools/Offloader/BufferFuzzy-error-16bit.test
spall: I wonder if we can easily modify these to print hex instead of decimal format.

bogner: Isn't that what @llvm-beanz's suggestion in [PR 180](https://github.com/llvm/offload-test-suite/pull/180#discussion_r2075869916) would do? I haven't tried it but I think this is just a matter of doing the following as he suggested to do in post-commit review:
```diff
diff --git a/lib/Support/Pipeline.cpp b/lib/Support/Pipeline.cpp
index 27eb3c0..5c539e0 100644
--- a/lib/Support/Pipeline.cpp
+++ b/lib/Support/Pipeline.cpp
@@ -127,7 +127,7 @@ void MappingTraits<offloadtest::Buffer>::mapping(IO &I,
     DATA_CASE(Int16, int16_t)
     DATA_CASE(Int32, int32_t)
     DATA_CASE(Int64, int64_t)
-    DATA_CASE(Float16, uint16_t)
+    DATA_CASE(Float16, llvm::yaml::Hex16)
     DATA_CASE(Float32, float)
     DATA_CASE(Float64, double)
     DATA_CASE(Bool, uint32_t) // Because sizeof(bool) is 1 but HLSL represents a bool using 4 bytes.
```

spall: I wasn't sure if that was all that change would do, or if it would cause differences elsewhere.

kmpeng: Changed it to `llvm::yaml::Hex16` for now - doesn't seem like it will cause differences elsewhere to me

bogner (approved):

spall (approved):

PR #185, opened 2025-05-13
Enable most of the vulkan tests under clang

bogner:
This removes `UNSUPPORTED: Clang-Vulkan` from most of then tests that had it. These tests pass with top-of-tree LLVM.

There are a couple of places we need to keep unsupported tests for now:
- Tests that use -fvk-use-scalar-layout flag
- Tests that use the vk::binding attribute

### DIFF
--- a/test/Basic/TestFloat32Pipeline.test
+++ b/test/Basic/TestFloat32Pipeline.test
@@ -39,7 +39,6 @@ DescriptorSets:
 ...
 #--- end
 
-# UNSUPPORTED: Clang-Vulkan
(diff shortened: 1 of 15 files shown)


llvm-beanz:
Looks like 3 tests are still failing:

```
Failed Tests (3):
  OffloadTest-clang-vk :: Basic/TestFloat32Pipeline.test
  OffloadTest-clang-vk :: Basic/simple.test
  OffloadTest-clang-vk :: Feature/StructuredBuffer/srv.test
```

I can check these out locally to see what's going on.

PR #188, opened 2025-05-14
Update workflows for LLVM organization

llvm-beanz:
This updates the workflows to work with the builders moved under the LLVM organization. It also removes the NV nightly builder since that bot isn't migrating to LLVM.


bogner (approved):

.github/workflows/windows-nv-clang-d3d12.yaml
```diff
@@ -6,16 +6,16 @@ permissions:
 
 on:
   workflow_dispatch:
-  schedule:
-    - cron: '0 0-12 * * 1-5' # run every 30 minutes overnight
-    - cron: '0 * * * 0,6' # run every 30 minutes on weekends
+#  schedule:
+#    - cron: '0 0-12 * * 1-5' # run every 30 minutes overnight
+#    - cron: '0 * * * 0,6' # run every 30 minutes on weekends
```
V-FEXrt, line 11: nit: Can we just delete this? Can use git history to bring it back if we end up needing it

.github/workflows/windows-nv-clang-vk.yaml
```diff
@@ -6,16 +6,16 @@ permissions:
 
 on:
   workflow_dispatch:
-  schedule:
-    - cron: '0 0-12 * * 1-5' # run every 30 minutes overnight
-    - cron: '0 * * * 0,6' # run every 30 minutes on weekends
+#  schedule:
```
V-FEXrt, line 9: Same here and for the rest

V-FEXrt (approved):

llvm-beanz:
Everything that was passing before the transfer is passing with this PR. So we're good.

Issue #189, opened 2025-05-15
Support Vulkan version from command line arguments


llvm-beanz:
The code currently configures Vulkan 1.1 devices because you need to setup the API version before you can query for devices so I just picked the lowest.

An incoming PR will change this to setup a 1.1 then query the device and recreate the instance and device for whatever is supported. This works around issues where Vulkan 1.3 is currently required by clang shaders. We should instead take this value from the command line.

Issue #186, opened 2025-05-13
Don't force Vulkan 1.3


llvm-beanz:
We are currently [forcing Vulkan 1.3](https://github.com/llvm-beanz/offload-test-suite/blob/main/test/lit.cfg.py#L56), which isn't ideal since we should really be able to write tests against any version of Vulkan.

For now the non-1.3 support in LLVM is pretty limited, so this might be the right call temporarily, but we should figure out a better long term solution.

PR #190, opened 2025-05-15
[Vulkan] Use the API version from the device

llvm-beanz:
Initially this code just hard-coded Vulkan 1.1 because you need to set it before you can query device capabilities (which is messy). This changes the code to setup a 1.1 device, query it, then recreate the device with the highest supported version (generally 1.3).

We should update this code to accept a vulkan version from command line arguments (see #189).

### DIFF
--- a/lib/API/VK/Device.cpp
+++ b/lib/API/VK/Device.cpp
@@ -170,6 +170,8 @@ public:
     }
   }
 
+  const VkPhysicalDeviceProperties &getProps() const { return Props; }
+
 private:
   void queryCapabilities() {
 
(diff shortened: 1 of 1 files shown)


bogner (approved):

bogner:
Should we / can we set the `-fspv-target-env` flag in lit based on this same logic? See https://github.com/llvm/offload-test-suite/blob/main/test/lit.cfg.py#L56

llvm-beanz:
> Should we / can we set the `-fspv-target-env` flag in lit based on this same logic? See https://github.com/llvm/offload-test-suite/blob/main/test/lit.cfg.py#L56

I filed a separate issue for that the other day (https://github.com/llvm/offload-test-suite/issues/186). I think we ideally don't want LIT to force that value on, it should be test-by-test, and we probably want some mechanism to run tests with multiple values to ensure compatibility. We would also want that support for shader models, but I haven't really figured out the right mechanism for that either.

spall (approved):

PR #200, opened 2025-05-20
Add offload-test-suite support for `ldexp`

### LINKED ISSUE #95: Add test for ldexp

A .test file should contain 
```
@@ -0,0 +1,54 @@
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# UNSUPPORTED: Clang-Vulkan
# RUN: split-file %s %t
```

kmpeng:
Closes #95.

- Test for Half type (`ldexp.fp16.test`)
- Test for 32 bit types (`ldexp.32.test`)


alsepkow:
Should 'ldexp.32.test' be something like 'ldexp.fp32.test' instead?

test/Feature/HLSLLib/ldexp.32.test
```diff
@@ -0,0 +1,62 @@
+#--- source.hlsl
+
+StructuredBuffer<float4> In : register(t0);
```
alsepkow, line 3: Is there a reason we use float4 instead of float?

spall: This way each <n x float> version can easily be tested. If it was float, a float4, etc, would need to be constructed.
Line 8 tests the float4 version of ldexp, like 10 tests float3 etc

test/Feature/HLSLLib/ldexp.32.test
```diff
@@ -0,0 +1,62 @@
...
+---
+Shaders:
+  - Stage: Compute
+    Entry: main
+    DispatchSize: [1, 1, 1]
+Buffers:
+  - Name: In
+    Format: Float32
+    Stride: 16
+    Data: [ 3.14159, 0, -inf, NaN, 1, 1.5, -0.5, inf ]
```
spall, line 25: The HLSL docs on this function are light on details, but referencing the c++ docs: https://en.cppreference.com/w/cpp/numeric/math/ldexp, I think it would be good to test overflow and underflow situations as well. 

Looking at the special cases it seems the only one missing, besides overflow and underflow, is where the first argument (num) is -0.

test/Feature/HLSLLib/ldexp.32.test
```diff
@@ -0,0 +1,62 @@
...
+StructuredBuffer<float4> In : register(t0);
+RWStructuredBuffer<float4> Out : register(u1);
+
+[numthreads(1,1,1)]
+void main() {
+  Out[0] = ldexp(In[0], In[1]);
+  Out[1].x = ldexp(In[1].x, In[1].y);
+  Out[1].yzw = ldexp(In[1].yzw, In[0].yzw);
+  Out[2].xy = ldexp(In[0].xy, In[1].yw);
+}
```
spall, line 12: I think it might be helpful to add a comment showing what value combinations are being tested; I believe this is correct (but the formatting is messed up)
```
// Num       Exp
// 3.14159  1
// 0             1.5
// -inf          -0.5
// Nan         inf
// 3.14159   0
// 1.5           0
// -.05         -inf
// inf           Nan
// 3.14159  1.5
// 0             inf
```

test/Feature/HLSLLib/ldexp.fp16.test
```diff
@@ -0,0 +1,64 @@
+#--- source.hlsl
+
+StructuredBuffer<half4> In : register(t0);
+RWStructuredBuffer<half4> Out : register(u1);
+
+[numthreads(1,1,1)]
+void main() {
```
spall, line 7: same comments on this test as the 32 bit one.

test/Feature/HLSLLib/ldexp.fp16.test
```diff
@@ -0,0 +1,64 @@
...
+---
+Shaders:
+  - Stage: Compute
+    Entry: main
+    DispatchSize: [1, 1, 1]
+Buffers:
+  - Name: In
+    Format: Float16
+    Stride: 8
+    Data: [ 0x4248, 0x0000, 0xFC00, 0x7E00, 0x3C00, 0x3E00, 0xB800, 0x7C00 ]
```
spall, line 25: comment what these values are in decimal form

test/Feature/HLSLLib/ldexp.fp16.test
```diff
@@ -0,0 +1,64 @@
...
+  - Name: Out
+    Format: Float16
+    Stride: 8
+    ZeroInitSize: 24
+  - Name: ExpectedOut # The result we expect
+    Format: Float16
+    Stride: 8
+    Data: [ 0x4648, 0x0000, 0xFC00, 0x7E00, 0x41A8, 0x3E00, 0x8000, 0x7E00, 0x4871, 0x7E00, 0x0000, 0x0000 ]
```
spall, line 33: comment what these are in decimal form

kmpeng:
> Should 'ldexp.32.test' be something like 'ldexp.fp32.test' instead?

I was also wondering this and think it would make more sense if it were `ldexp.fp32.test`. I was following the issue description that said to name the test for 32 bits `ldexp.32.test` - didn't know if it was written that way for a reason. @spall Any thoughts?

test/Feature/HLSLLib/ldexp.32.test
kmpeng: Added this in the main function but didn't know if that made sense. Lmk if you think there's a better place to move it

spall:
> > Should 'ldexp.32.test' be something like 'ldexp.fp32.test' instead?
> 
> I was also wondering this and think it would make more sense if it were `ldexp.fp32.test`. I was following the issue description that said to name the test for 32 bits `ldexp.32.test` - didn't know if it was written that way for a reason. @spall Any thoughts?

Since not all GPUs support all bit sizes, or all types, the idea was to separate the types we need to check for gpu capability for in separate files.  So a function which supports half and int16 would have both a fp16.test and int16.test file because those would be separate capabilities we need to check for.  For 32 bit types, they should be supported by all GPUs, so they would go into one file 32.test.  Since ldexp operates on only half and float I think the better change would be to name the 16 bit test file ldexp.16.test, to make it clear there isn't a missing int16.test file.

PR #203, opened 2025-05-22
Implement ULP Comparison for doubles

### LINKED ISSUE #82: Add ULP Comparison support for Double
After  'BufferFuzzy' rule is added as part of PR https://github.com/llvm-beanz/offload-test-suite/pull/77, add support to do ULP comparison for Double values.

AC: Add a function to do ULP Comparison for Double. Update 'BufferFuzzy' to support Double and call the new function to compare Double values. Add Tests to show it testing valid and invalid cases.

spall:
Implement ULP comparison support for doubles.
Abstracts BufferFuzzy code to be less repetitive. 
Adds Success and Failure tests for BufferFuzzy with Float64.
Closes #82

### DIFF
--- a/lib/Support/Check.cpp
+++ b/lib/Support/Check.cpp
@@ -13,15 +13,17 @@
 #include "llvm/Support/Error.h"
 #include "llvm/Support/raw_ostream.h"
 
-template<typename T>
-static bool isDenorm(T F) { return std::fpclassify(F) == FP_SUBNORMAL; }
+template <typename T> static bool isDenorm(T F) {
+  return std::fpclassify(F) == FP_SUBNORMAL;
+}
 
 static bool isFloat16NAN(uint16_t Val) {
   return (Val & 0x7c00) == 0x7c00 && (Val & 0x03ff) != 0;
 }
 
 static bool compareDoubleULP(const double &FSrc, const double &FRef,
-			      unsigned ULPTolerance, offloadtest::DenormMode DM) {
+                             unsigned ULPTolerance,
+                             offloadtest::DenormMode DM) {
   if (FSrc == FRef)
     return true;
   if (std::isnan(FSrc) || std::isnan(FRef))
@@ -40,7 +42,7 @@ static bool compareDoubleULP(const double &FSrc, const double &FRef,
 }
 
 static bool compareFloatULP(const float &FSrc, const float &FRef,
-			     unsigned ULPTolerance, offloadtest::DenormMode DM) {
+                            unsigned ULPTolerance, offloadtest::DenormMode DM) {
   if (FSrc == FRef)
     return true;
   if (std::isnan(FSrc) || std::isnan(FRef))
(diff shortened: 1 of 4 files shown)


spall:
BufferFuzzy-64bit.test fails on clang due to https://github.com/llvm/llvm-project/issues/140998

lib/Support/Check.cpp
```diff
@@ -13,12 +13,34 @@
...
+template <typename T> static bool isDenorm(T F) {
+  return std::fpclassify(F) == FP_SUBNORMAL;
+}
 
 static bool isFloat16NAN(uint16_t Val) {
   return (Val & 0x7c00) == 0x7c00 && (Val & 0x03ff) != 0;
 }
 
+static bool compareDoubleULP(const double &FSrc, const double &FRef,
+                             unsigned ULPTolerance,
```
Icohedron, line 25: Should `ULPTolerance` be a `uint64_t`?
Its use is to be compared to `AbsDiff` which is a `const uint64_t`

Icohedron (approved):
Overall, LGTM

test/Tools/Offloader/BufferFuzzy-error-64bit.test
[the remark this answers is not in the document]
spall: We could add a test for that; I think it was left out because that failure case should hit the same code as any two values which are not the same and are not within ulp range.

lib/Support/Pipeline.cpp
```diff
@@ -49,6 +50,8 @@ void MappingTraits<offloadtest::Pipeline>::mapping(IO &I,
         if (!isFloatingPointFormat(R.ActualPtr->Format) ||
             !isFloatingPointFormat(R.ExpectedPtr->Format))
           I.setError(Twine("BufferFuzzy only accepts Float buffers"));
+        if (R.ActualPtr->Format != R.ExpectedPtr->Format)
+          I.setError(Twine("Buffers must have the same type"));
```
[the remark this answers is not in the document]
spall: Yeah I can add that

lib/Support/Check.cpp
```diff
@@ -60,40 +82,54 @@ static bool testBufferExact(offloadtest::Buffer *B1, offloadtest::Buffer *B2) {
   return true;
 }
 
+template <typename T>
+static bool testAll(std::function<bool(const T &, const T &)> Fn,
```
alsepkow, line 86: Should we give this parameter a more descriptive name like "ComparisonFn" instead of "Fn" ?

lib/Support/Check.cpp
```diff
@@ -60,40 +82,54 @@ static bool testBufferExact(offloadtest::Buffer *B1, offloadtest::Buffer *B2) {
...
+                    llvm::ArrayRef<T> Arr1, llvm::ArrayRef<T> Arr2) {
+  if (Arr1.size() != Arr2.size())
+    return false;
+
+  for (unsigned I = 0, E = Arr1.size(); I < E; ++I) {
```
alsepkow, line 91: I think 'I' should be a size_t for consistency with the return type of llvm::ArrayRef.size()? Although I doubt it would ever cause a real issue.

PR #207, opened 2025-05-27
Update ULP comparison support for Float16 to reflect changes in DXC

### LINKED ISSUE #202: Update `compareFloat16ULP` to align with DXC's implementation
ULP comparison support for Float16 was recently added based on the existing pattern in DXC's `compareHalfULP` function (see issue #81). There will likely be changes to the DXC implementation via PR [#7393](https://github.com/microsoft/DirectXShaderCompiler/pull/7393), which we should also update in our `compareFloat16ULP` implementation to maintain consistency with DXC.

Changes:
1. Handling +0 and -0 as equal
2. ULP differences are calculated with a monotonic ordering

spall:
Adopt the changes to 'compareHalfULP' from this PR: https://github.com/microsoft/DirectXShaderCompiler/pull/7393/
Closes #202

### DIFF
--- a/lib/Support/Check.cpp
+++ b/lib/Support/Check.cpp
@@ -14,6 +14,8 @@
 #include "llvm/Support/raw_ostream.h"
 #include <cmath>
 
+#define FLOAT16_BIT_SIGN 0x8000
+
 // limited to float, double, and long double
 template <typename T> static bool isDenorm(T F) {
   return std::fpclassify(F) == FP_SUBNORMAL;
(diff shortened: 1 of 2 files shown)


kmpeng (approved):
LGTM

lib/Support/Check.cpp
```diff
@@ -14,6 +14,8 @@
 #include "llvm/Support/raw_ostream.h"
 #include <cmath>
 
+#define FLOAT16_BIT_SIGN 0x8000
```
llvm-beanz, line 17: Can we instead make this a `constexpr uint16_t` value?

spall: sure i can define it in the function too

kmpeng: The `0x7c00` and `0x03ff` in `isFloat16NAN` are also a `#define` in DXC (`FLOAT16_BIT_EXP` & `FLOAT16_BIT_MANTISSA`), but when I wrote it I just put the values straight in. Might want to change the ones in `isFloat16NAN` to whatever you decide as well for consistency?

llvm-beanz (approved):

PR #219, opened 2025-06-03
[VK] Fix crash on exit when using VVL

Keenuts:
VVL has internal state which conflicts requires a specific cleanup
ordering. Because the VKContext relied on the global dtor, this order
was not respected.

There are 2 solutions:
 - use atexit
 - require an explicit call to vkDestroyDevice.

Thing is, users of `Device` except the object to be usable as a simple
singleton, with no explicit `destroy` to call.
For this reasons, it seems using `atexit` is the cleanest option from
the user perspective.

### DIFF
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -62,7 +62,8 @@ public:
 #endif
 
 #ifdef OFFLOADTEST_ENABLE_VULKAN
-  static llvm::Error initializeVXDevices();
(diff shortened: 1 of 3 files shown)


Keenuts:
Thanks! Btw, I don't have write access to this repo, so you'll have to merge the 2 PRs

llvm-beanz (approved):

PR #222, opened 2025-06-04
Add all of the 1.1 - 1.4 vulkan features to api-query

bogner:
This defines all of the features enums for vulkan 1.1 through 1.4, and wires them up so that api-query can list which ones are available.


lib/API/VK/Device.cpp
```diff
@@ -206,6 +213,14 @@ class VKDevice : public offloadtest::Device {
 #define VULKAN12_FEATURE_BOOL(Name)                                            \
   Caps.insert(                                                                 \
       std::make_pair(#Name, make_capability<bool>(#Name, Features12.Name)));
+#include "VKFeatures.def"
```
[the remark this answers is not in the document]
bogner: It is, in fact, included 4 times. This is typically how we deal with .def files from tablegen in LLVM, where we've defined a number of things in the same file and depending on the `#define`s we do work with various parts of them.

However, in this case we've written this file by hand so we don't really have tablegen's constraints, and it's probably more sensible to just define all of the macros and include this once. I'll update it.

lib/API/VK/Device.cpp
```diff
@@ -181,9 +181,16 @@ class VKDevice : public offloadtest::Device {
     Features11.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_1_FEATURES;
     VkPhysicalDeviceVulkan12Features Features12{};
     Features12.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_2_FEATURES;
+    VkPhysicalDeviceVulkan13Features Features13{};
+    Features12.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_3_FEATURES;
```
spall, line 185: Features13?

lib/API/VK/Device.cpp
```diff
@@ -181,9 +181,16 @@ class VKDevice : public offloadtest::Device {
...
+    VkPhysicalDeviceVulkan14Features Features14{};
+    Features12.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_4_FEATURES;
```
spall, line 187: Features14?

spall (commented):
Just noticed this.

PR #223, opened 2025-06-04
Fix handling of device features for DX

llvm-beanz:
Previously this code was creating a merged feature set for WARP and the available device. This instead only configures the features for the native device when not using WARP, and only configures WARPs features when you are using WARP.

### DIFF
--- a/test/lit.cfg.py
+++ b/test/lit.cfg.py
@@ -79,10 +79,22 @@ devices = yaml.safe_load(query_string)
 
 for device in devices['Devices']:
   if device['API'] == "DirectX" and config.offloadtest_enable_d3d12:
+    if not config.offloadtest_test_warp:
+      # Don't add warps features if we're not using warp.
+      if "Microsoft Basic Render Driver" in device['Description']:
+        continue
+      if "Intel" in device['Description']:
+        config.available_features.add("DirectX-Intel")
+      if "NVIDIA" in device['Description']:
+        config.available_features.add("DirectX-NV")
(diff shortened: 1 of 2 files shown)


test/lit.cfg.py
```diff
@@ -79,10 +79,22 @@
...
+    if not config.offloadtest_test_warp:
+      # Don't add warps features if we're not using warp.
+      if "Microsoft Basic Render Driver" in device['Description']:
+        continue
+      if "Intel" in device['Description']:
+        config.available_features.add("DirectX-Intel")
+      if "NVIDIA" in device['Description']:
+        config.available_features.add("DirectX-NV")
+      if "AMD" in device['Description']:
+        config.available_features.add("DirectX-AMD")
```
bogner, line 91: These don't need to be nested under the check for if we're using warp or not - they're trivially false for warp devices. I think we could make this logic a bit clearer if we don't nest these checks.

How about something like this?
```c++
    is_warp = "Microsoft Basic Render Driver" in device['Description']
    if is_warp and not config.offloadtest_test_warp:
      continue
    if config.offloadtest_test_warp and not is_warp:
      continue
```

llvm-beanz: Oh... actually no they do.

We're inside an iterator iterating over all the devices, so whether or not we're selecting warp is relevant.

llvm-beanz: Oh... I guess we can continue instead of nest. Will update.

spall (approved):
lgtm

PR #230, opened 2025-06-10
Enable isinf.16.test with metal shader converter 3

llvm-beanz:
The Metal Shader Converter 3 beta released today fixes this test. This change marks it as unsupported on Metal with older versions of the shader converter.


alsepkow (commented):
LGTM. But I can't approve.

PR #231, opened 2025-06-10
Rework PR jobs to use a matrix

llvm-beanz:
This moves the PR workflows into a single matrix configuration which allows for completely separate workflows from the scheduled jobs. This fixes two problems with the PR workflows.

First it fixes the README files badges to only be based on scheduled jobs. This will allow the red/green status badges to reflect the actual state of main instead of the state of the last PR build.

Second, it uses concurrency groups to cancel in-progress runs for a pull request if the PR is updated. This should reduce the load on the builders when they are processing PRs with lots of updates.


.github/workflows/pr-matrix.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+jobs:
+  Exec-Tests-Windows:
+    strategy:
+      fail-fast: false
+      matrix:
+        SKU: [windows-intel]
+        TestTarget: [check-hlsl-d3d12, check-hlsl-warp-d3d12, check-hlsl-vk, check-hlsl-clang-d3d12, check-hlsl-clang-warp-d3d12, check-hlsl-clang-vk]
+
+    uses: ./.github/workflows/test-all.yaml
```
V-FEXrt, line 24: nit: This file could probably use a rename since it now seems to be a dependency of all the other files

V-FEXrt (approved):
Nice! LGTM! Love GHAction matrix runner

PR #237, opened 2025-06-16
Print the VK error code

s-perron:
If the device fails to be created and the error is not recorginzied, the
error message impossible to act on. If we output the error number, then
the user can look it up, and see what needs to be done.

### DIFF
--- a/lib/API/VK/Device.cpp
+++ b/lib/API/VK/Device.cpp
@@ -866,7 +866,7 @@ public:
     if (Res == VK_ERROR_INCOMPATIBLE_DRIVER)
       return llvm::createStringError(std::errc::no_such_device,
                                      "Cannot find a compatible Vulkan device");
-    if (Res) {
+    if (Res)
       return llvm::createStringError(std::errc::no_such_device,
                                      "Unknown Vulkan initialization error %d",
                                      Res);
--- a/lib/API/VK/Device.cpp
+++ b/lib/API/VK/Device.cpp
@@ -870,7 +870,6 @@ public:
       return llvm::createStringError(std::errc::no_such_device,
                                      "Unknown Vulkan initialization error %d",
                                      Res);
-    }
 
     DeviceCount = 0;
     if (vkEnumeratePhysicalDevices(Instance, &DeviceCount, nullptr))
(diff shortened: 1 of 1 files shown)


llvm-beanz (approved):

lib/API/VK/Device.cpp
```diff
@@ -834,9 +834,11 @@ class VKContext {
     if (Res == VK_ERROR_INCOMPATIBLE_DRIVER)
       return llvm::createStringError(std::errc::no_such_device,
                                      "Cannot find a compatible Vulkan device");
-    if (Res)
+    if (Res) {
```
llvm-beanz, line 837: nit:
```suggestion
    if (Res)
```

lib/API/VK/Device.cpp
```diff
@@ -834,9 +834,11 @@ class VKContext {
...
       return llvm::createStringError(std::errc::no_such_device,
-                                     "Unknown Vulkan initialization error");
+                                     "Unknown Vulkan initialization error: %d",
+                                     Res);
+    }
```
llvm-beanz, line 841:
```suggestion
```

lib/API/VK/Device.cpp
```diff
@@ -865,9 +867,11 @@ class VKContext {
     if (Res == VK_ERROR_INCOMPATIBLE_DRIVER)
       return llvm::createStringError(std::errc::no_such_device,
                                      "Cannot find a compatible Vulkan device");
-    if (Res)
+    if (Res) {
```
llvm-beanz, line 870:
```suggestion
    if (Res)
```

lib/API/VK/Device.cpp
```diff
@@ -865,9 +867,11 @@ class VKContext {
...
       return llvm::createStringError(std::errc::no_such_device,
-                                     "Unknown Vulkan initialization error");
+                                     "Unknown Vulkan initialization error %d",
+                                     Res);
+    }
```
llvm-beanz, line 874:
```suggestion
```

llvm-beanz (commented):
A few style nits for LLVM coding standards.

PR #256, opened 2025-06-18
[Vulkan] Add support for MoltenVK on macOS

llvm-beanz:
This gets Vulkan support working on macOS if the Vulkan SDK and MoltenVK are available. All current Vulkan tests except particle_life pass with Vulkan on macOS with DXC, and sign.32.test fails with Clang.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,11 +41,7 @@ macro(add_offloadtest_library name)
   set_target_properties(OffloadTest${name} PROPERTIES FOLDER "Offload Test/Libraries")
 endmacro()
 
-if (APPLE)
-  set(OPTIONAL_MOLTEN COMPONENTS MoltenVK)
-endif()
-
-find_package(Vulkan ${OPTIONAL_MOLTEN})
+find_package(Vulkan)
 message(STATUS "Vulkan Include Dirs: ${Vulkan_INCLUDE_DIRS}")
 if (Vulkan_INCLUDE_DIRS)
   set(OFFLOADTEST_ENABLE_VULKAN On)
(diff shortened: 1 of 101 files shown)


CMakeLists.txt
```diff
@@ -41,7 +41,11 @@ macro(add_offloadtest_library name)
   set_target_properties(OffloadTest${name} PROPERTIES FOLDER "Offload Test/Libraries")
 endmacro()
 
-find_package(Vulkan)
+if (APPLE)
+  set(OPTIONAL_MOLTEN COMPONENTS MoltenVK)
+endif()
+
+find_package(Vulkan ${OPTIONAL_MOLTEN})
```
llvm-beanz, line 48: TODO: Remove this. This wasn't actually needed.

llvm-beanz, line 48: Not needed.
```suggestion
find_package(Vulkan)
```

lib/API/CMakeLists.txt
```diff
@@ -24,7 +24,8 @@ if (APPLE)
                             "-framework MetalKit"
                             "-framework AppKit"
                             "-framework Foundation"
-                            "-framework QuartzCore")
+                            "-framework QuartzCore"
+                            ${api_frameworks})
```
llvm-beanz, line 28: Not needed.

```suggestion
                            "-framework QuartzCore")
```

test/Feature/HLSLLib/sign.32.test
```diff
@@ -176,6 +176,9 @@ DescriptorSets:
 # https://github.com/microsoft/DirectXShaderCompiler/issues/7512
 # XFAIL: DXC-Vulkan
 
+# No idea what is going on here, but Out1 and Out3 are coming through as 0's.
+# XFAIL: Clang && Vulkan-Darwin
```
farzonl, line 180: Particle life is probably harder, but this one feel like we should have a follow up task to figure out what is going wrong.

PR #257, opened 2025-06-18
Initial support for Texture2D types

### LINKED ISSUE #208: Basic Texture2D support
We need to implement Texture2D support for the offload suite. For basic support we should just use defined buffers that we convert to textures. This should be fairly straightforward.

llvm-beanz:
This adds initial support for Texture2D and RWTexture2D across DirectX, Vulkan and Metal.

Fixes #208

### DIFF
--- a/lib/API/VK/Device.cpp
+++ b/lib/API/VK/Device.cpp
@@ -519,7 +519,7 @@ public:
     ImageCreateInfo.usage =
         VK_IMAGE_USAGE_TRANSFER_DST_BIT |
         (R.isReadWrite()
-             ? (VK_IMAGE_USAGE_STORAGE_BIT| VK_IMAGE_USAGE_TRANSFER_SRC_BIT)
+             ? (VK_IMAGE_USAGE_STORAGE_BIT | VK_IMAGE_USAGE_TRANSFER_SRC_BIT)
              : VK_IMAGE_USAGE_SAMPLED_BIT);
 
     VkImage Image;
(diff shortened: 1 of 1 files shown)


lib/API/VK/Device.cpp
```diff
@@ -556,41 +648,77 @@ class VKDevice : public offloadtest::Device {
     llvm::SmallVector<VkWriteDescriptorSet> WriteDescriptors;
     assert(IS.BufferViews.empty());
     llvm::SmallVector<VkDescriptorBufferInfo> RawBufferInfos;
+    llvm::SmallVector<VkDescriptorImageInfo> ImageInfos;
+
+    // This massively over-reserves the descriptor info structs, but prevents
+    // dumb things from happening if these get reallocated...
+    const uint32_t DescriptorCount = P.getDescriptorCount();
+    RawBufferInfos.reserve(DescriptorCount);
+    ImageInfos.reserve(DescriptorCount);
```
bogner, line 657: If we need the memory to be fixed in place we just shouldn't use a data structure that's resizable at all. Something like a `std::unique_ptr<VkDescriptorBufferInfo[]>` would be better.

test/Feature/Textures/SRVToUAV.test.yaml
bogner, line 1: We haven't used this `.yaml` suffix on the other tests. Mistake?

bogner (commented):
This mostly looks reasonable to me, though I'm not the most familiar with the host-side APIs

bogner (approved):
It's a step in the right direction - looks good. As you point out in the comments in the VK implementation we'll need to revisit this somewhat for arrays, but I think that'll be easier to do if this is in place first.

include/Support/Pipeline.h
```diff
@@ -149,6 +153,23 @@ struct Resource {
...
+    switch (Kind) {
+    case ResourceKind::Buffer:
+    case ResourceKind::RWBuffer:
+    case ResourceKind::StructuredBuffer:
+    case ResourceKind::RWStructuredBuffer:
+    case ResourceKind::ByteAddressBuffer:
+    case ResourceKind::RWByteAddressBuffer:
+    case ResourceKind::ConstantBuffer:
+      return false;
+    case ResourceKind::Texture2D:
```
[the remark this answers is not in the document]
llvm-beanz: In the cases like `IsStructuredBuffer` there are only ever two cases that will be true and we've added those two types to the enum, any other case will be false. That means as we add more enum values we don't need to revisit the switch statement because it will be correct in the future.

This switch is different. As we add new enum values, we should consider whether or not they are textures. Not all the things we add in the future will be textures, and we haven't added all the possible texture formats.

PR #258, opened 2025-06-18
Make validation errors a test error.

s-perron:
The current use of the validation layer will emit a message, but it will
not cause tests to fail. By adding a call back that returns true for
error messages and warnings, test that have a validation error or
warning will fail.

### DIFF
--- a/test/Feature/WaveOps/WaveActiveAllTrue.test
+++ b/test/Feature/WaveOps/WaveActiveAllTrue.test
@@ -63,6 +63,9 @@ DescriptorSets:
 ...
 #--- end
 
+# https://github.com/llvm/llvm-project/issues/145513
+# XFAIL: Clang-Vulkan
+
 # RUN: split-file %s %t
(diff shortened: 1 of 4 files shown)


llvm-beanz:
@s-perron, I was discussing related issues with @spall last night. We definitely want to get this in. Sarah spent time yesterday chasing a failure that had validation errors which didn't cause the test to fail outright, instead it was failing in verification.

Filing a tracking issue on the relevant compiler(s) and XFAILing the tests seems like the right approach to me.

llvm-beanz (approved):

s-perron:
FYI: I'm trying to track down why the Intel tests had validation error for [f2aadab](https://github.com/llvm/offload-test-suite/pull/258/commits/f2aadabc1a04fe0a3d70dbd932c0cf59f414783a). I do not have an Intel GPU available right now, so I'll be updating to PR to see what works.

spall (approved):

PR #270, opened 2025-06-21
Add `distance` tests

bob80905:
This PR adds tests for the `distance` function. Adds float16 and float32 test files.
Fixes https://github.com/llvm/offload-test-suite/issues/175


bob80905:
Should I add a test case for NaNs? Or other non-normal floats?

test/Feature/HLSLLib/distance.32.test
```diff
@@ -0,0 +1,95 @@
...
+  float R1 = distance(X[0].xy, Y[0].xy);
+  Result[1] = R1;
+  // distance({0.0, 0.0, 0.0}, {3.0, 4.0, 12.0}) = 13.0
+  float R2 = distance(X[0].xyz, Y[0].xyz);
+  Result[2] = R2;
+  // distance({0.0, 0.0, 0.0, 0.0}, {3.0, 4.0, 12.0, 84.0}) = 85.0
+  float R3 = distance(X[0].xyzw, Y[0].xyzw);
+  Result[3] = R3;
+
+  // test denormal cases
```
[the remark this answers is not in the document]
spall: https://github.com/llvm/offload-test-suite/pull/268/files
you can look here for examples of 16 and 32 bit float values which should be denorm according to c++

test/Feature/HLSLLib/distance.32.denormal.test
```diff
@@ -0,0 +1,92 @@
...
+StructuredBuffer<float4> X : register(t0);
+StructuredBuffer<float4> Y : register(t1);
+
+RWStructuredBuffer<float> Result : register(u2);
+
+[numthreads(1,1,1)]
+void main() {
+  // test special cases
+  // distance(inf, 1.0) = inf
+  Result[0] = distance(X[0].x, Y[0].x);
```
tex3d, line 12: The degenerate scalar case is not that interesting for distance, since it just translates to fsub. Vector distance translates to fsub, fmul+fadd per component, sqrt.

So these scalar cases are really only testing the behavior of fsub.

test/Feature/HLSLLib/distance.fp16.test
```diff
@@ -0,0 +1,84 @@
...
+      Kind: RWStructuredBuffer
+      DirectXBinding:
+        Register: 2
+        Space: 0
+      VulkanBinding:
+        Binding: 2
+...
+#--- end
+
+# UNSUPPORTED: Clang-Vulkan
```
spall, line 80: Why is Clang-Vulkan unsupported? Sorry if this has been explained before and I forgot or missed it.

spall: Can you leave a comment about that in the tests?

test/Feature/HLSLLib/distance.32.denormal.test
```diff
@@ -0,0 +1,91 @@
...
+---
+Shaders:
+  - Stage: Compute
+    Entry: main
+    DispatchSize: [1, 1, 1]
+Buffers:
+  - Name: X
+    Format: Float32
+    Stride: 16
+    Data: [ inf, NaN, -inf, 0 ]
```
alsepkow, line 44: Would -0 also an interesting value to test?

bob80905: It would, looks like it didn't affect any of the expected results however.

test/Feature/HLSLLib/distance.fp16.test
```diff
@@ -0,0 +1,100 @@
+#--- source.hlsl
```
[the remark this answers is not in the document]
spall: yeah i think that is good

Icohedron (approved):

PR #271, opened 2025-06-24
XFAIL isinf.16 test for Clang on nVidia drivers

bogner:
DXC never generated fp16 variants of this, so some drivers have issues with it.


spall (approved):
fails on my machine as well.

PR #274, opened 2025-06-30
Add optional field Fill and Size for buffer stage

ArberSephirotheca:
This PR add optional field Fill and Size for buffer stage in the pipeline.
`Fill` specifies the value to use, and `Size` determines the number of entries. Together they initialize the buffer with repeated Fill values.

The common use cases is when you want to create a larger buffer that fill with a known value
```yaml
  - Name: Buffer
    Format: UInt32
    Stride: 4
    Fill: 1
    Size: 1234578
```
This creates a `UInt32` buffer with 1234578 entries and filled with 1s.


bogner:
This shouldn't be necessary any more after https://github.com/llvm/offload-test-suite/pull/474

PR #278, opened 2025-07-08
Un-XFAIL test now that llvm/llvm-project#144580 is resolved

bogner:
This test no longer fails because of the original issue, but still has issues
on intel drivers due to llvm/llvm-project#148270

### DIFF
--- a/test/Basic/TestPipeline.test
+++ b/test/Basic/TestPipeline.test
@@ -40,9 +40,6 @@ DescriptorSets:
 ...
 #--- end
 
-# Tracking issue: https://github.com/llvm/llvm-project/issues/144580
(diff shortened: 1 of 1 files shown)


s-perron (approved):

s-perron (changes requested):
I don't think we can enable this test yet. The problem is that the Intel driver does not support storage images with unknown format. Note that the test will pass if you use a driver that supports this feature, which make the XFAIL awkward.

 I need to change the clang compiler to match DXC, and guess at the format. I just opened https://github.com/llvm/llvm-project/issues/148270 to fix this issue.

s-perron (approved):

PR #280, opened 2025-07-08
Enable most of the cbuffer tests for clang

bogner:
These have been working for some time but we never enabled the tests.

### DIFF
--- a/test/Feature/CBuffer/arrays-16bit.test
+++ b/test/Feature/CBuffer/arrays-16bit.test
@@ -59,8 +59,8 @@ DescriptorSets:
 # DXC's vulkan support does not layout cbuffers compatibly with DXIL
 # UNSUPPORTED: Vulkan
 
-# https://github.com/llvm/llvm-project/issues/110722
-# XFAIL: Clang
+# https://github.com/llvm/llvm-project/issues/138996
+# UNSUPPORTED: Clang
 
 # RUN: split-file %s %t
 # RUN: %dxc_target -enable-16bit-types -T cs_6_5 -Fo %t.o %t/source.hlsl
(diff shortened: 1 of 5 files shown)


bogner:
Test failure is the XPASS being resolved in https://github.com/llvm/offload-test-suite/pull/284

Icohedron (approved):

test/Feature/CBuffer/arrays-16bit.test
```diff
@@ -59,8 +59,8 @@ DescriptorSets:
 # DXC's vulkan support does not layout cbuffers compatibly with DXIL
 # UNSUPPORTED: Vulkan
 
-# https://github.com/llvm/llvm-project/issues/110722
-# XFAIL: Clang
+# https://github.com/llvm/llvm-project/issues/138996
+# UNSUPPORTED: Clang
```
llvm-beanz, line 63: Do we want this UNSUPPORTED or XFAIL?

test/Feature/CBuffer/lit.local.cfg
```diff
@@ -1,4 +1,4 @@
-if 'Clang' in config.available_features:
+if 'Clang-Vulkan' in config.available_features:
     config.unsupported = True
```
llvm-beanz, line 2: Should we just be enabling these now?

llvm-beanz (commented):
General feedback: should these be unsupported or XFAIL so that we know when they start working?

test/Feature/CBuffer/lit.local.cfg
bogner: Well, none of these tests pass under vulkan. I could remove this and add XFAILs to the tests directly if you think that makes more sense.

llvm-beanz: I don't have a strong feeling, I just want to make sure that we don't somehow lose track and not enable the tests once they do start passing.

bogner:
pre-checkin CI is in a bit of a rough spot, but the cbuffer tests themselves are passing.

PR #287, opened 2025-07-11
Add `reflect` tests

bob80905:
This PR adds tests for the `reflect` function. Adds float16 and float32 test files.
Fixes https://github.com/llvm/offload-test-suite/issues/170


Icohedron (approved):
Tests look correct to me

test/Feature/HLSLLib/reflect.16.test
```diff
@@ -0,0 +1,209 @@
...
+// This test tests three different reflection scenarios:
+// -----> |
+// -----> /
+// -----> -
+// each test case will cycle through 1-4 dimensions.
+
+StructuredBuffer<half4> IncidentRay : register(t0);
+StructuredBuffer<half4> Wall1 : register(t1);
+StructuredBuffer<half4> Wall2 : register(t2);
+StructuredBuffer<half4> Wall3 : register(t3);
```
tex3d, line 12: You're using StructuredBuffers, which means you can define everything you need in one buffer for simplicity, right?  Same with the output buffer.

test/Feature/HLSLLib/reflect.16.test
```diff
@@ -0,0 +1,209 @@
...
+    DispatchSize: [1, 1, 1]
+Buffers:
+  - Name: IncidentRay
+    Format: Float16
+    Stride: 8
+    Data: [ 0x4100, 0x8000, 0x0000, 0x0000 ]
+  - Name: Wall1
+    Format: Float16
+    Stride: 8
+    Data: [ 0xbc00, 0x0000, 0x0000, 0x0000 ]
```
tex3d, line 77: Why are most of the components for all of the inputs zeros?  Axis aligned edge cases are probably the least interesting.  A few random number inputs in a reasonable range should be sufficient to test the math expansion, and would be better than leaving z and w dimensions zeroed for all inputs (in my opinion).

test/Feature/HLSLLib/reflect.32.test
```diff
@@ -0,0 +1,209 @@
...
+    DispatchSize: [1, 1, 1]
+Buffers:
+  - Name: IncidentRay
+    Format: Float32
+    Stride: 16
+    Data: [ 2.5, -0.0, 0.0, 0.0 ]
+  - Name: Wall1
+    Format: Float32
+    Stride: 16
+    Data: [ -1.0, 0.0, 0.0, 0.0 ]
```
tex3d, line 77: Using 1.0 and so many zero components means this is only testing a degenerate edge case.  I really think we should be testing with combinations of numbers in reasonable ranges and using a ULP tolerance at least as high as the potential variance in expansions and operation precisions.

tex3d (commented):
This seems too focused on a degenerate edge case, uses more resources than it would seem necessary, and should include literal cases to test constant-folding.

test/Feature/HLSLLib/reflect.16.test
```diff
@@ -0,0 +1,187 @@
...
+    Data: [ 0xb0f8, 0xbb3e, 0xb1a8, 0x0000, 0xb0f8, 0xbb3e, 0xb1a8, 0x0000 ]
+
+  - Name: Result4D
+    Format: Float16
+    Stride: 8
+    ZeroInitSize: 16
+  - Name: ExpectedResult4D
+    Format: Float16
+    Stride: 8
+    Data: [ 0x3400, 0xb800, 0x3800, 0xba00, 0x3400, 0xb800, 0x3800, 0xba00 ]
```
inbelic, line 97: nit: Can we annotate what the decimal values are for these (and elsewhere)

test/Feature/HLSLLib/reflect.16.test
```diff
@@ -0,0 +1,187 @@
...
+  // 3D case, using half4 for alignment
+  half4 result3D = half4(reflect(IncidentRay3D[0].xyz, normalize(Wall3D[0].xyz)), 0.0);
+  Result3D[0] = result3D;
+  half4 result3D_constant = half4(reflect(half3(0.5, -0.25, 0.75), half3(0.5, 0.5, 0.70710677)), 0.0);
+  Result3D[1] = result3D_constant;
+
+  // 4D case
+  half4 result4D = reflect(IncidentRay4D[0].xyzw, normalize(Wall4D[0].xyzw));
+  Result4D[0] = result4D;
+  half4 result4D_constant = reflect(half4(0.5, -0.25, 0.75, -0.5), half4(0.5, 0.5, 0.5, 0.5));
```
inbelic, line 33: I believe all the values that are used in this test (and all others) are exactly representable in IEE 754. Should we have an extra test that uses values that will be rounded?

bob80905: 0.70710677 gets rounded to 0.70703125 in the 3D test. Does that resolve your concern?

inbelic (approved):

PR #294, opened 2025-07-20
[VK] Fix handling of vk descriptor binding

llvm-beanz:
This should fix handling for bindings which was oddly always broken.


alsepkow (commented):
lgtm

PR #298, opened 2025-07-24
add test for asuint16

### LINKED ISSUE #179: Add test for asuint16
https://github.com/llvm/llvm-project/blob/0ee40ca5ccbf036179ee11530cc91de7d31131aa/clang/lib/Headers/hlsl/hlsl_intrinsics.h#L113

- [ ] Make sure to test for all vector sizes

I think this test should use BufferExact rule

A .test file should contain 
```
@@ -0,0 +1,54 @@
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# RUN: split-file %s %t
```

spall:
Add test for asuint16 for half, uint16_t and int16_t.
Closes #179

### DIFF
--- a/test/Feature/HLSLLib/asuint16.int16.test
+++ b/test/Feature/HLSLLib/asuint16.int16.test
@@ -92,9 +92,6 @@ DescriptorSets:
         Binding: 3
 #--- end
 
-# https://github.com/llvm/llvm-project/issues/150482
-# XFAIL: Clang-DirectX
-
 # REQUIRES: Int16
 # RUN: split-file %s %t
 # RUN: %dxc_target -enable-16bit-types -T cs_6_5 -Fo %t.o %t/source.hlsl
--- a/test/Feature/HLSLLib/asuint16.int16.test
+++ b/test/Feature/HLSLLib/asuint16.int16.test
@@ -1,0 +1,101 @@
+#--- source.hlsl
+StructuredBuffer<uint16_t4> In0 : register(t0);
+StructuredBuffer<int16_t4> In1 : register(t1);
+
+RWStructuredBuffer<uint16_t4> Out0 : register(u2);
+RWStructuredBuffer<uint16_t4> Out1 : register(u3);
+
+
+[numthreads(1,1,1)]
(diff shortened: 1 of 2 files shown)


test/Feature/HLSLLib/asuint16.int16.test
```diff
@@ -0,0 +1,98 @@
...
+    Stride: 8
+    Data: [0, 100, 65535, 10, 20, 30, 40, 50, 60, 70, 80, 6000, 0, 100, 65535, 10]
+  - Name: Out1
+    Format: Int16
+    Stride: 8
+    ZeroInitSize: 32
+  - Name: ExpectedOut1
+    Format: Int16
+    Stride: 8
+    Data: [ 0, 32767, -100, -32768, 10, 20, 30, 6000, -3000, -30000, 50, 60, 0, 32767, -100, -32768 ]
```
bob80905, line 53: I don't understand why there are negative values, shouldn't asuint interpret the bits as an unsigned integer, and thus eliminate the sign in the final representation?

bob80905 (commented):
Main could use some indentation

kmpeng (approved):
LGTM

PR #302, opened 2025-07-28
Add support for resource arrays to the offload test suite

hekota:
Introduces support for resource arrays across the offload test suite, including:
- Data input and output handling
- Comparison of expected results
- Integration with both DirectX and Vulkan runtimes (Metal is not supported yet)

A new property `ArraySize` has been added to the `Buffer` definition. If `ArraySize` is `1` (the default), the `Data` field of the `Buffer` remains unchanged:

```
Data: [0, 1, 2, 3]
```

If `ArraySize` is greater than `1`, the `Data` field must be specified as a list of arrays:
```
Data:
  - [0, 1, 2, 3]
  - [4, 5, 6, 7]
  - [8, 9, 10, 11]
```

If the resource array includes resources with counters, their values will be output using the `Counters` field:
```
Counters: [1, 2, 3]
```

Closes llvm/wg-hlsl#291


lib/Support/Pipeline.cpp
```diff
@@ -93,29 +93,101 @@ void MappingTraits<offloadtest::DescriptorSet>::mapping(
...
+//
+static void setCounters(IO &I, offloadtest::Buffer &B) {
+  // counters are printed only if they exist and only on output
+  if (!I.outputting() || B.Counters.empty())
+    return;
+
+  if (B.ArraySize == 1) {
+    assert(B.Counters.size() == 1 &&
+           "expected a single counter for a single buffer");
+    I.mapRequired("Counter", B.Counters.back());
```
bob80905, line 185: Can't this just be
`I.mapRequired("Counter", B.Counters);` ?

test/Feature/ResourceArrays/array-global.test
```diff
@@ -0,0 +1,136 @@
...
+# RUN: %offloader %t/pipeline.yaml %t.o | FileCheck %s
+
+// CHECK: Name: BufB
+// CHECK: Format: Int32
+// CHECK: ArraySize: 4
+// CHECK: Data:
+// CHECK:   - [ 0, 1, 2, 3, 4, 5, 6, 7 ]
+// CHECK:   - [ 0, 2, 4, 6, 8, 10, 12, 14 ]
+// CHECK:   - [ 0, 3, 6, 9, 12, 15, 18, 21 ]
+// CHECK:   - [ 0, 4, 8, 12, 16, 20, 24, 28 ]
```
bob80905, line 136: Can we see the output of BufD as well? I wonder if the floats are being printed well.

test/Feature/StructuredBuffer/inc_counter_array.test
```diff
@@ -0,0 +1,57 @@
...
+# RUN: %dxc_target -T cs_6_0 -Fo %t.o %t/source.hlsl
+# RUN: %offloader %t/pipeline.yaml %t.o | FileCheck %s
+
+# CHECK: Creating UAV: { Size = 4100, Register = u0, Space = 0, HasCounter = 1 }
+# CHECK: UAV: HeapIdx = 0 EltSize = 4 NumElts = 1 HasCounter = 1
+
+# CHECK: Name: Out
+# CHECK: Counters: [ 1, 2, 3, 4 ]
+# CHECK: Data:
+      - [ 0x0 ]
```
bob80905, line 54: I don't mind, but others might want the data to copy the previous pattern of printing out the integer, rather than hex format.

test/Feature/Textures/SRVToUAV-array.test
```diff
@@ -0,0 +1,91 @@
+#--- source.hlsl
+
```
bob80905, line 2: Would be great to continue the pattern of describing the purpose of the test at the top here.

test/Feature/StructuredBuffer/inc_counter_array.test
llvm-beanz: I think if the data format is hex (as it is in this test), printing as hex is the expected result.

lib/Support/Pipeline.cpp
hekota: That would output `Counter: [ 1 ]` even if the resource is not an array. I prefer to keep it as `Counter: 1`.

lib/API/VK/Device.cpp
```diff
@@ -803,53 +806,58 @@ class VKDevice : public offloadtest::Device {
...
         WDS.sType = VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET;
         WDS.dstSet = IS.DescriptorSets[SetIdx];
         WDS.dstBinding = R.VKBinding->Binding;
-        WDS.descriptorCount = 1;
+        WDS.descriptorCount = R.BufferPtr->ArraySize;
         WDS.descriptorType = getDescriptorType(R.Kind);
         if (R.isTexture())
-          WDS.pImageInfo = &DescriptorData[BufIdx].ImageInfo;
+          WDS.pImageInfo =
+              &DescriptorData[IndexOfFirstBufferDataInArray].ImageInfo;
```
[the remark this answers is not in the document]
hekota: I'll update it use small vectors space that are pre-allocated to the correct size for each type and number of descriptors.

PR #306, opened 2025-07-31
Add `refract` tests

bob80905:
This PR adds tests for the refract function. Adds float16 and float32 test files.
Fixes https://github.com/llvm/offload-test-suite/issues/191

### DIFF
--- a/test/Feature/HLSLLib/refract.16.test
+++ b/test/Feature/HLSLLib/refract.16.test
@@ -1,0 +1,192 @@
+#--- source.hlsl
+
+// This test tests three different refraction scenarios
+// One in 2D, 3D, and 4D
+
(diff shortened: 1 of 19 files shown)


test/Feature/HLSLLib/refract.16.test
```diff
@@ -0,0 +1,193 @@
...
+    - Name: Result4D
+      Kind: RWStructuredBuffer
+      DirectXBinding:
+        Register: 8
+        Space: 0
+      VulkanBinding:
+        Binding: 8      
+...
+#--- end
+# UNSUPPORTED: Clang-Vulkan
```
llvm-beanz, line 188: I think we settled on wanting to use `XFAIL` instead of `UNSUPPORTED` so that we do track when support comes in.

Icohedron (approved):

PR #308, opened 2025-08-01
Add tests for resource arrays

hekota:
Add tests for arrays of resources to the offload test suite, including arrays declared locally or used as function arguments, multi-dimensional arrays, or subsets of multi-dimensional arrays. One test for global arrays was already added in #302 when support for resource arrays was added to the offload test suite.

Closes llvm/wg-hlsl#292

### DIFF
--- a/test/Feature/ResourceArrays/array-local2.test
+++ b/test/Feature/ResourceArrays/array-local2.test
@@ -1,22 +1,24 @@
 #--- source.hlsl
 
-// Another variation testing resource arrays used as a local
-// variable and function argument, making sure the array is
-// always a local copy.
+// Another variation testing resource arrays used as a local variable and
+// function argument. In this case the function SomeFn modifies the array B,
+// which should be a local copy of array A that is defined in main function.
+// The test verifies that this modification of local array B does not affect
+// array A.
 
 RWBuffer<int> X : register(u0);
 RWBuffer<int> Y : register(u1);
 
-void SomeFn(RWBuffer<int> Arr[2], uint Idx, int Val0) {
-  Arr[0] = Y;
-  Arr[0][Idx] = Val0;
+void SomeFn(RWBuffer<int> B[2], uint Idx, int Val0) {
+  B[0] = Y;
+  B[0][Idx] = Val0;
 }
 
 [numthreads(4,1,1)]
 void main(uint GI : SV_GroupIndex) {
-  RWBuffer<int> Arr[2] = {X, Y};
-  SomeFn(Arr, GI, 1);
-  Arr[0][GI] = 2;
+  RWBuffer<int> A[2] = {X, Y};
+  SomeFn(A, GI, 1);
+  A[0][GI] = 2;
 }
 
 //--- pipeline.yaml
@@ -74,7 +76,8 @@ DescriptorSets:
 
 # UNSUPPORTED: Clang
 
-# DXC-DirectX has a bug here
+# DXC-DirectX has a bug here because it does not create a local copy
+# of a function argument if the type is resource or resource array.
 https://github.com/microsoft/DirectXShaderCompiler/issues/7678
 # XFAIL: DXC-DirectX
 
--- a/test/Feature/ResourceArrays/array-local2.test
+++ b/test/Feature/ResourceArrays/array-local2.test
@@ -74,9 +74,8 @@ DescriptorSets:
 
 # UNSUPPORTED: Clang
 
-# DXC-DirectX has a bug here and produces wrong output:
-# BufX: [ 0, 0, 0, 0 ]
-# BufY: [ 2, 2, 2, 2 ]
+# DXC-DirectX has a bug here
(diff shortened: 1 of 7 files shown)


test/Feature/ResourceArrays/multi-dim-array-subset.test
```diff
@@ -0,0 +1,101 @@
...
+    Format: Float32
+    ZeroInitSize: 16
+
+  - Name: BufOut2
+    Format: Float32
+    ZeroInitSize: 16
+
+  - Name: ExpectedOuts
+    Format: Float32
+    Data: [ 1.1, 6.6, 12.12, 17.17 ]
```
hekota, line 51: Can somebody please double check my math, if the expected results are correct?

DXC output is `[ 2, 12, 13.2, 18.7 ]`. I can't figure out how could it come up with that, given that each `BufOut1` value should be a sum of 2 values from `BufIn`.

alsepkow: The math looks correct to me. Even tried asking co-pilot to confirm that.
Is it possible there is an issue because In[4][2] is assigned to register(u0), and then Out is register(u0, space1)?

test/Feature/ResourceArrays/multi-dim-array-local.test
```diff
@@ -0,0 +1,94 @@
...
+      DirectXBinding:
+        Register: 2
+        Space: 0
+...
+#--- end
+
+# UNSUPPORTED: Clang
+
+# DXC has a bug and produces wrong output:
+# [ 210.12, 321.231, 432.342, 543.453, 654.564, 765.675, 876.786, 987.897 ]
```
alsepkow, line 82: Similar comment here about whether or not we want to reference an issue.

test/Feature/ResourceArrays/multi-dim-array-subset.test
hekota: Please disregard - this was a user error :) This is working correctly on DXC. I probably had the test source in a bad state so it did not compile, and the test ran and compared some older code version with new pipeline definition.

test/Feature/ResourceArrays/array-local2.test
```diff
@@ -0,0 +1,87 @@
+#--- source.hlsl
+
+// Another variation testing resource arrays used as a local
+// variable and function argument, making sure the array is
+// always a local copy.
+
+RWBuffer<int> X : register(u0);
+RWBuffer<int> Y : register(u1);
+
+void SomeFn(RWBuffer<int> Arr[2], uint Idx, int Val0) {
```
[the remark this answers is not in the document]
hekota: I will update the test description to explain what is happening. The `in` specifier is implied and there is intentionally no `out`. The bug in DXC is that it does not create a local copy of a function argument if the type is a resource array.

test/Feature/ResourceArrays/multi-dim-unbounded-array.test
```diff
@@ -0,0 +1,73 @@
...
+      - [ 5, 6, 7, 8 ]
+      - [ 6, 7, 8, 9 ]
+      - [ 7, 8, 9, 10 ]
+
+  - Name: ExpectedBuf
+    Format: Int32
+    ArraySize: 8
+    Data:
+      - [ 0, 1, 2, 3 ]
+      - [ 401, 402, 403, 404 ]
```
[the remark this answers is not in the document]
hekota: The reason why this different from the `unbounded-array.test` is because this is a two-dimensional array indexed by `(x, y)`. The `Buf` data is loaded from a flattened `Data` list like this:
```
Buf[0][0]: [ 0, 1, 2, 3 ]
Buf[0][1]: [ 1, 2, 3, 4 ]
Buf[1][0]: [ 2, 3, 4, 5 ]
Buf[1][1]: [ 3, 4, 5, 6 ]
...
```
Now when we have `numthreads(4,2,1)`, the `SV_GroupIndex`  (variable `GI`) for thread `(0, 1, 0)` is calculated as `0 * 4 * 2 + 1 * 4 + 0 = 4` (see the diagram [here](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/sv-groupthreadid)). 

The test is adding `100 * GI` to `Buf` values, so for `Buf[0][1]` the output is `[ 401, 402, 403, 404 ]`. And since the output again a two-dimensional array, it gets written out the same was as it was read, so `Buf[0][1]` is the second row.

test/Feature/ResourceArrays/multi-dim-array-global.test
```diff
@@ -0,0 +1,82 @@
+#--- source.hlsl
+
+// This test verified handling of global multi-dimensional resource arrays.
+
+RWBuffer<int> In[2][3] : register(u0);
+RWBuffer<int> Out[2][3] : register(u6);
+
+[numthreads(4,2,1)]
```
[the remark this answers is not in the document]
hekota: Good catch!

PR #313, opened 2025-08-05
Add windows-amd and windows-nvidia GitHub workflows for the two new machines

Icohedron:
This PR duplicates the existing windows-intel github workflows to support the new windows-amd and windows-nvidia machines. 

windows-amd and windows-nvidia have also been added to the pr-matrix workflow SKUs for the Exec-Tests-Windows job.

### DIFF
--- a/.github/workflows/pr-matrix.yaml
+++ b/.github/workflows/pr-matrix.yaml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        SKU: [windows-intel]
+        SKU: [windows-amd, windows-intel, windows-nvidia]
         TestTarget: [check-hlsl-d3d12, check-hlsl-warp-d3d12, check-hlsl-vk, check-hlsl-clang-d3d12, check-hlsl-clang-warp-d3d12, check-hlsl-clang-vk]
 
     uses: ./.github/workflows/build-and-test-callable.yaml
--- a/.github/workflows/windows-amd-clang-d3d12.yaml
+++ b/.github/workflows/windows-amd-clang-d3d12.yaml
@@ -1,0 +1,20 @@
+name: Windows D3D12 AMD Clang
+
+permissions:
+  contents: read
+  checks: write
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 */2 * * *' # run every 2 hours
+
+jobs:
+  Windows-D3D12-AMD-Clang:
+    uses: ./.github/workflows/build-and-test-callable.yaml
+    with:
+      OS: windows
+      SKU: windows-amd
+      TestTarget: check-hlsl-clang-d3d12
+      OffloadTest-branch: ${{ github.ref }}
+      LLVM-ExtraCMakeArgs: -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DOFFLOADTEST_USE_CLANG_TIDY=On
--- a/.github/workflows/windows-amd-clang-vk.yaml
+++ b/.github/workflows/windows-amd-clang-vk.yaml
@@ -1,0 +1,20 @@
+name: Windows Vulkan AMD Clang
+
+permissions:
+  contents: read
+  checks: write
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 */2 * * *' # run every 2 hours
+
+jobs:
+  Windows-VK-AMD-Clang:
+    uses: ./.github/workflows/build-and-test-callable.yaml
+    with:
+      OS: windows
+      SKU: windows-amd
+      TestTarget: check-hlsl-clang-vk
+      OffloadTest-branch: ${{ github.ref }}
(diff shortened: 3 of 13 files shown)


damyanp:
The duplication here concerns me.  Why can we use a matrix for the PR jobs, but not the others?

damyanp:
> I'm not sure we need to duplicate the WARP workflows though. Wouldn't the WARP test result be expected to be the same across all machines? We could designate just one machine to running the WARP tests.

They'd certainly be expected to have the same results.

As the same time, if they gave different results then that'd be good to know!

damyanp (approved):

llvm-beanz:
> As the same time, if they gave different results then that'd be good to know!

I think if we wanted to run the WARP tests on different underlying hardware we should do that only in the periodic builds not the pre-merge builds. We could alternatively just allow the WARP scheduled builds to use any windows builder which will cycle it through based on machine availability. That might be good enough to provide some coverage without slowing down our builds too much.

README.md
```diff
@@ -5,9 +5,13 @@ Experimental Runtime test suite for HLSL
...
+| Windows Vulkan AMD GPU | ![DXC](https://github.com/llvm/offload-test-suite/actions/workflows/windows-amd-dxc-vk.yaml/badge.svg) | ![Clang](https://github.com/llvm/offload-test-suite/actions/workflows/windows-amd-clang-vk.yaml/badge.svg) |
 | Windows Vulkan Intel GPU | ![DXC](https://github.com/llvm/offload-test-suite/actions/workflows/windows-intel-dxc-vk.yaml/badge.svg) | ![Clang](https://github.com/llvm/offload-test-suite/actions/workflows/windows-intel-clang-vk.yaml/badge.svg) |
+| Windows Vulkan NVIDIA GPU | ![DXC](https://github.com/llvm/offload-test-suite/actions/workflows/windows-nvidia-dxc-vk.yaml/badge.svg) | ![Clang](https://github.com/llvm/offload-test-suite/actions/workflows/windows-nvidia-clang-vk.yaml/badge.svg) |
 | macOS Apple M1 | ![DXC](https://github.com/llvm/offload-test-suite/actions/workflows/macos-dxc-mtl.yaml/badge.svg) | ![Clang & DXC](https://github.com/llvm/offload-test-suite/actions/workflows/macos-clang-mtl.yaml/badge.svg) |
```
llvm-beanz, line 17: Can you also create a new markdown doc file that includes the system specs for the new build machines? I think good information to have there would be something like:

```markdown
# Build Machine Information

## Windows Intel GPU

* CPU: Intel Xeon <I'll figure out the model>
* GPU: Intel UHD Graphics 630
* RAM: <I should figure this out too>
```

docs/BuildMachineInfo.md
```diff
@@ -0,0 +1,19 @@
...
+
+* CPU: AMD Ryzen 7 9700X
+* GPU: AMD Radeon RX 9070
+* RAM: 32 GiB
+
+## Windows Intel GPU
+
+* CPU: Intel Xeon <I'll figure out the model>
+* GPU: Intel UHD Graphics 630
+* RAM: <I should figure this out too>
```
Icohedron, line 13: Waiting on @llvm-beanz to provide the specs for the intel machine.

.github/workflows/pr-matrix.yaml
```diff
@@ -18,8 +18,23 @@ jobs:
...
+      SKU: ${{ matrix.SKU }}
+      TestTarget: ${{ matrix.TestTarget }}
+      OffloadTest-branch: ${{ github.event.pull_request.head.sha }}
+      LLVM-ExtraCMakeArgs: -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DOFFLOADTEST_USE_CLANG_TIDY=On
+
+  Exec-Tests-Windows-Warp:
+    strategy:
+      fail-fast: false
+      matrix:
+        SKU: [windows-amd]
```
Icohedron, line 36: I'm not sure how to pick a SKU randomly or based on availability, so I picked the AMD machine because it has the better CPU and the most RAM to work with between the NVIDIA and AMD machines. 

I'm not sure yet what the specs of the Intel machine are, nor if we would would want to keep running the Warp tests on it because it's not a lab machine.

README.md
damyanp: But maybe not needed for this PR.

docs/BuildMachineInfo.md
llvm-beanz: I'm having trouble remoting into the machine. Can you just leave blank lines? I'll hunt it down and update after this PR merges.

.github/workflows/pr-matrix.yaml
llvm-beanz: This seems reasonable. The Intel machine I have is old and slow, so doing less on it is probably good.

.github/workflows/windows-intel-dxc-warp-d3d12.yaml
llvm-beanz, line 1: Let's just delete these workflow files if we're not going to run them. We can always restore them from git if we need to.

llvm-beanz (approved):
All looks good once the bots go green.

PR #318, opened 2025-08-06
Add `dot` tests

### LINKED ISSUE #124: Add test for dot
https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-dot
https://github.com/llvm/llvm-project/blob/d5da826159bdae439c194f20ab2cdb63d9f52169/clang/lib/Headers/hlsl/hlsl_alias_intrinsics.h#L878

- [ ] Test for Double type (< intrinsic >.fp64.test) (only supports scalar double not vector)

A .test file should contain 
```
@@ -0,0 +1,54 @@
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# RUN: split-file %s %t
```

Tests for 16 bit floating point should have a `# REQUIRES: Half`. don't forget -enable-16bit-types
Tests for 16 bit integer should have a `# REQUIRES: Int16`. don't forget -enable-16bit-types

kmpeng:
Closes #124.

Adds tests for `dot` testing 16 bit int types, half, 32 bit types, 64 bit int types, and double (only supports scalar double not vector). Didn't use the [HLK test inputs](https://github.com/microsoft/DirectXShaderCompiler/blob/57177f77a4dc6996400ac97a0d618799c82374e8/tools/clang/unittests/HLSLExec/ShaderOpArithTable.xml#L4766) because it was testing 9 inputs for each of vector sizes 2-4, and the tests were mostly for nans, infs, and denorms.

### DIFF
--- a/test/Feature/HLSLLib/dot.fp16.test
+++ b/test/Feature/HLSLLib/dot.fp16.test
@@ -8,14 +8,16 @@ RWStructuredBuffer<half> Out : register(u2);
 [numthreads(1,1,1)]
 void main() {
   Out[0] = dot(X[0].x, Y[0].x);
-  Out[1] = dot(X[0].xy, Y[0].xy);
-  Out[2] = dot(X[0].xyz, Y[0].xyz);
-  Out[3] = dot(X[0], Y[0]);
+  Out[1] = dot(half(1.125), half(7.29));
 
-  Out[4] = dot(X[1].x, Y[1].x);
-  Out[5] = dot(X[1].xy, Y[1].xy);
-  Out[6] = dot(X[1].xyz, Y[1].xyz);
-  Out[7] = dot(X[1], Y[1]);
+  Out[2] = dot(X[0].xy, Y[0].xy);
+  Out[3] = dot(half2(1.125, -2.5), half2(7.29, 3.14));
+
+  Out[4] = dot(X[0].xyz, Y[0].xyz);
+  Out[5] = dot(half3(1.125, -2.5, -4.75), half3(7.29, 3.14, -1.1));
+  
+  Out[6] = dot(X[0], Y[0]);
+  Out[7] = dot(half4(1.125, -2.5, -4.75, 6.625), half4(7.29, 3.14, -1.1, -3.5));
 }
 //--- pipeline.yaml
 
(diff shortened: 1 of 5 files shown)


test/Feature/HLSLLib/dot.fp16.test
```diff
@@ -0,0 +1,86 @@
...
+    ZeroInitSize: 16
+  - Name: ExpectedOut
+    Format: Float16
+    Stride: 2
+    Data: [ 0x481a, 0x481a, 0x359f, 0x359f, 0x4594, 0x4594, 0xcc67, 0xcc67 ]
+    # [ 8.20125, 8.20125, 0.35125, 0.35125, 5.57625, 5.57625, -17.61125, -17.61125 ]
+Results:
+  - Result: Test1
+    Rule: BufferFloatEpsilon
+    Epsilon: 1
```
bob80905, line 52: I don't think you intended this epsilon to be this big. Perhaps you meant to use ULP?

test/Feature/HLSLLib/dot.fp64.test
```diff
@@ -0,0 +1,83 @@
...
+
+RWStructuredBuffer<double> Out : register(u2);
+
+
+[numthreads(1,1,1)]
+void main() {
+  Out[0] = dot(X[0], Y[0]);
+  Out[1] = dot(double(1.125), double(7.29));
+
+  Out[2] = dot(X[1], Y[1]);
```
bob80905, line 13: Is there a reason this test doesn't test the vectorized overload of dot?
How about `dot(double4(X[0], ... X[3]), ...);` ?

kmpeng: `dot` only supports [scalar double](https://github.com/llvm/llvm-project/blob/d5da826159bdae439c194f20ab2cdb63d9f52169/clang/lib/Headers/hlsl/hlsl_alias_intrinsics.h#L937) operations, it doesn't support vector

test/Feature/HLSLLib/dot.fp16.test
kmpeng: Oops I meant to have it as `0.008` like the other ones. Thanks for catching that!

kmpeng: Oops I meant to have it at the same value `0.008` as the other fp tests. Would that be too large though for float16?

bob80905 (approved):

test/Feature/HLSLLib/dot.fp16.test
bob80905: I think .008 is typical for fp16 tests.

bob80905 (changes requested):
Ah accident, I'll approve after the tolerance is updated

spall (approved):
lgtm

PR #343, opened 2025-08-10
[WARP] Make default WARP version "Last Known Good"

### LINKED ISSUE #340: Devise a scheme for managing WARP versions
The team recently has encountered test issues with varied WARP versions. Using `System Version` seems to be the wrong approach.

Offline discussions have mostly landed on the following proposal that needs to be finalized and staffed.

- Have a periodic cron job that runs with the latest version of warp to highlight incoming breakages
- Have a periodic cron job that runs with the last known good version of warp to ensure LKG stays good
- Have premerge tests run against LKG to minimize infra breakages for incoming PRs
- Use a cmake script to set/enforce LKG version for local development

llvm-beanz:
We've encountered a fair number of problems with the System as the default verison of WARP, instead using a Last Known Good will hopefully make it easier for contributors to get up and working.

This change adds two new special versions for the `WARP_VERSION` option: `LKG` and `Latest`. At the moment the `LKG` version is 1.0.14.2, which is also the latest released version.

This PR also adds a set of build jobs to run periodic builds with the current WARP pre-release build.

Fixes #340


llvm-beanz:
@damyanp and/or @Icohedron, could you give this a glance again? Getting it working required a bit more change than I expected.

For x86_64 we need to go back to WARP 1.0.10.1 as the LKG because I'm seeing some odd issues with the length test that looks like `sqrt` is being computed against the wrong SSA value. This complicates the CMake to fetch Warp because the older NuGet packages were structured differently.

For arm64 we set 1.0.14.2 as the LKG version because it passes all the tests that 1.0.10.1 does, and 1.0.10.1 doesn't have an arm64 binary in the NuGet package.

I've already talked with @jenatali about the firstbithigh issue that was fixed for x86 but persists on arm64. I'll also put together bugs on the WARP issues we're seeing here with the latest WARP running our length test and @kmpeng's PR (https://github.com/llvm/offload-test-suite/pull/316).

PR #347, opened 2025-08-12
[NFC] Simplify WaveActiveMax test

llvm-beanz:
This was a test I had written to cover some edge cases, and the test is a bit awkward and buggy. This corrects the test to hopefully be more consistent across vendors.


farzonl (approved):

test/Feature/WaveOps/EdgeConditions/lit.cfg.py
```diff
@@ -0,0 +1,2 @@
+if "EdgeConditions" in config.available_features:
+    config.unsupported = True
```
bogner: This logic is exactly backwards

PR #295, opened 2025-07-21
Add `WaveReadLaneAt` tests

inbelic:
This pr implements testing for the `WaveReadLaneAt` intrinsic.

- `WaveReadLaneAt.[16|32|64].test` adds testing of the basic types (and vectors)
- `WaveReadLaneAt.udt.test` adds testing for a matrix, user defined struct and for floating point edge-case values
- `WaveReadLaneAt.index.test` adds testing of various use-cases with different indices

Resolves: https://github.com/llvm/offload-test-suite/issues/144

PR #349, opened 2025-08-12
Add `WaveReadLaneAt` tests

bob80905:
This PR implements testing for the WaveReadLaneAt intrinsic. It's an improved copy of @inbelic's work here: https://github.com/llvm/offload-test-suite/pull/295

- WaveReadLaneAt.[16|32|64].test adds testing of the basic types (and vectors)
- WaveReadLaneAt.udt.test adds testing for a matrix, user defined struct and for floating point edge-case values
- WaveReadLaneAt.index.test adds testing of various use-cases with different indices

Resolves: https://github.com/llvm/offload-test-suite/issues/144

### DIFF
--- a/test/WaveOps/WaveReadLaneAt.64.test
+++ b/test/WaveOps/WaveReadLaneAt.64.test
@@ -1,0 +1,143 @@
+#--- source.hlsl
+
+StructuredBuffer<int64_t4> InInt : register(t0);
+RWStructuredBuffer<int64_t4> OutInt : register(u1);
+
+StructuredBuffer<uint64_t4> InUInt : register(t2);
+RWStructuredBuffer<uint64_t4> OutUInt : register(u3);
+
+StructuredBuffer<float64_t4> InFloat : register(t4);
+RWStructuredBuffer<float64_t4> OutFloat : register(u5);
+
+[numthreads(4,1,1)]
+void main(uint64_t3 TID : SV_GroupThreadID) {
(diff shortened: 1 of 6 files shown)


test/WaveOps/WaveReadLaneAt.64.test
```diff
@@ -0,0 +1,143 @@
...
+      Kind: RWStructuredBuffer
+      DirectXBinding:
+        Register: 5
+        Space: 0
+      VulkanBinding:
+        Binding: 5
+...
+#--- end
+
+# REQUIRES: Double, Int64
```
llvm-beanz, line 139: @bogner, I know we were talking about needing to double up testing, but I'm curious for your thoughts on this one.

Combining double and int64 in the same test means we don't run this on devices that only support one or the other. While I think int64 is fairly universal at this point because 64-bit pointer math kinda requires it, there are a number of mobile architectures that don't support double (including Apple GPUs).

Does it make sense to combine 64-bit types or should we split them?

inbelic (approved):
LGTM, just if we should split up the testcase as mentioned

test/WaveOps/WaveReadLaneAt.64.test
bogner: Yeah, splitting these makes sense. I suspect as a rule of thumb any time we're `REQUIRE`ing two different features in these tests we should probably split them up for maximum test coverage.

PR #364, opened 2025-08-19
Enable fixed-size resource array tests for Clang

hekota:
Since the support for fixed-size resource array has been implemented in Clang, these tests are now passing.

### DIFF
--- a/test/Feature/ResourceArrays/array-global.test
+++ b/test/Feature/ResourceArrays/array-global.test
@@ -118,10 +118,6 @@ DescriptorSets:
 ...
 #--- end
 
-# Resource arrays are not yet implemented in Clang:
-# https://github.com/llvm/llvm-project/issues/133835
-# XFAIL: Clang
-
 # Offload tests are missing support for resource arrays on Metal
 # https://github.com/llvm/offload-test-suite/issues/305
 # XFAIL: Metal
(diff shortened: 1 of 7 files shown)


inbelic (approved):

test/Feature/ResourceArrays/array-global.test
```diff
@@ -118,17 +118,16 @@ DescriptorSets:
...
 # https://github.com/llvm/offload-test-suite/issues/305
 # XFAIL: Metal
 
+# https://github.com/llvm/llvm-project/issues/149561
+# XFAIL: Clang-Vulkan && !VK_KHR_shader_float_controls2
+
 # RUN: split-file %s %t
 # RUN: %if !Vulkan %{ %dxc_target -T cs_6_0 -Fo %t.o %t/source.hlsl %}
-# RUN: %if Vulkan %{ %dxc_target -T cs_6_0 -fspv-target-env=vulkan1.3 -fvk-use-scalar-layout -Fo %t.o %t/source.hlsl %}
+# RUN: %if Vulkan %{ %dxc_target -T cs_6_0 -fspv-target-env=vulkan1.3 -Fo %t.o %t/source.hlsl %}
```
bogner, line 130: It might be good to mention in the commit message that the tests that are removing the scalar layout flag here weren't actually affected by it in the first place.

PR #366, opened 2025-08-20
[RootSignature] Add `obj2yaml` testing

inbelic:
This pr adds testing using `obj2yaml` for root signatures.

There are types of tests added:
- edge-cases: which check for edge cases as specified in the test cases
- consistency: which checks that `DXC` and `clang` set the root signature values to be the same. The particular values are not necessarily important, as they will have been covered in the edge-cases 

Resolves: https://github.com/llvm/offload-test-suite/issues/260.

### DIFF
--- a/test/Feature/RootSignatures/RootDescriptorAndTables.test
+++ b/test/Feature/RootSignatures/RootDescriptorAndTables.test
@@ -83,7 +83,7 @@ DescriptorSets:
 # OBJ-NEXT:       Descriptor:
 # OBJ-NEXT:       RegisterSpace:   0
 # OBJ-NEXT:       ShaderRegister:  2
-# OBJ-NEXT:     - ParameterType:   0
+# OBJ:          - ParameterType:   0
 # OBJ-NEXT:       ShaderVisibility: 0
 # OBJ-NEXT:       Table:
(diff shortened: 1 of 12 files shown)


llvm-beanz (approved):
This LGTM. I wonder if longer term we would be better off testing this with a more compact text representation supported via `llvm-objdump` (https://github.com/llvm/llvm-project/issues/152526).

test/Feature/RootSignatures/RootDescriptorAndTables.test
```diff
@@ -64,6 +64,38 @@ DescriptorSets:
...
+# OBJ:   RootParametersOffset: 24
+# OBJ:   NumStaticSamplers: 0
+# OBJ:   StaticSamplersOffset: 116
+# OBJ:   Parameters:
+# OBJ: - ParameterType:   2
+# OBJ:   ShaderVisibility: 0
+# OBJ:   Descriptor:
+# OBJ:   RegisterSpace:   0
+# OBJ:   ShaderRegister:  2
+# OBJ: - ParameterType:   0
```
bogner, line 86: Looks like we aren't checking everything (hence not using CHECK-NEXT I suppose), which is fine, but how did you decide which fields are interesting?

PR #368, opened 2025-08-20
Add `any` tests

### LINKED ISSUE #131: Add test for any
https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-any
https://github.com/llvm/llvm-project/blob/b474c3f69eae4230fc9a2eff07ba63d2bbac65c0/clang/lib/Headers/hlsl/hlsl_alias_intrinsics.h#L315

- [ ] make sure to test for all vector sizes

A .test file should contain 
```
@@ -0,0 +1,54 @@
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# RUN: split-file %s %t
```
Tests for 16 bit floating point should have a `# REQUIRES: Half`. Don't forget -enable-16bit-types
Tests for 16 bit integer should have a `# REQUIRES: Int16`. Don't forget -enable-16bit-types

kmpeng:
Closes #131.

Adds tests for `any` testing 16 bit int types, half, 32 bit types, 64 bit int types, and double.

I moved the bool tests from `any.32.test` to a separate file because [this bug](https://github.com/llvm/llvm-project/issues/140824) is causing the vec1 test to fail on Clang, and I didn't want to XFAIL the rest of the 32 bit tests over it. We can combine them in the future once the bug is fixed.
Also have XFAILs for this Intel driver bug: #370.


test/Feature/HLSLLib/any.fp64.test
```diff
@@ -0,0 +1,64 @@
...
+    - Name: Out
+      Kind: RWStructuredBuffer
+      DirectXBinding:
+        Register: 1
+        Space: 0
+      VulkanBinding:
+        Binding: 1
+#--- end
+
+# Bug https://github.com/llvm/offload-test-suite/issues/370 XFAIL: DXC && DirectX-Intel
```
bogner, line 59: I think it's clearer to split this (and the other similar one) into two lines. That said, I'll defer to @spall if there's disagreement here as she's working on getting us to be consistent about how we do this.
```suggestion
# Bug https://github.com/llvm/offload-test-suite/issues/370
# XFAIL: DXC && DirectX-Intel
```

bogner (approved):

PR #379, opened 2025-08-25
[VK] Remove XFAIL from passing tests

llvm-beanz:
These tests are now passing, so we can remove the XFAIL annotations.


llvm-beanz:
> I'm also seeing that `test/WaveOps/WaveReadLaneAt.index.test` is XPASSing for clang-vulkan

I have that one fixed in this PR: https://github.com/llvm/offload-test-suite/pull/373

bogner:
I believe that these tests still fail the spirv validator as per https://github.com/llvm/llvm-project/issues/145513#issuecomment-3162542505 - are they actually passing on all of the bots?

PR #380, opened 2025-08-25
Speed up the dot2add test with some parallelism

bogner:
This test was running extremely slowly under warp for me, with the dxc version taking 15-20s and the clang version taking around 45-50s. Some parallelism reduces these both to under 2s on my machine, without really affecting the timing of any of the actual GPU versions of the test.

Shaves nearly 30s off of `ninja check-hlsl` locally.


damyanp:
While this change seems like a good one, these speed ups make no sense just looking at this change in isolation.  It would be good to get a proper understanding of what causes the difference in timing here.

llvm-beanz (approved):

kmpeng (approved):

tex3d (approved):

PR #382, opened 2025-08-26
Add flags to control debug and validation layers

llvm-beanz:
This PR adds flags to enable the debug and validation layers for runtimes. Both default to `off` in the offloader tool, but the default CMake configuration enables the debug layer during test suite runs.

### DIFF
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -30,6 +30,11 @@ namespace offloadtest {
 
 struct Pipeline;
 
+struct DeviceConfig {
+  bool EnableDebugLayer = false;
+  bool EnableValidationLayer = false;
+};
+
 class Device {
 protected:
   std::string Description;
@@ -46,7 +51,7 @@ public:
   llvm::StringRef getDescription() const { return Description; }
 
   static void registerDevice(std::shared_ptr<Device> D);
-  static llvm::Error initialize();
+  static llvm::Error initialize(const DeviceConfig Config);
   static void uninitialize();
 
   using DeviceArray = llvm::SmallVector<std::shared_ptr<Device>>;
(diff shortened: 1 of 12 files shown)


include/API/Device.h
```diff
@@ -46,7 +51,7 @@ class Device {
   llvm::StringRef getDescription() const { return Description; }
 
   static void registerDevice(std::shared_ptr<Device> D);
-  static llvm::Error initialize();
+  static llvm::Error initialize(const DeviceConfig Config);
```
damyanp, line 54: Isn't this `const` redundant / ignored?

lib/API/DX/Device.cpp
```diff
@@ -1169,18 +1169,19 @@ class DXDevice : public offloadtest::Device {
...
-    return Err;
+    if (auto Err = HR::toError(D3D12GetDebugInterface(IID_PPV_ARGS(&Debug1)),
+                               "failed to create D3D12 Debug Interface"))
+      return Err;
 
-  Debug1->EnableDebugLayer();
-  Debug1->SetEnableGPUBasedValidation(true);
-#endif
+    if (Config.EnableDebugLayer)
+      Debug1->EnableDebugLayer();
```
damyanp, line 1182: Is it intentional that just setting EnableValidationLayer wouldn't actually enable the validation layer.  Both flags would need to be set.

bogner (approved):
Generally looks good, though it looks like an unrelated change slipped in.

damyanp (approved):

include/API/Device.h
llvm-beanz: You can modify a pass-by-value variable, it just modifies the instance.

lib/API/DX/Device.cpp
llvm-beanz: Does GBV require the debug layer? I assumed they were separate runtime settings.

lib/API/VK/Device.cpp
```diff
@@ -330,10 +330,13 @@ class VKDevice : public offloadtest::Device {
...
-    Features11.pNext = &Features12;
-    Features12.pNext = &Features13;
+    if (Props.apiVersion >= VK_MAKE_API_VERSION(0, 1, 2, 0))
+      Features11.pNext = &Features12;
+    if (Props.apiVersion >= VK_MAKE_API_VERSION(0, 1, 3, 0))
+      Features12.pNext = &Features13;
 #ifdef VK_VERSION_1_4
-    Features13.pNext = &Features14;
+    if (Props.apiVersion >= VK_MAKE_API_VERSION(0, 1, 4, 0))
+      Features13.pNext = &Features14;
```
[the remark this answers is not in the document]
llvm-beanz: Ooops. Pulled that out over here: https://github.com/llvm/offload-test-suite/pull/383

lib/API/DX/Device.cpp
llvm-beanz: I think the docs indicate that you do need the debug layer for GBV, so I've updated the code.

PR #388, opened 2025-08-27
Add `select` (basic types) tests

### LINKED ISSUE #132: Add test for select

A .test file should contain 
```
@@ -0,0 +1,54 @@
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# UNSUPPORTED: Clang-Vulkan
# RUN: split-file %s %t
```

kmpeng:
Closes #132.

Adds tests for `select` (basic types) testing 16 bit int types, half, 32 bit types, 64 bit int types, and double.

For the `Cond` buffers (and `TrueVal`, `FalseVal` for the bool test), I had to use `bool` instead of `bool4` because select gives an "Invalid overload type" error if it receives a vec1 bool instead of a scalar as input.

### DIFF
--- a/test/Feature/HLSLLib/select.int16.test
+++ b/test/Feature/HLSLLib/select.int16.test
@@ -1,0 +1,166 @@
+#--- source.hlsl
+StructuredBuffer<bool> Cond : register(t0);
+StructuredBuffer<int16_t4> TrueVal0 : register(t1);
+StructuredBuffer<int16_t4> FalseVal0 : register(t2);
+StructuredBuffer<uint16_t4> TrueVal1 : register(t3);
+StructuredBuffer<uint16_t4> FalseVal1 : register(t4);
+
+RWStructuredBuffer<int16_t4> Out0 : register(u5);
+RWStructuredBuffer<uint16_t4> Out1 : register(u6);
+
+
+[numthreads(1,1,1)]
+void main() {
+  bool4 Cond0 = bool4(Cond[0], Cond[1], Cond[2], Cond[3]);
+  bool3 Cond1 = bool3(Cond[4], Cond[5], Cond[6]);
+  bool2 Cond2 = bool2(Cond[8], Cond[9]);
+  bool2 Cond3 = bool2(Cond[10], Cond[11]);
+
+  // int16_t
+  // vec4
+  Out0[0] = select(Cond0, TrueVal0[0], FalseVal0[0]);
+  Out0[1] = select(Cond0, TrueVal0[0].x, FalseVal0[0]);
+  Out0[2] = select(Cond0, TrueVal0[0], FalseVal0[0].x);
+  Out0[3] = select(Cond0, TrueVal0[0].x, FalseVal0[0].x);
+  // vec3 + scalar
+  Out0[4] = int16_t4(select(Cond1, TrueVal0[1].xyz, FalseVal0[1].xyz), select(Cond[7], TrueVal0[1].w, FalseVal0[1].w));
+  Out0[5] = int16_t4(select(Cond1, TrueVal0[1].x, FalseVal0[1].xyz), select(bool(1), int16_t(1), int16_t(-1)));
+  Out0[6] = int16_t4(select(Cond1, TrueVal0[1].xyz, FalseVal0[1].x), select(bool(0), int16_t(2), int16_t(-2)));
+  Out0[7] = int16_t4(select(Cond1, TrueVal0[1].x, FalseVal0[1].x), select(bool(1), int16_t(3), int16_t(-3)));
+  // vec2
+  Out0[8] = int16_t4(select(Cond2, TrueVal0[2].xy, FalseVal0[2].xy), select(Cond3, TrueVal0[2].z, FalseVal0[2].zw));
+  Out0[9] = int16_t4(select(Cond2, TrueVal0[2].xy, FalseVal0[2].x), select(Cond3, TrueVal0[2].z, FalseVal0[2].z));
+
+  // uint16_t
+  // vec4
+  Out1[0] = select(Cond0, TrueVal1[0], FalseVal1[0]);
+  Out1[1] = select(Cond0, TrueVal1[0].x, FalseVal1[0]);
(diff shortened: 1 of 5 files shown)


kmpeng:
If anyone has thoughts on how to better organize these tests, please let me know!

Icohedron (commented):
Should there be scalar tests for select as well?

Icohedron (approved):

test/Feature/HLSLLib/select.int16.test
```diff
@@ -0,0 +1,166 @@
...
+  Out0[1] = select(Cond0, TrueVal0[0].x, FalseVal0[0]);
+  Out0[2] = select(Cond0, TrueVal0[0], FalseVal0[0].x);
+  Out0[3] = select(Cond0, TrueVal0[0].x, FalseVal0[0].x);
+  // vec3 + scalar
+  Out0[4] = int16_t4(select(Cond1, TrueVal0[1].xyz, FalseVal0[1].xyz), select(Cond[7], TrueVal0[1].w, FalseVal0[1].w));
+  Out0[5] = int16_t4(select(Cond1, TrueVal0[1].x, FalseVal0[1].xyz), select(bool(1), int16_t(1), int16_t(-1)));
+  Out0[6] = int16_t4(select(Cond1, TrueVal0[1].xyz, FalseVal0[1].x), select(bool(0), int16_t(2), int16_t(-2)));
+  Out0[7] = int16_t4(select(Cond1, TrueVal0[1].x, FalseVal0[1].x), select(bool(1), int16_t(3), int16_t(-3)));
+  // vec2
+  Out0[8] = int16_t4(select(Cond2, TrueVal0[2].xy, FalseVal0[2].xy), select(Cond3, TrueVal0[2].z, FalseVal0[2].zw));
```
alsepkow, line 31: Was the second select call intended to have TrueVal[2].z or should they both be TrueVal[2].zw?

test/Feature/HLSLLib/select.int16.test
```diff
@@ -0,0 +1,166 @@
...
+StructuredBuffer<uint16_t4> TrueVal1 : register(t3);
+StructuredBuffer<uint16_t4> FalseVal1 : register(t4);
+
+RWStructuredBuffer<int16_t4> Out0 : register(u5);
+RWStructuredBuffer<uint16_t4> Out1 : register(u6);
+
+
+[numthreads(1,1,1)]
+void main() {
+  bool4 Cond0 = bool4(Cond[0], Cond[1], Cond[2], Cond[3]);
```
alsepkow, line 14: nit: I'm wondering if it would improve readability of these tests to get rid of 'StructuredBuffer<bool> Cond : register(t0);' and instead use constants.

`bool4 Cond0 = bool4(Cond[0], Cond[1], Cond[2], Cond[3]);`
Would become 
`bool4 Cond0 = bool4(true, false, true, false);`

test/Feature/HLSLLib/select.int16.test
```diff
@@ -0,0 +1,166 @@
+#--- source.hlsl
```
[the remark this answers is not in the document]
kmpeng: All these intrinsic tests have been organized so the int16 and int64 tests are split into different files. I'm not too sure what the reason is for that though - maybe because of the `REQUIRES: Int16` and `REQUIRES: Int64` lines in the tests?

Either way, feels like a larger question about the overall design of the intrinsic functional tests, so I'll defer to @spall

test/Feature/HLSLLib/select.int16.test
kmpeng: I believe we wanted to write most of the intrinsic tests with buffers, and only have a few that use constants (e.g. [lines 27-29](https://github.com/llvm/offload-test-suite/blob/5e35b6465026e6cc8c0bec903bb73e0be67270d1/test/Feature/HLSLLib/select.int16.test#L27-L29)). Again, not really sure of the reason so tagging @spall to see if this would be a valid change

test/Feature/HLSLLib/select.int16.test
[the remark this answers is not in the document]
spall: Theoretically the source portion of the test file could be shared, but the larger portion of most tests is the yaml and run lines which would not be shared. Doesn't seem worth it.

bob80905 (approved):
LGTM

test/Feature/HLSLLib/select.int16.test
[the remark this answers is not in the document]
alsepkow: I wouldn't block the review on it. But it just seems like something that might be worth thinking about in general. These all seem ripe for copy/paste which just feels problematic and a little mistake prone. But maybe you could make similar arguments for the potential added complexity of trying to remove duplications.

test/Feature/HLSLLib/select.int16.test
alsepkow: Just a nit, so don't block on this. But if part of our goal is to test with buffers then maybe some buffers with specific conditional values would help make this more readable? i.e a buffer that only has true values with a name that reflects that. 

I just found it confusing to try and reason through what the actual result of each expression was going to be. Which then made it harder to reason about coverage.

alsepkow:
LGTM.

PR #389, opened 2025-08-28
Add GitHub actions for arm64 and qualcomm

llvm-beanz:
This adds some actions to run scheduled and "extra" PR builds on newly allocated qualcomm hardware.


.github/workflows/windows-qc-dxc-d3d12.yaml
```diff
@@ -0,0 +1,21 @@
+name: Windows D3D12 AMD DXC
+
+permissions:
+  contents: read
+  checks: write
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 * * * *' # run every 30 minutes
```
hekota, line 10: Why is this one 30 minutes and the others 2 hours?

hekota (approved):
LGTM, just one question.

.github/workflows/windows-qc-dxc-d3d12.yaml
llvm-beanz: The comment is wrong here because this is actually every hour... but it also shouldn't really be different (at least not initially). So I'm going to change this to every 2 hours as well and we can see where the bot utilization is.

PR #392, opened 2025-08-28
Enable passing vulkan tests

s-perron:
Enable vulkan tests that are now able to pass using clang. Some of the
tests require small updates in the options used.


s-perron:
@spall FYI

PR #395, opened 2025-08-29
Fix the arm64 / Qualcomm workflows so they actually run on arm64 / Qualcomm

Icohedron:
Perhaps the last PR (#389) putting these up was too hastily made and reviewed!

### DIFF
--- a/.github/workflows/windows-qc-clang-warp-d3d12.yaml
+++ b/.github/workflows/windows-qc-clang-warp-d3d12.yaml
@@ -1,4 +1,4 @@
-name: Windows D3D12 Warp Clang
+name: Windows ARM64 D3D12 Warp Clang
 
 permissions:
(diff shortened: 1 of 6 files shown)


spall (approved):
why do some of the names mention QC and some mention ARM64, but theyre all running on the same machine?

.github/workflows/windows-qc-clang-warp-d3d12.yaml
```diff
@@ -1,4 +1,4 @@
-name: Windows D3D12 Warp Clang
+name: Windows ARM64 D3D12 Warp Clang
```
bob80905, line 1: For what reason does ARM64 come into this?

bob80905 (approved):

llvm-beanz (approved):

PR #422, opened 2025-09-02
[Docs] Support tiers and status readability

llvm-beanz:
This PR updates the README to try and improve the readability of the status block. It also sorts the builds into support tiers so that we can more clearly explain what is PR blocking and what is a priority to fix immediately.


Icohedron (approved):
I'm not sure I like the look of having the "Tier X Targets" labels being their own rows in the table. I prefer having separate tables for each tier.  
Otherwise, the changes LGTM.

Icohedron:
Also, once all targets become Tier 1, wouldn't we be back to the same cluttered table we had before? XD

docs/CI.md
```diff
@@ -0,0 +1,80 @@
...
+Because these configurations are less stable and rely more heavily on tooling
+outside our direct control PRs may liberally `XFAIL` new tests that only fail on
+Tier 2 configurations.
+
+Configurations:
+* macOS Metal
+
+### Experimental Configurfations
+
+Experimental configuraitons are new testing configurations that have been
```
damyanp, line 34:
```suggestion
Experimental configurations are testing configurations that have been
```

docs/CI.md
```diff
@@ -0,0 +1,80 @@
...
+* GPU: Intel UHD Graphics 630
+* RAM: 64 GiB
+
+### Windows NVIDIA GPU
+
+* CPU: Intel(R) Core(TM) i5-14400F
+* GPU: NVIDIA GeForce RTX 5070
+* RAM: 16 GiB
+
+### Windows Qualcomm GPU
```
damyanp, line 76: I don't think that there's anything explicitly mapping this test machine to the ARM64 WARP tests?

damyanp (approved):

PR #423, opened 2025-09-02
Don't use switch fallthrough

llvm-beanz:
In SPIRV even trivial switch fallthrough is undefined behavior, so we shouldn't rely on it in wave operations tests.

### DIFF
--- a/test/Feature/WaveOps/WaveActiveCountBits.test
+++ b/test/Feature/WaveOps/WaveActiveCountBits.test
@@ -6,13 +6,16 @@ RWStructuredBuffer<uint> Out : register(u1);
 void main(uint3 threadID : SV_DispatchThreadID) {
   bool B1 = false;
 
-  if (value[threadID.x] == 0)
-    Out[threadID.x + 4] = WaveActiveCountBits(true); // threads 0 and 1
+  if (value[threadID.x] == 0) // thread 0 and 1.
+    Out[threadID.x + 4] = WaveActiveCountBits(true);
+  // thread 0-3 set b1 to true.
   if (value[threadID.x] == 0 || value[threadID.x] == 1)
-    B1 = true; // set b1 to true for thread 3
-  else
-    Out[threadID.x + 4] = WaveActiveCountBits(false); // thread 2; expect 0
-  // should be 3 because B1 set to true for threads 0,1, and 3.
+    B1 = true;
+  else // thread 4
+    Out[threadID.x + 4] = WaveActiveCountBits(false);
+
+  // Out[threadID.x + 4] on thread 3 is never written to and should remain 0.
+  // All threads count 3 because B1 set to true for threads 0,1, and 3.
   uint Count = WaveActiveCountBits(B1);
   Out[threadID.x] = Count;
 }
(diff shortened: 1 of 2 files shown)


damyanp (approved):
LGTM, although arguably the question should be whether or not it is UB in HLSL, and if it isn't then the SPIR-V backend needs to be updated to handle this.

test/Feature/WaveOps/WaveActiveCountBits.test
```diff
@@ -6,16 +6,12 @@ RWStructuredBuffer<uint> Out : register(u1);
...
-      B1 = true; // set b1 to true for thread 3
-      break;
-    default:
-      Out[threadID.x + 4] = WaveActiveCountBits(false); // thread 2; expect 0
-      break;
-  }
+  if (value[threadID.x] == 0)
+    Out[threadID.x + 4] = WaveActiveCountBits(true); // threads 0 and 1
+  if (value[threadID.x] == 0 || value[threadID.x] == 1)
+    B1 = true; // set b1 to true for thread 3
```
tex3d, line 12: This sets `B1` to true for threads 0, 1, and 2, but not 3, due to LUT `value` being `[0, 0, 1, 2]`.

test/Feature/WaveOps/WaveActiveCountBits.test
```diff
@@ -6,16 +6,12 @@ RWStructuredBuffer<uint> Out : register(u1);
...
+  else
+    Out[threadID.x + 4] = WaveActiveCountBits(false); // thread 2; expect 0
```
tex3d, line 14: This writes output for thread 3, since it's the only thread not going through the main `if` branch.

Incidentally, the expected outputs are also truncated, excluding this output, while it includes an expected `0` output for thread 2, which was never written from the shader (which passes due to zero-initialization of output buffer).

tex3d:
@damyanp said:
> LGTM, although arguably the question should be whether or not it is UB in HLSL, and if it isn't then the SPIR-V backend needs to be updated to handle this.

Trivial fall-through was supposed to be legal and well-defined in HLSL. Non-trivial fall-through was always supposed to be illegal in HLSL, but was never blocked by DXC, unfortunately.

test/Feature/WaveOps/WaveActiveCountBits.test
```diff
@@ -6,16 +6,12 @@ RWStructuredBuffer<uint> Out : register(u1);
...
-    case 2:
-      B1 = true; // set b1 to true for thread 3
-      break;
-    default:
-      Out[threadID.x + 4] = WaveActiveCountBits(false); // thread 2; expect 0
-      break;
-  }
+  if (value[threadID.x] == 0)
+    Out[threadID.x + 4] = WaveActiveCountBits(true); // threads 0 and 1
+  if (value[threadID.x] == 0 || value[threadID.x] == 1)
```
tex3d, line 12: Note: This isn't equivalent to the original behavior. `== 1` should be `== 2` for equivalent behavior.

tex3d (commented):
In `test/Feature/WaveOps/WaveActiveCountBits.test`, my comments about the comments being different than behavior would be fixed with a fix to the second if condition, from:
```cpp
  if (value[threadID.x] == 0 || value[threadID.x] == 1)
```
to:
```cpp
  if (value[threadID.x] == 0 || value[threadID.x] == 2)
```
Which would match the original shader.

I see changes to address my other comments, but they aren't quite right now either. For instance, `// thread 0-3 set b1 to true.` and `// thread 4` aren't right. Probably better to restore original comments and fix the above condition.

farzonl (approved):

PR #425, opened 2025-09-03
[VK] Surface validation errors as errors

llvm-beanz:
This will force tests to fail when validation errors occur so that they fail everywhere instead of just on some devices.

### DIFF
--- a/lib/API/VK/Device.cpp
+++ b/lib/API/VK/Device.cpp
@@ -114,55 +114,9 @@ static std::string getMessageSeverityString(
 static VkBool32
 debugCallback(VkDebugUtilsMessageSeverityFlagBitsEXT MessageSeverity,
               VkDebugUtilsMessageTypeFlagsEXT MessageType,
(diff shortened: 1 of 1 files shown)


lib/API/VK/Device.cpp
```diff
@@ -1346,3 +1272,136 @@ llvm::Error Device::initializeVKDevices(const DeviceConfig Config) {
...
+
+  std::error_code convertToErrorCode() const override {
+    return llvm::inconvertibleErrorCode();
+  }
+
+private:
+  VkDebugUtilsMessageSeverityFlagBitsEXT MessageSeverity;
+  std::string Message;
+  std::string MessageID;
+  std::vector<std::string> Objects;
```
bogner: SmallVector would be slightly more efficient here (it doesn't really matter though)

bogner (approved):
Looks good. A few nitpicks and take-it-or-leave-it comments inline.

PR #429, opened 2025-09-05
Add WaveActiveMax tests

bob80905:
Adds WaveActiveMax  tests.
Fixes https://github.com/llvm/offload-test-suite/issues/122

### DIFF
--- a/test/WaveOps/WaveActiveMax.fp16.test
+++ b/test/WaveOps/WaveActiveMax.fp16.test
@@ -93,23 +93,23 @@ Buffers:
     Format: Float16
     Stride: 2
     # 1 half is 2 bytes, * 4 halves for 4 threads, * 4 thread masks, * 2 value sets
-    ZeroInitSize: 64  
+    FillSize: 64  
   - Name: Out2
     Format: Float16
     Stride: 4
-    ZeroInitSize: 128
+    FillSize: 128
   - Name: Out3
     Format: Float16
     Stride: 8
-    ZeroInitSize: 256
+    FillSize: 256
   - Name: Out4
     Format: Float16
     Stride: 8
-    ZeroInitSize: 256
+    FillSize: 256
   - Name: Out5
     Format: Float16
     Stride: 8
-    ZeroInitSize: 8
+    FillSize: 8
   - Name: Masks
     Format: Int32
     Stride: 16
--- a/test/WaveOps/WaveActiveMax.fp32.test
+++ b/test/WaveOps/WaveActiveMax.fp32.test
@@ -93,23 +93,23 @@ Buffers:
     Format: Float32
     Stride: 4
     # 1 float is 4 bytes, * 4 halves for 4 threads, * 4 thread masks, * 2 value sets
-    ZeroInitSize: 128  
+    FillSize: 128  
   - Name: Out2
     Format: Float32
     Stride: 8
-    ZeroInitSize: 256
+    FillSize: 256
   - Name: Out3
     Format: Float32
     Stride: 16
-    ZeroInitSize: 512
+    FillSize: 512
   - Name: Out4
     Format: Float32
     Stride: 16
-    ZeroInitSize: 512
+    FillSize: 512
   - Name: Out5
     Format: Float32
     Stride: 16
-    ZeroInitSize: 16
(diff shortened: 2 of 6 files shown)


test/WaveOps/WaveActiveMax.fp16.test
```diff
@@ -0,0 +1,178 @@
...
+---
+Shaders:
+  - Stage: Compute
+    Entry: main
+    DispatchSize: [1, 1, 1]
+Buffers:
+  - Name: In
+    Format: Float16
+    Stride: 8
+    # 1, 10, 100, 1000, 2, 20, 200, 2000, 3, 30, 300, 3000, 4, 40, 400, 4000
```
tex3d, line 59: All tests appear to use the same whole-number test values, which always increase in value for higher thread ids. I feel like this could miss implementation errors like:
- implicit casting for wave op to int (no fractional values)
- implicit casting for wave op to different bit-size (no values requiring selected bit size to accurately express)
- not handling/preserving denorms when required (for half and double, as well as float when denorm mode is preserve)
- mishandling of negative values
- mishandling of INF/-INF
  - I believe inf/-inf should be reliably handled for this op, but could be wrong
- just returning the value from the highest active thread index (bad implementation)

tex3d: Additional note: the comment with the values is helpful, but would be even more helpful if formatted so you could line up values compared across threads, like:
```
    # 1, 10, 100, 1000, # thread 0
    # 2, 20, 200, 2000, # thread 1
    # 3, 30, 300, 3000, # thread 2
    # 4, 40, 400, 4000  # thread 3
```

tex3d: Additionally, more sets of values could be tested with an outer loop around the value set, if that's desired given my feedback on the limitations of this chosen set of values.

tex3d: Another option for selecting sets of threads could have been an input mask input set instead of always using max of values from threads [0,n] for each n=[0,3]. That could look like:
```
    # 13 active mask sets for threads 0, 1, 2, 3:
    # 1 1 1 1
    # 1 0 0 0
    # 0 1 0 0
    # 0 0 1 0
    # 0 0 0 1
    # 0 1 1 1
    # 1 0 1 1
    # 1 1 0 1
    # 1 1 1 0
    # 1 1 0 0
    # 0 0 1 1
    # 0 1 1 0
    # 1 0 0 1
```

An updated shader that could work with this and multiple value sets:
```hlsl
#define VALUE_SETS 2
#define NUM_MASKS 13
#define NUM_THREADS 4

struct MaskStruct {
    int mask[4];
};

StructuredBuffer<half4> In  : register(t0);
StructuredBuffer<MaskStruct> Masks  : register(t1);
RWStructuredBuffer<half4> Out1 : register(u2); // test scalar
RWStructuredBuffer<half4> Out2 : register(u3); // test half2
RWStructuredBuffer<half4> Out3 : register(u4); // test half3
RWStructuredBuffer<half4> Out4 : register(u5); // test half4
RWStructuredBuffer<half4> Out5 : register(u6); // constant folding

[numthreads(NUM_THREADS,1,1)]
void main(uint3 tid : SV_GroupThreadID)
{
    for (int ValueSet = 0; ValueSet < VALUE_SETS; ValueSet++) {
        const uint ValueSetOffset = ValueSet * NUM_MASKS * NUM_THREADS;
        half4 v = In[ValueSet * NUM_THREADS + tid.x];
        for (int MaskIdx = 0; MaskIdx < NUM_MASKS; MaskIdx++) {
            const uint OutIdx = ValueSetOffset + MaskIdx * NUM_THREADS + tid.x;
            if (Masks[MaskIdx].mask[tid.x]) {
                Out1[OutIdx].x = WaveActiveMax( v.x );
                Out2[OutIdx].xy = WaveActiveMax( v.xy );
                Out3[OutIdx].xyz = WaveActiveMax( v.xyz );
                Out4[OutIdx] = WaveActiveMax( v );
            }
        }
    }

    // constant folding case
    Out5[0] = WaveActiveMax(half4(1,2,3,4));
}
```

spall (approved):

PR #432, opened 2025-09-08
Add WaveReadLaneFirst tests

bob80905:
Adds WaveReadLaneFirst tests.
Fixes https://github.com/llvm/offload-test-suite/issues/91

### DIFF
--- a/test/WaveOps/WaveReadLaneFirst.128Threads.test
+++ b/test/WaveOps/WaveReadLaneFirst.128Threads.test
@@ -16,12 +16,10 @@ StructuredBuffer<MaskStruct> Masks  : register(t6);
 void main(uint3 tid : SV_GroupThreadID)
 {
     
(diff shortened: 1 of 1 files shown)


damyanp:
There's some feedback that @tex3d is driving on #429 that applies pretty much to this one as well, so we should probably bottom out on that and see how it applies here.

inbelic (commented):
It seems to me that we should also have tests where `NUM_THREADS` is a multiple of `WaveSize` (say 128 threads and a wave size of 32) to ensure that it runs as expected across multiple waves.

The currently added tests are consistent with `WaveActiveMax`, so that looks good to me

inbelic (commented):
Just a couple small nits

inbelic (approved):

bob80905:
> LGTM...
> 
> But, man do I long for a more dynamic testing framework that could support parameterized scenarios, generated inputs, and computed expected results.

Big agree over here.

PR #441, opened 2025-09-15
Add Graphics support to DX, VK and MTL

llvm-beanz:
This PR adds support for simple graphics pipelines comprised of a vertex shader and a pixel shader.

The added test case generates an image added in the golden image suite as https://github.com/llvm/offload-golden-images/pull/1.

Additions to the YAML pipeline description include a new Binding structure which enables specifying a vertex buffer and associated vertex attributes as well as a buffer to serve as the render target.

The shader stage enumeration has been extended to support `Vertex` and `Pixel` stages (`Pixel` mapping to "Fragment" in the MTL and VK APIs).

### DIFF
--- a/test/Graphics/SimpleTriangle.test
+++ b/test/Graphics/SimpleTriangle.test
@@ -1,0 +1,80 @@
+#--- vertex.hlsl
+struct PSInput
+{
+    float4 position : SV_POSITION;
+    float4 color : COLOR;
+};
+
+PSInput main(float4 position : POSITION, float4 color : COLOR)
+{
+    PSInput result;
+
+    result.position = position;
+    result.color = color;
+
+    return result;
+}
+
+
+#--- pixel.hlsl
+struct PSInput
+{
+    float4 position : SV_POSITION;
+    float4 color : COLOR;
+};
+
+float4 main(PSInput input) : SV_TARGET
+{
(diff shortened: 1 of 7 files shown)


test/Graphics/SimpleTriangle.test
```diff
@@ -0,0 +1,80 @@
...
+#--- end
+
+# UNSUPPORTED: Clang
+# REQUIRES: goldenimage
+
+# RUN: split-file %s %t
+# RUN: %dxc_target -T vs_6_0 -Fo %t-vertex.o %t/vertex.hlsl
+# RUN: %dxc_target -T ps_6_0 -Fo %t-pixel.o %t/pixel.hlsl
+# RUN: %offloader %t/pipeline.yaml %t-vertex.o %t-pixel.o -r Output -o %t/Output.png
+# RUN: imgdiff %t/Output.png %goldenimage_dir/hlsl/Graphics/SimpleTriangle.png -rules %t/rules.yaml
```
farzonl, line 80: is this going to be a reliable way to check differences across dxc vs clang-dxc and DirectX vs SPIRV? I feel like there could be some visually imperseptable differences caused by rounding and our diffing might need a fuzzy match or atleast seperate golden images per api.

Issue #305, opened 2025-07-31
[Metal] Add support for resource arrays for Metal


hekota:
We need to add support for resource arrays for Metal.
DirectX and Vulkan support is tracked here: https://github.com/llvm/offload-test-suite/pull/302

PR #456, opened 2025-09-29
Add more test for resource arrays without NonUniformResourceIndex

hekota:


test/Feature/ResourceArrays/multi-dim-array-subset.test
```diff
@@ -0,0 +1,78 @@
...
+        Register: 0
+        Space: 1
+...
+#--- end
+
+# Vulkan does not support multi-dimensional resource arrays
+# UNSUPPORTED: Vulkan
+
+# Resource arrays are not yet supported on Metal
+# UNSUPPORTED: Metal
```
bogner: Should be an XFAIL (Linking to https://github.com/llvm/offload-test-suite/issues/305 I think?)

test/Feature/ResourceArrays/multi-dim-array-subset.test
```diff
@@ -0,0 +1,78 @@
...
+    - Name: BufOut
+      Kind: RWStructuredBuffer
+      DirectXBinding:
+        Register: 0
+        Space: 1
+...
+#--- end
+
+# Vulkan does not support multi-dimensional resource arrays
+# UNSUPPORTED: Vulkan
```
bogner: Can we split this into two XFAILs, one for `DXC && Vulkan` and one for `Clang && Vulkan`? I'm not sure that there's an appropriate dxc task to link to, but I think we do expect to make this work in clang.

PR #461, opened 2025-10-01
Vulkan: Add support for counter buffers

s-perron:
This commit introduces support for counter buffers in the Vulkan backend.

The changes include:
- Extending the VulkanBinding struct to include an optional CounterBinding.
- Implementing the necessary logic in the Vulkan device code to handle the creation, management, and cleanup of counter buffers.
- Updating the descriptor set layout and write processes to correctly handle counter buffers.
- Enabling the inc_counter.test for Vulkan, as it is now expected to pass.
- Adding a new test case to verify that an error is correctly reported when a resource with a counter is missing the required CounterBinding.
- Updating inc_counter_array.test to use only uniform resource indices.

### DIFF
--- a/include/Support/Pipeline.h
+++ b/include/Support/Pipeline.h
@@ -65,6 +65,7 @@ struct DirectXBinding {
 
 struct VulkanBinding {
   uint32_t Binding;
+  std::optional<uint32_t> CounterBinding;
 };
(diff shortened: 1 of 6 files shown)


luciechoi:
LGTM

PR #475, opened 2025-10-17
Enable GetDimensions test for Clang & DirectX

hekota:
These tests are now passing for DirectX. The SPIR-V backend does not yet support the intrinsic needed for this test, so the link now points to a new issue that is SPIR-V only.

### DIFF
--- a/test/Feature/ByteAddressBuffer/GetDimensions.test
+++ b/test/Feature/ByteAddressBuffer/GetDimensions.test
@@ -79,8 +79,8 @@ DescriptorSets:
 ...
 #--- end
 
-# Unimplemented https://github.com/llvm/wg-hlsl/issues/126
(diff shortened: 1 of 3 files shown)


spall (approved):
I guess I'd just suggest updating the pr description to specify they're not passing on clang vulkan

hekota:
My apologies, you have every right to be confused! I made a terrible typo in the PR description ("not passing" vs. "now passing"), and I did not explain the Vulkan part. In my mind this was "remove an XFAIL" change since the tests are now passing on Clang and DirectX. But then I realized Vulkan does not have support for the needed intrinsic yet, so I added the XFAIL back with a new issue link for Vulkan only. The description is now updated.

damyanp (approved):
LGTM, thanks!

Issue #182, opened 2025-05-12
Add support for reserved resources to the offload tester


damyanp:
We want to write tests for CheckAccessFullyMapped, but to do that we need reserved resources.  We may also want to consider being able to adjust tile mappings.

PR #476, opened 2025-10-17
Add partially mapped resources tests

bob80905:
This PR is the first step in bringing support to testing with partially mapped and unmapped resources.
It lays out the infrastructure necessary to initialize resources with only a certain number of tiles mapped.
The Load overload that takes the 2nd status argument is implemented in DXC, so we take advantage of that and test that when accessing data in an unmapped resource, we get 0'd data and the status integer is also 0.

The test is currently unsupported in Clang and Vulkan, but will be supported in the future.
Fixes https://github.com/llvm/offload-test-suite/issues/182

### DIFF
--- a/test/Feature/HLSLLib/PartiallyMappedResources.test
+++ b/test/Feature/HLSLLib/PartiallyMappedResources.test
@@ -22,7 +22,7 @@ void main() {
     Out[idx] = Result.x;
   else 
     Out[idx] = 9003;
-    
+
   idx += 1;
 
   Result = X.Load(4100, status);
@@ -92,7 +92,7 @@ Buffers:
     Format: Bool
     Stride: 4
     # Only the first data access should be accessing fully mapped memory
-    Data: [true, false, false, false]
+    Data: [1, 0, 0, 0]
 
 Results:
   - Result: Test


lib/API/DX/Device.cpp
```diff
@@ -753,7 +1134,7 @@ class DXDevice : public offloadtest::Device {
       // Zero any remaining bytes
       if (R.size() < CBVSize) {
         void *ExtraData = static_cast<char *>(ResDataPtr) + R.size();
-        memset(ExtraData, 0, CBVSize - R.size() - 1);
+        memset(ExtraData, 0, CBVSize - R.size());
```
bogner: Is this an unrelated bug fix? I'm nervous about a change like this without accompanying tests.

test/Feature/HLSLLib/PartiallyMappedResources.test
```diff
@@ -0,0 +1,132 @@
...
+      Kind: RWStructuredBuffer
+      DirectXBinding:
+        Register: 3
+        Space: 0
+      VulkanBinding:
+        Binding: 3
+#--- end
+
+# XFAIL: Clang
+# XFAIL: Vulkan
```
bogner: These need comments like `Unimplemented <link to issue>` or `Bug <link to issue>` so that we can track why they're XFAIL'd and loop back appropriately.

bogner (commented):
This is great. We may want to revisit and see if we can do some code deduplication in a nice way at some point, but I think this makes sense as is for now.

A few of minor comments.

lib/API/DX/Device.cpp
```diff
@@ -244,9 +244,12 @@ class DXDevice : public offloadtest::Device {
     ComPtr<ID3D12Resource> Upload;
     ComPtr<ID3D12Resource> Buffer;
     ComPtr<ID3D12Resource> Readback;
+    // In unmapped cases, the Heap lifetime needs to be preserved
```
damyanp: I don't understand what this comment means.

test/Feature/HLSLLib/PartiallyMappedResources.test
```diff
@@ -0,0 +1,132 @@
...
+DescriptorSets:
+  - Resources:
+    - Name: X
+      Kind: StructuredBuffer
+      DirectXBinding:
+        Register: 0
+        Space: 0
+      VulkanBinding:
+        Binding: 0
+      TilesMapped: 1
```
damyanp: Are we sure `TilesMapped` really belongs on the DescriptorSets part here?  I would expect it to be a part of the Buffers.

My reasoning here is that it becomes ambiguous what to do if multiple descriptors refer to the same resource but have different TilesMapped settings on them.

lib/API/DX/Device.cpp
```diff
@@ -571,6 +684,12 @@ class DXDevice : public offloadtest::Device {
     return Bundle;
   }
 
+  llvm::Expected<ResourceBundle> createSRV(Resource &R, InvocationState &IS) {
```
damyanp: Another possible approach would be to always use reserved resources, and then just have `tilesMapped == std::nullopt` or whatever result in mapping the entire resource.

test/Feature/HLSLLib/PartiallyMappedResources.test
damyanp: Oh, actually it looks like maybe the current structure of this creates the descriptors at the same time as the resources, so even though it doesn't look like it they are actually tied together....I keep being surprised that this is how this code works, but fixing that is way beyond the scope of this PR.

However, I do still think it makes sense to describe how the resource is mapped into memory in the Buffers section rather than the DescriptorSets one.

lib/API/DX/Device.cpp
damyanp: Although I'm not trying to push you too hard to a "always use reserved resources" design, this is an example where using the same type of resource in both cases would lead to a simpler implementation.

lib/API/DX/Device.cpp
```diff
@@ -653,19 +732,75 @@ class DXDevice : public offloadtest::Device {
               "Failed to create committed resource (readback buffer)."))
         return Err;
 
-      // Initialize the UAV data
+      // Tile mapping setup (optional if NumTiles > 0)
+      const UINT NumTiles = getNumTiles(R.TilesMapped, ResDesc.Width);
+
+      ComPtr<ID3D12Heap> Heap; // optional, only created if NumTiles > 0
+
+      if (NumTiles > 0) {
```
bogner: This whole block (up to the `updateTileMappings` call) looks identical to the SRV version. Can we pull out a helper function to call in both cases?

bob80905: Turns out I'll have to leave unmapped UAVs for another time, so I'll pull out a helper function in a future PR.

lib/API/DX/Device.cpp
```diff
@@ -521,51 +528,115 @@ class DXDevice : public offloadtest::Device {
...
+                            "Failed to create heap for tiled SRV resource."))
+          return Err;
+
+        // Fill tile coordinates and region sizes
+        for (UINT I = 0; I < NumTiles; ++I) {
+          StartCoords[I] = {I, 0, 0, 0};
+          RegionSizes[I].NumTiles = 1;
+          RegionSizes[I].UseBox = FALSE;
+          HeapRangeStartOffsets[I] = I;
+        }
```
damyanp: There was some talk about changing this to work with a bitmask of tiles to map, or maybe just a list of which tiles to map?  Is this planned for a future change, or are we not going to do it at all?

If we're always going to be mapping a contiguous range of tiles then this can be simplified.

lib/API/DX/Device.cpp
```diff
@@ -521,51 +528,115 @@ class DXDevice : public offloadtest::Device {
...
+      const D3D12_RANGE Range = {0, 0}; // no reads expected
+      if (SUCCEEDED(UploadBuffer->Map(0, &Range, &ResDataPtr))) {
+        memcpy(ResDataPtr, ResData.get(), R.size());
+        UploadBuffer->Unmap(0, nullptr);
+      } else {
+        return llvm::createStringError(std::errc::io_error,
+                                       "Failed to map SRV upload buffer.");
+      }
 
+      // Add GPU upload commands
```
damyanp: This comment adds no value since the name of the function is pretty clear about what it does.

lib/API/DX/Device.cpp
```diff
@@ -765,6 +961,12 @@ class DXDevice : public offloadtest::Device {
     return Bundle;
   }
 
+  llvm::Expected<ResourceBundle> createCBV(Resource &R, InvocationState &IS) {
+    if (R.TilesMapped)
+      return createReservedCBV(R, IS);
+    return createCommittedCBV(R, IS);
+  }
```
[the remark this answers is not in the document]
bob80905: I plan to eventually always use reserved CBV's, yes. Though at this point it makes sense to me to do it incrementally.

lib/API/DX/Device.cpp
bob80905: Ok, simplified. that bitmask idea should be a future change.

lib/API/DX/Device.cpp
```diff
@@ -1361,6 +1563,18 @@ class DXDevice : public offloadtest::Device {
     return llvm::Error::success();
   }
 
+  llvm::Error waitThenReturnErr(llvm::Error Err, InvocationState &IS) {
```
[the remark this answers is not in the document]
damyanp: Still sounds like something that should be separated out from the "do stuff with reserved resources" change, even if you can't write a test for it.

lib/API/DX/Device.cpp
damyanp: Are you still intending to undo this change, or did you find it is actually required?

lib/API/DX/Device.cpp
```diff
@@ -521,51 +528,103 @@ class DXDevice : public offloadtest::Device {
...
   }
 
+  UINT getNumTiles(std::optional<int> NumTiles, UINT64 Width) {
+    UINT Ret;
+    if (NumTiles.has_value())
+      Ret = static_cast<UINT>(*NumTiles);
+    else
+      // Map the entire buffer by computing how many 64KB tiles cover it
+      Ret = static_cast<UINT>(
+          (Width + D3D12_TILED_RESOURCE_TILE_SIZE_IN_BYTES - 1) /
```
spall: could this operation overflow? Why is width a uint64?

lib/API/DX/Device.cpp
```diff
@@ -753,7 +813,7 @@ class DXDevice : public offloadtest::Device {
       // Zero any remaining bytes
       if (R.size() < CBVSize) {
         void *ExtraData = static_cast<char *>(ResDataPtr) + R.size();
-        memset(ExtraData, 0, CBVSize - R.size() - 1);
+        memset(ExtraData, 0, CBVSize - R.size());
```
spall: should this line be unmodified? I see a long unresolvd comment thread about it and the conclusion is the change should be undone, but I still see a modification?

lib/API/DX/Device.cpp
```diff
@@ -244,9 +249,11 @@ class DXDevice : public offloadtest::Device {
     ComPtr<ID3D12Resource> Upload;
     ComPtr<ID3D12Resource> Buffer;
     ComPtr<ID3D12Resource> Readback;
+    ComPtr<ID3D12Heap> Heap;
     ResourceSet(ComPtr<ID3D12Resource> Upload, ComPtr<ID3D12Resource> Buffer,
-                ComPtr<ID3D12Resource> Readback)
-        : Upload(Upload), Buffer(Buffer), Readback(Readback) {}
+                ComPtr<ID3D12Resource> Readback,
+                ComPtr<ID3D12Heap> Heap = nullptr)
```
spall: why is this defaulted to null but the others are not?

lib/API/DX/Device.cpp
spall: could width or d3d12_tiled_resource_tile_size_in_bytes be uint32 max or half that amount; could this operation overflow after you change width to be a uint32?

bob80905: Yes, though for this API:
[UpdateTileMappings](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12commandqueue-updatetilemappings)
The pRangeTileCounts param is a UINT. Its value in my code comes directly from counting the number of tiles. So I don't think this should ever be > uint32_max.

damyanp:
Code LGTM, on assumption that all these tests failing aren't related to this change.

bogner:
> > Code LGTM, on assumption that all these tests failing aren't related to this change.
> 
> [llvm/llvm-project#166449](https://github.com/llvm/llvm-project/pull/166449) Actually, the above PR is a prerequisite for this PR, otherwise the new file will fail.

The test will only fail with clang without that change, it should pass with dxc - we generally mark such tests as `XFAIL: Clang` waiting on the unimplemented feature.

test/Feature/HLSLLib/PartiallyMappedResources.test
```diff
@@ -0,0 +1,161 @@
...
+  uint idx = 0;
+
+  uint status;
+  int4 Result = X.Load(0, status);
+  bool CAFMResult = CheckAccessFullyMapped(status);
+  CAFM[idx] = CAFMResult;
+  if (CAFMResult)
+    Out[idx] = Result.x;
+  else {
+    if (all(Result == int4(0,0,0,0)))
```
bogner: Is Result even defined if `CheckAccessFullyMapped` returns false? We've read from an unmapped tile in that case...

PR #478, opened 2025-10-21
Add tests for WaveActiveMin

KungFuDonkey:
Adds tests for WaveActiveMin, based on the WaveActiveMax tests in #429 

I generated the results using a python script, but I cannot figure out how to run the tests. Any guidance on that would help


spall (commented):
add newlines to the files missing them at the end. 
And it looks like WaveActiveMin.test is missing the XFAIL: Clang.

KungFuDonkey:
> add newlines to the files missing them at the end. And it looks like WaveActiveMin.test is missing the XFAIL: Clang.

Thanks for the suggestion! I updated the branch, so it should be ready now

test/WaveOps/WaveActiveMin.int16.test
```diff
@@ -0,0 +1,315 @@
...
+
+...
+#--- end
+
+# Bug https://github.com/llvm/llvm-project/issues/156775
+# XFAIL: Clang
+
+# Bug https://github.com/llvm/offload-test-suite/issues/393
+# XFAIL: Metal
+
```
Icohedron:
```suggestion

# REQUIRES: Int16
```

bogner:
Stale PR. Looks like tests for WaveActiveMin have made it in in the meantime and I don't see anything here that isn't covered.

PR #484, opened 2025-10-23
Add `select` non-boolean vector condition tests

kmpeng:
This PR is the first part of the fix for [#164018](https://github.com/llvm/llvm-project/issues/164018). Adds non-boolean vector condition tests to the `select` tests.


alsepkow (commented):
LGTUS: Finn and Alex pair PR review

Icohedron (approved):

PR #487, opened 2025-10-26
Added WaveActiveBitOrTests

KungFuDonkey:
Adds tests for WaveActiveBitOr,

These tests are only for uint32 and uint64, as uint16 and all other int types are not supported in dxil and spirv

### DIFF
--- a/test/WaveOps/WaveActiveBitOr.int64.test
+++ b/test/WaveOps/WaveActiveBitOr.int64.test
@@ -1,0 +1,197 @@
+#--- source.hlsl
+StructuredBuffer<uint64_t4> In  : register(t0);
+RWStructuredBuffer<uint64_t4> Out1 : register(u1);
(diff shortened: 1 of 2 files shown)


test/WaveOps/WaveActiveBitOr.int64.test
```diff
@@ -0,0 +1,197 @@
...
+#--- end
+
+
+
+# Bug https://github.com/llvm/llvm-project/issues/156775
+# XFAIL: Clang
+
+# Bug https://github.com/llvm/offload-test-suite/issues/393
+# XFAIL: Metal
+
```
Icohedron:
```suggestion

# REQUIRES: Int64
```

bogner:
Stale PR. Looks like tests for WaveActiveMin have made it in in the meantime and I don't see anything here that isn't covered.

PR #488, opened 2025-10-27
Add ScreenSpaceUV Graphics test

Alexander-Johnston:
Relies on https://github.com/llvm/offload-golden-images/pull/2

There's no particular target to this test (though it does use smoothstep), but I wrote it while debugging something else and figured it makes another good litmus test for things working alongside SimpleTriangle.


bogner (approved):
I think we managed to merge the goldenimage for this test many months ago but somehow forgot about the test itself. This LGTM, though it would be nice to clang-format the hlsl  so it matches llvm style (it should mostly work if you copy the files into independent .hlsl files and run it on that - there are known issues with single bracket attributes but none of those are present here so I suspect it will work just fine)

Alexander-Johnston:
Formatted. Should be mergable now

bogner:
The pre-checkin testing is showing failures. Can you take a look?

Alexander-Johnston:
Will do, cheers!

PR #496, opened 2025-11-05
Remove XFAILs from `select` tests

kmpeng:
These were fixed by https://github.com/llvm/llvm-project/pull/166279.


kmpeng:
Hm seems like a regression in the SPIRV backend that's unrelated to my changes - I tried removing the non-bool condition tests and also reverting the hlsl_alias_intrinsics.h code back to how it was before the bug was fixed, but it's still crashing. Will file a bug and XFAIL.

Issue #490, opened 2025-10-27
[WARP] Convergence tests failure


luciechoi:
[Convergence tests](https://github.com/llvm/offload-test-suite/pull/473#issuecomment-3444843014) are failing on Warp.

Created a separate issue for [Clang](https://github.com/llvm/llvm-project/issues/165288)

spall:
> Hello Sarah,
> 
> We don't know where the open an issue for wrap. Can you bring this to the attention of the right people? Thanks.

Could you or @luciechoi summarize the issue you are seeing with warp, so I can send it along to the right people? Thanks.

s-perron:
We just added at test that fails when compiled with DXC and run on WARP. I'm guessing the issue is not DXC (it pass on Intel when compiled with DXC). I'll leave that for you to decided how much you want to look at code generated by DXC before passing it off to WARP.

https://github.com/llvm/offload-test-suite/blob/e939efa266beb592a592c1a4504ce87eb0d3936b/test/Feature/MaximalReconvergence/subgroup_uniform_control_flow.test

PR #500, opened 2025-11-10
Remove `subgroup_uniform_control_flow` test.

luciechoi:
Update 11/03/2025
We would like to remove this test and add similar tests when porting vulkan conformance tests. Currently, there's not a good way to specify the subgroup size inside the tests.

~Fixes https://github.com/llvm/offload-test-suite/issues/490~

### DIFF
--- a/test/Feature/HLSLLib/dot2add.test
+++ b/test/Feature/HLSLLib/dot2add.test
@@ -92,9 +92,6 @@ DescriptorSets:
 # Bug https://github.com/llvm/llvm-project/issues/149561
 # XFAIL: Clang && Vulkan && !VK_KHR_shader_float_controls2
 
-# Bug https://github.com/microsoft/DirectXShaderCompiler/issues/7695
-# XFAIL: DXC && Vulkan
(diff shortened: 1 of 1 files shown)


luciechoi:
cc: @s-perron

test/Feature/MaximalReconvergence/subgroup_uniform_control_flow.test
```diff
@@ -1,22 +1,31 @@
 #--- source.hlsl
-RWStructuredBuffer<uint> Out : register(u0);
+RWStructuredBuffer<float> Out : register(u0);
 
 [numthreads(8,1,1)]
 void main(uint3 TID : SV_GroupThreadID) {
+    float maxActiveLaneCount = min(WaveGetLaneCount(), 8);
```
s-perron: I would leave this with `WaveActiveSum(1)` so we have a working test. No need for a comment.

This is not testing the wave intrinsics, so as long as we have something that is suppose to work we are fine.

PR #503, opened 2025-11-13
[Matrix] Add single subscript test

### LINKED ISSUE #502: [Matrix] Implement tests for single subscript that allows vector r and l values aswell as vector swizzle
Add testing for https://github.com/llvm/llvm-project/issues/166206

example
```hlsl
RWStructuredBuffer<int> In;
RWStructuredBuffer<int> Out;
cbuffer Constants {
    uint row;
    uint col;
};
[numthreads(1, 1, 1)]
void CSMain(uint GI : SV_GroupIndex) {
    int2x3 M = int2x3(0, 1, 2, 
                      3, 4, 5);
    Out[0] = M[row].x;
}
```

farzonl:
fixes #502 
https://godbolt.org/z/xaEE44cbE
Test 3 cases (loads and stores):
1. single subscript as an l-value swizzle
2. single subscript where the store/load takes a vector r-value swizzle
3. single subscript where we store/load the vector as is.


test/Basic/matrix_single_subscript_load.test
```diff
@@ -0,0 +1,71 @@
...
+[numthreads(1,1,1)]
+void main() {
+    int4x4 A = int4x4(In[0],  In[1],  In[2],
+                      In[3],  In[4],  In[5],
+                      In[6],  In[7],  In[8],
+                      In[9],  In[10], In[11],
+                      In[12], In[13], In[14],
+                      In[15]);
+    
+    for (int i = 0; i < 4; i++) {
```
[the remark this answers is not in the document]
farzonl: Even in llvm-project we don't run clang-format on tests, but the capital I is a c++ coding standard for c/c++ in llvm-project. I wasn't aware we defined a coding standard for HLSL but I suppose we could adopt the c/c++ one?

alsepkow (commented):
LGTM. Minor nit on the first for loop in both tests. Take it or leave it.

test/Basic/matrix_single_subscript_load.test
```diff
@@ -0,0 +1,71 @@
...
+RWBuffer<int> Out : register(u1);
+
+[numthreads(1,1,1)]
+void main() {
+    int4x4 A = int4x4(In[0],  In[1],  In[2],
+                      In[3],  In[4],  In[5],
+                      In[6],  In[7],  In[8],
+                      In[9],  In[10], In[11],
+                      In[12], In[13], In[14],
+                      In[15]);
```
Icohedron: Formatting nit:
```suggestion
    int4x4 A = int4x4(In[0],  In[1],  In[2],  In[3],
					  In[4],  In[5],  In[6],  In[7],
					  In[8],  In[9],  In[10], In[11],
					  In[12], In[13], In[14], In[15]);

```

test/Basic/matrix_single_subscript_load.test
```diff
@@ -0,0 +1,71 @@
...
+    
+    for (int i = 0; i < 4; i++) {
+        int4 B;
+        if (i % 2 == 0)
+            B = A[i];
```
tex3d: Just a comment, not a request for change:

These subscript accesses look dynamic, but these loops are likely to be unrolled, either before IR (in DXC, Clang, etc.) or after (in a driver compiler). That unrolling would make all accesses static, even eliminating the intermediate local variables to copy values directly from static indices of In to static indices of Out. That's a fine path to test, but this might not test other important dynamic indexing code paths, if desired.

I guess this gets at the fuzzy purpose issue with the offload-test-suite. This seems fine to test this scenario end-to-end, whatever path it may take through optimizations, and maybe that's all that's desired for this test.

test/Basic/matrix_single_subscript_load.test
```diff
@@ -0,0 +1,71 @@
...
+        else if (i % 3 == 0)
+            B = A[i].abgr;
```
tex3d: Note: `abgr` is also a symmetric swizzle. `bagr` is another asymmetric swizzle you could use.

test/Basic/matrix_single_subscript_store.test
```diff
@@ -0,0 +1,71 @@
...
+    int4x4 A;
+    for(int i = 0; i < 4; i++) {
+        if(i % 2 == 0)
+            A[i].rbag = vec;
+        // A[3] is exactly In Buffer
+        else if(i % 3 == 0)
+            A[i] = vec;
+        // A[1] is reverse In Buffer
+        else
+            A[i] = vec.abgr;
```
tex3d: Note: another symmetric swizzle.

test/Basic/matrix_single_subscript_load.test
```diff
@@ -0,0 +1,71 @@
...
+        else
+            B.grab = A[i];
```
tex3d: Assignment swizzling is always harder to follow, but this one happens to be identical in mapping to the same swizzle if used on source instead. That seems a little too convenient, so maybe the target swizzle should be different to avoid that symmetry and make sure it's not treated the same somehow.

`grab` is symmetric (equivalent to `grab` on source):
`B.grab` = `{5,6,7,8}` -> `{6,5,8,7}`
`B` = `{5,6,7,8}.grab` -> `{6,5,8,7}`

`agrb` is not symmetric (equivalent to `bgar` on source):
`B.agrb` = `{5,6,7,8}` -> `{7,6,8,5}`
`B` = `{5,6,7,8}.agrb` -> `{8,6,5,7}`
`B` = `{5,6,7,8}.bgar` -> `{7,6,8,5}`

Perhaps a comment with the equivalent in source swizzle form would help too:
```suggestion
        else // i == 1
            B.agrb = A[i]; // equivalent: B = A[1].bgar
```

Corresonding update to expected values would be:
```diff
-    Data: [ 1, 2, 3, 4, 6, 5, 8, 7, 9, 10, 11, 12, 16, 15, 14, 13 ]
+    Data: [ 1, 2, 3, 4, 7, 6, 8, 5, 9, 10, 11, 12, 16, 15, 14, 13 ]
```

test/Basic/matrix_single_subscript_load.test
tex3d: Note: `abgr` is also a symmetric swizzle. `bagr` is another asymmetric swizzle you could use.

tex3d (approved):
Minor comments/suggestions, nothing necessarily requiring changes regarding:
- dynamic/static subscript indexing and potential optimizations
- symmetrical swizzles (equivalent when used in source or dest)

PR #504, opened 2025-11-13
Add `ByteAddressBuffer` templated load/store tests

### LINKED ISSUE #266: Add tests for ByteAddressBuffer's templated load/store member functions
While we test the basic functionality of ByteAddressBuffer in https://github.com/llvm/offload-test-suite/blob/main/test/Feature/RawBuffers/ByteAddressBuffers.test, we need to also test templated loads and stores. These member functions are described here: https://github.com/microsoft/DirectXShaderCompiler/wiki/ByteAddressBuffer-Load-Store-Additions

kmpeng:
Closes #266.

Adds tests for ByteAddressBuffer's templated load/store member functions. These test one 32-bit type (uint), bool, one 16-bit type (int16_t), one 64-bit type (int64_t), a struct with an array in it, a small struct, and a big struct. I didn't test every single type of scalar since we're mostly trying to test the various different shapes that could be used.

### DIFF
--- a/include/Support/Pipeline.h
+++ b/include/Support/Pipeline.h
@@ -199,7 +199,8 @@ struct Resource {
   }
 
   uint32_t getElementSize() const {
-    // ByteAddressBuffer uses DXGI_FORMAT_R32_TYPELESS which has 4-byte elements
+    // ByteAddressBuffers are treated as 4-byte elements to match their memory
+    // format.
     return isByteAddressBuffer() ? 4 : BufferPtr->getElementSize();
   }
 


test/Feature/ByteAddressBuffer/ByteAddressBuffers.test
```diff
@@ -1,41 +1,98 @@
...
+  Mapped[7] = CheckAccessFullyMapped(status);
+  bool4 V7 = In1.Load<bool4>(64);
+  Out.Store<bool4>(112, U7 + V7);
+
+  // array
+  int U8[4] = In0.Load<int[4]>(0, status);
+  Mapped[8] = CheckAccessFullyMapped(status);
+  int V8[4] = In1.Load<int[4]>(0);
+  int TempArray[4];
+  for(int i = 0; i < 4; i++) {
```
alsepkow: nit: Should we follow coding standards in test code and make this a capital 'I' ?

lib/API/DX/Device.cpp
```diff
@@ -161,7 +169,7 @@ static D3D12_RESOURCE_DESC getResourceDescription(const Resource &R) {
 }
 
 static D3D12_SHADER_RESOURCE_VIEW_DESC getSRVDescription(const Resource &R) {
-  const uint32_t EltSize = R.getElementSize();
+  const uint32_t EltSize = R.isByteAddressBuffer() ? 4 : R.getElementSize();
```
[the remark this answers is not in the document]
alsepkow: Ah, it looks like this is assuming that a byte address buffer will be using a stride of 4. I'm not sure if that's a good assumption to hardcode?

test/Feature/ByteAddressBuffer/ByteAddressBuffers-16bit.test
```diff
@@ -0,0 +1,118 @@
...
+        Space: 0
+    - Name: Mapped
+      Kind: RWBuffer
+      DirectXBinding:
+        Register: 0
+        Space: 1
+...
+#--- end
+
+# UNSUPPORTED: Vulkan, Metal
```
alsepkow: Can we double check that this is still meant to be unsupported?

test/Feature/ByteAddressBuffer/ByteAddressBuffers.test
```diff
@@ -1,41 +1,98 @@
...
-  uint3 v3 = In2.Load3(32);
-  Out.Store3(32, u3 + v3);
-
-  uint4 u4 = In1.Load4(48, status);
-  MappedBuf[3] = CheckAccessFullyMapped(status);
-  uint4 v4 = In2.Load4(48);
-  Out.Store4(48, u4 + v4);
+  // uint
+  uint U0 = In0.Load(0, status);
+  Mapped[0] = CheckAccessFullyMapped(status);
```
alsepkow: Do you think it's worthwhile to have a test case where the data **isn't** fully mapped?

alsepkow (commented):
Finn/Alex pair review: Left a few minor comments.

test/Feature/ByteAddressBuffer/ByteAddressBuffers.test
kmpeng: I don't think it would be worth the effort—ByteAddressBuffers are currently created as committed resources, which means the tiles are fully mapped on creation. I would have to make changes to Device.cpp to create them as tiled resources to test this.

test/Feature/ByteAddressBuffer/ByteAddressBuffers-16bit.test
kmpeng: Seems like both are supported now! But only the Load without the status bit

kmpeng:
I have a question for people reviewing this: ByteAddressBuffer load/stores work with arrays in DXC (and [explicitly tests them](https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl)), but we agreed to move towards disallowing array returns in 202x. Should I keep the array test here? And should I implement array support for these methods in Clang?

bogner:
> I have a question for people reviewing this: ByteAddressBuffer load/stores work with arrays in DXC (and [explicitly tests them](https://github.com/microsoft/DirectXShaderCompiler/blob/main/tools/clang/test/CodeGenDXIL/hlsl/intrinsics/buffer-agg-load-stores.hlsl)), but we agreed to move towards disallowing array returns in 202x. Should I keep the array test here? And should I implement array support for these methods in Clang?

We discussed this elsewhere but for posterity I'll note it here: no, we shouldn't test the array behaviour or implement the array behaviour in clang if we're disallowing array returns in 202x. Tests that we reject array versions should be written, but those belong in the clang/dxc repos and don't need to be covered in the offload test suite here.

test/Feature/ByteAddressBuffer/ByteAddressBuffers.test
```diff
@@ -1,41 +1,83 @@
...
+  Out.Store<bool4>(112, U7 + V7);
+
+  // array
+  int U8[4] = In0.Load<int[4]>(0);
+  int V8[4] = In1.Load<int[4]>(0);
+  int TempArray[4];
+  for(int I = 0; I < 4; I++) {
+    TempArray[I] = U8[I] + V8[I];
+  }
+  Out.Store<int[4]>(128, TempArray);
```
bogner: As mentioned elsewhere, since we're planning on disallowing array returns we probably don't need this one. However, it would probably be good to have a struct with an array in it instead, so that we can at least get a relatively large read covered.

PR #508, opened 2025-11-19
Add fwidth graphics test

Alexander-Johnston:
Adds a Graphics test for the fwidth intrinsic. The test uses fwidth to provide a blur around the edge of a circle projected on a screensized quad.

I can't see a related issue open for this, but I'm providing it in-line with https://github.com/llvm/llvm-project/pull/161378

Relies on https://github.com/llvm/offload-golden-images/pull/4


farzonl:
These tests all appear to be passing. will merge.

PR #509, opened 2025-11-19
Updates for Clang 21.1

hekota:
This change addresses new warnings introduced in Clang 21.1.
- Updates a function to be static instead of being in anonymous namespace.
- Disables `-misc-const-correctness` warning in 3 cases where it is reported on a variable that cannot be made `const`.


alsepkow:
That const warning sounds like something nice to have. Do we intend to follow up? Are we submitting a quick fix right now to unblock builds?

PR #550, opened 2025-12-02
Fix more Clang 21.1 lint warnings

hekota:
Somehow these slipped through the cracks in my previous change.

### DIFF
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -1022,7 +1022,7 @@ public:
             HR::toError(IS.CmdList->Close(), "Failed to close command list."))
       return Err;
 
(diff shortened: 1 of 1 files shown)


lib/API/DX/Device.cpp
```diff
@@ -1022,7 +1022,8 @@ class DXDevice : public offloadtest::Device {
             HR::toError(IS.CmdList->Close(), "Failed to close command list."))
       return Err;
 
-    ID3D12CommandList *CmdLists[] = {IS.CmdList.Get()};
+    ID3D12CommandList *CmdLists[] = {
+        IS.CmdList.Get()}; // NOLINT(misc-const-correctness)
```
damyanp:
```suggestion
    ID3D12CommandList * const CmdLists[] = { IS.CmdList.Get()} }
```

Issue #591, opened 2025-12-04
[Intel][Vulkan][Clang] `Basic/Matrix/matrix_elementwise_vector_cast.test` fails on Intel under Vulkan with Clang


Icohedron:
The test `Basic/Matrix/matrix_elementwise_vector_cast.test` passes on all runners/workflows except for Intel when running under Vulkan with Clang.

This is perhaps the same as https://github.com/llvm/offload-test-suite/issues/599

```
╭────┬──────────────────────┬─────────────┬────────────────────────────────┬────────┬──────────────────────────────────────────────────╮
│  # │      timestamp       │   run-id    │            workflow            │ status │                       test                       │
├────┼──────────────────────┼─────────────┼────────────────────────────────┼────────┼──────────────────────────────────────────────────┤
│  0 │ 2025-12-04T12:08:25Z │ 19928434658 │ Windows D3D12 AMD Clang        │ PASS   │ Basic/Matrix/matrix_elementwise_vector_cast.test │
│  1 │ 2025-12-04T16:04:50Z │ 19935463445 │ Windows D3D12 Intel Clang      │ PASS   │ Basic/Matrix/matrix_elementwise_vector_cast.test │
│  2 │ 2025-12-04T16:05:54Z │ 19935497232 │ Windows D3D12 NVIDIA Clang     │ PASS   │ Basic/Matrix/matrix_elementwise_vector_cast.test │
│  3 │ 2025-12-04T16:01:09Z │ 19935345622 │ Windows D3D12 QC Clang         │ PASS   │ Basic/Matrix/matrix_elementwise_vector_cast.test │
│  4 │ 2025-12-04T12:04:14Z │ 19928323680 │ Windows D3D12 Warp Clang       │ PASS   │ Basic/Matrix/matrix_elementwise_vector_cast.test │
│  5 │ 2025-12-04T16:06:26Z │ 19935513385 │ Windows ARM64 D3D12 Warp Clang │ PASS   │ Basic/Matrix/matrix_elementwise_vector_cast.test │
│  6 │ 2025-12-04T12:01:06Z │ 19928239631 │ Windows D3D12 AMD DXC          │ PASS   │ Basic/Matrix/matrix_elementwise_vector_cast.test │
│  7 │ 2025-12-04T17:03:14Z │ 19937217662 │ Windows D3D12 Intel DXC        │ PASS   │ Basic/Matrix/matrix_elementwise_vector_cast.test │
[...]

PR #593, opened 2025-12-04
Update XFAIL conditions of `matrix_elementwise_vector_cast` and `matrix_trunc_cast` to match runner results

Icohedron:
The two tests are XPASSing on all machines except Intel under Vulkan with Clang. See:
- https://github.com/llvm/offload-test-suite/issues/591
- https://github.com/llvm/offload-test-suite/issues/592

But I believe the existing linked issue for the XFAIL is still the most relevant one so I did not change the linked bug. 
https://github.com/llvm/offload-test-suite/issues/591 and https://github.com/llvm/offload-test-suite/issues/592 can probably be closed as duplicates of https://github.com/llvm/offload-test-suite/issues/599.

### DIFF
--- a/test/Basic/Matrix/matrix_elementwise_vector_cast.test
+++ b/test/Basic/Matrix/matrix_elementwise_vector_cast.test
@@ -73,7 +73,7 @@ DescriptorSets:
 ...
 #--- end
 
-# Bug https://github.com/llvm/llvm-project/issues/170538
+# Bug https://github.com/llvm/offload-test-suite/issues/599
 # XFAIL: Intel && Vulkan && Clang
 
 # RUN: split-file %s %t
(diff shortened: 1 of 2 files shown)


farzonl:
hmm if this is a driver bug maybe it shouldn't be in https://github.com/llvm/llvm-project and instead we should close https://github.com/llvm/llvm-project/issues/170538.

Icohedron:
> hmm if this is a driver bug maybe it shouldn't be in https://github.com/llvm/llvm-project and instead we should close [llvm/llvm-project#170538](https://github.com/llvm/llvm-project/issues/170538).

We can also transfer the issue from llvm-project to the offload-test-suite repo. I believe the issue link will also redirect to the new one if we do that.

PR #607, opened 2025-12-05
Remove XFAIL for crash in SPIRVPreLegalizer for select

bogner:
Now that llvm/llvm-project#166642 is resolved, we no longer crash trying to generate SPIR-V. However, there does seem to be a crash in the intel driver when trying to run these shaders (llvm/offload-test-suite#615)


Icohedron (approved):
LGTM, assuming the test results from the runners agree. (Which so far seems to be the case)

PR #614, opened 2025-12-08
Add tests for resources in structs

hekota:
This change adds tests for resource inside used-defined structs and classes. It includes 4 variants:
- a single resource inside a struct
- resource array inside a struct
- array of structs that contain a resource
- a class with a mix of UAV and SRV resources that also inherits a resource member from its base class

Closes https://github.com/llvm/wg-hlsl/issues/368

### DIFF
--- a/include/Support/Pipeline.h
+++ b/include/Support/Pipeline.h
@@ -288,6 +288,34 @@ struct IOBindings {
   }
 };
 
+// Describes a contiguous group of bytes in a push constant block.
+struct PushConstantValue {
+  // Format used to describe those bytes in the YAML.
+  DataFormat Format;
+  // The bytes of this group.
+  llvm::SmallVector<char, 4> Data;
+  // The offset of this group from the start of the push constant buffer.
+  uint32_t OffsetInBytes;
+};
+
+// Describes the content of the push constant buffer.
+struct PushConstantBlock {
+  // The stages this is push constant is active for.
+  Stages Stage;
+
+  // The values/group of values composing this buffer.
+  llvm::SmallVector<PushConstantValue> Values;
+
+  // True if there are no push constants.
+  bool empty() const { return size() == 0; }
+
+  // Layout the push constant content in output.
+  void getContent(llvm::SmallVectorImpl<uint8_t> &output) const;
+
+  // Returns the size in bytes of the whole push constant once laid out.
+  uint32_t size() const;
+};
+
 struct SpecializationConstant {
   uint32_t ConstantID;
   DataFormat Type;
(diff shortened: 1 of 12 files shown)


test/Feature/ResourcesInStructs/res-in-struct-simple-one.test
```diff
@@ -0,0 +1,71 @@
...
+// This test verifies handling of a simple resource in a struct.
+
+struct A {
+  RWBuffer<int> Buf;
+};
+
+[[vk::binding(5)]]
+A a1 : register(u5);
+
+A a2;
```
farzonl: This is register(u0)?

test/Feature/ResourcesInStructs/res-in-struct-simple-array.test
```diff
@@ -0,0 +1,81 @@
...
+};
+
+[[vk::binding(3)]]
+B b1 : register(u3);
+B b2;
+  
+[numthreads(4,1,1)]
+void main(uint3 ID : SV_GroupThreadID) {
+  float x = b1.Bufs[0][ID.x];
+  float y = b1.Bufs[1][ID.x];
```
farzonl: Your buffers are all ints whats the point of doing these floating point implicit casts when you are going to store it back as an int?

farzonl (approved):

test/Feature/ResourcesInStructs/res-in-struct-simple-one.test
hekota: Yes, I wanted to test implicit binding as well. I can add a comment.

test/Feature/ResourcesInStructs/res-in-struct-mix.test
```diff
@@ -0,0 +1,135 @@
...
+    Format: Int32
+    Data:
+     [ 10, 20, 30, 40 ]
+
+  - Name: Buf_g_e_UavBufs1
+    Format: Int32
+    Data:
+      [ 100, 200, 300, 400 ]
+
+  - Name: Buf_g_UavBufs2
```
kmpeng:
```suggestion
  - Name: Buf_g_UavBuf2
```
nit: make the Uav name the same as the variable

test/Feature/ResourcesInStructs/res-in-struct-mix.test
```diff
@@ -0,0 +1,135 @@
...
+
+class F {
+  StructuredBuffer<int> SrvBuf1;
+};
+
+class G : F {
+  E e;
+  StructuredBuffer<int> SrvBuf2;
+  RWBuffer<int> UavBuf2;
+  uint n;
```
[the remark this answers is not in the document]
hekota: This is to make sure the struct a field that is not a resource. Mixing of resources and other types is actually not supported by SPIR-V.

PR #618, opened 2025-12-15
[Github] Test workflows when modifying them

boomanaiden154:
The postcommit workflows set to run as cron jobs are difficult to test without landing changes in main. Add a workflow trigger to run them on pull requests when the workflow definition gets modified.

This will also soon be listed in LLVM's CI best practices document soon. https://github.com/llvm/llvm-project/pull/172235

Landing this primarily because it makes it easier to test some changes I want to make soon.


boomanaiden154:
@Icohedron and @llvm-beanz to review since I do not have permissions to request reviews here.

PR #631, opened 2025-12-18
[Matrix] Add basic single subscript tests

### LINKED ISSUE #630: [Matrix] Add Basic single subscript tests that don't swizzle
The implementation of swizzle on single subscript is significantly different enough from  get\set single index  behavior  and even constant row index behavior that to test for completion of this behavior we need a more basic set of tests.

The subset of tests needed are
```hlsl
export float4 getMatrix(float4x4 M, int index) {
    return M[index];
}

export void setMatrix(out float4x4 M, int index, float4 V) {
    M[index] = V;
}

export void setMatrix3(out float4x4 M, float4 V) {
    M[3].abgr = V;
}

export void setMatrix4(out float4x4 M, float4 V) {
    M[3].rgba = V;
}
```

The following tests will be excluded
```hlsl
export void setMatrix2(out float4x4 M, int index, float4 V) {
    M[index].abgr = V;
}

export float3 getMatrix3(float4x4 M, int index) {
   return M[index].rgb;
}
```

farzonl:
fixes #630

This change just does single subscript gets\sets.
It reserves swizzle use to only when a matrix single index is constant in the subscript.

### DIFF
--- a/test/Basic/Matrix/matrix_const_single_subscript.i2x3.test
+++ b/test/Basic/Matrix/matrix_const_single_subscript.i2x3.test
@@ -14,8 +14,8 @@ void main() {
     Mat[0] = B.zyx;
 
     int2x3 Mat2;
-    Mat2[Index[1]] = Mat[1].yzx;
-    Mat2[1].zxy = Mat[Index[1]];
+    Mat2[Index[0]] = Mat[1].yzx;
+    Mat2[0].zxy = Mat[Index[1]];
 
     const uint COLS = 3;
     for(uint i = 0; i < 6; i++) {
@@ -50,7 +50,7 @@ Buffers:
     FillSize: 24
   - Name: ExpectedOut2
     Format: Int32
-    Data: [ 3, 2, 1, 5, 4, 6 ]
+    Data: [ 5, 4, 6, 3, 2, 1 ]
 Results:
   - Result: Out
     Rule: BufferExact


test/Basic/Matrix/matrix_single_subscript_basic.test
```diff
@@ -0,0 +1,113 @@
...
+    int4 D = int4(In[12], In[13], In[14], In[15]);
+
+    int4x4 Mat;
+    Mat[Index[0]] = A;
+    Mat[Index[1]] = B;
+    Mat[Index[2]] = C;
+    Mat[Index[3]] = D;
+
+    int4x4 Mat2;
+    Mat2[Index[3]] = getMatrix(Mat, Index[0]);
```
spall: why the indirection here? why not inline the body of GetMatrix?

farzonl: sure

test/Basic/Matrix/matrix_const_single_subscript.i2x3.test
```diff
@@ -0,0 +1,104 @@
...
+  - Stage: Compute
+    Entry: main
+    DispatchSize: [1, 1, 1]
+Buffers:
+  - Name: In
+    Format: Int32
+    Data: [ 1, 2, 3, 4, 5, 6]
+  - Name: Index
+    Format: Int32
+    Data: [ 1, 0]
```
spall: index at zero isn't used? You could make this of size 1?

spall (approved):
small comment but lgtm

PR #637, opened 2025-12-23
[Github] Move elevated permissions to the job level

boomanaiden154:
This makes the offload test suite workflows more conformant with https://llvm.org/docs/CIBestPractices.html#top-level-read-permissions. This change should be NFC. This mainly protects against new jobs being added to the workflows that do not need elevated permissions but getting them anyways.


boomanaiden154:
CC @Keenuts @llvm-beanz @Icohedron for review

Icohedron (approved):

PR #646, opened 2026-01-12
[lit][offloader][ninja] Add ability to toggle gpu by name

farzonl:
This change allows you to set an enviornment variable to pick which gpu the offloader will use. For example say I have an Intel GPU I can do:
```
set "OFFLOADTEST_GPU_NAME=Intel(R) Arc(TM) A770 Graphics" && ninja -C ..\llvm-build\ check-hlsl-clang-d3d12
```
Or if I just want to specify the vendor I can do
```
set "OFFLOADTEST_GPU_NAME=Intel" && ninja -C ..\llvm-build\ check-hlsl-clang-d3d12
```

This was made via `llvm::regex`  so you can use any casing. Say on linux we would do 

```bash
vulkaninfo --summary | grep deviceName
        deviceName         = NVIDIA GeForce RTX 4070 Ti (NVK AD104)
        deviceName         = llvmpipe (LLVM 20.1.2, 256 bits)
OFFLOADTEST_GPU_NAME=nvidia ninja -C ../llvm_builds/llvm_rel_with_debug_offload/ check-hlsl-clang-vk
OFFLOADTEST_GPU_NAME=llvmpipe ninja -C ../llvm_builds/llvm_rel_with_debug_offload/ check-hlsl-clang-vk
```

I am also adding a way to use the adapter index because in testing I found typing out even just the substring can be more annoying than a single char for the gpu index. The index is just the order we see them in the device array.

I am also adding the ability to toggle between vulkan drivers on the mac. To do this we just needed to encode the driver name into the Description. Since `KosmicKrisp` is apple silicon only I limit its usage to `if defined(__APPLE__) && defined(__aarch64__)`.

```bash
OFFLOADTEST_GPU_NAME=KosmicKrisp ninja -C ../builds/llvm_relwithdebinfo_offload check-hlsl-clang-vk
```
This change was useful for me to pick between multiple discrete gpus on a single rig, egpus, and integrated gpus. It works now for both DirectX and Vulkan devices.

### DIFF
--- a/test/lit.cfg.py
+++ b/test/lit.cfg.py
@@ -144,7 +144,7 @@ if config.offloadtest_enable_debug:
 if config.offloadtest_enable_validation:
     offloader_args.append("-validation-layer")
 if ShouldSearchByGPuName:
-    offloader_args.extend([f'-adapter-substring="{GPUName}"'])
+    offloader_args.extend([f'-adapter-regex="{GPUName}"'])
 tools.append(
     ToolSubst("%offloader", command=FindTool("offloader"), extra_args=offloader_args)
 )
--- a/tools/offloader/offloader.cpp
+++ b/tools/offloader/offloader.cpp
@@ -63,10 +63,11 @@ static cl::opt<bool> Validation("validation-layer",
 
 static cl::opt<bool> UseWarp("warp", cl::desc("Use warp"));
 
-static cl::opt<std::string>
-    AdapterSubstring("adapter-substring",
-                     cl::desc("Which GPU vendor to use by name"),
-                     cl::value_desc("<name>"), cl::init(""));
+static cl::opt<std::string> AdapterRegex(
+    "adapter-regex",
+    cl::desc(
+        "Case-insensitive regular expression to match GPU adapter description"),
+    cl::value_desc("<regex>"), cl::init(""));
 
 static std::unique_ptr<MemoryBuffer> readFile(const std::string &Path) {
   const ExitOnError ExitOnErr("gpu-exec: error: ");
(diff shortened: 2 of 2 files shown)


test/lit.cfg.py
```diff
@@ -197,14 +203,16 @@ def setDeviceFeatures(config, device, compiler):
...
 target_device = None
-
 # Find the right device to configure against
 for device in devices["Devices"]:
     is_warp = "Microsoft Basic Render Driver" in device["Description"]
+    is_gpu_name_match = GPUName in device["Description"]
     if device["API"] == "DirectX" and config.offloadtest_enable_d3d12:
-        if is_warp and config.offloadtest_test_warp:
+        if ShouldSearchByGPuName and is_gpu_name_match:
+            target_device = device
```
farzonl: the reason we have to pick the right target device in lit and not just set the offloader args is because setDeviceFeatures checks vendor names to turn on\off vendor specifc tests and we need this to match what we are telling the offloader.

include/API/Device.h
```diff
@@ -33,6 +33,8 @@ struct Pipeline;
 struct DeviceConfig {
   bool EnableDebugLayer = false;
   bool EnableValidationLayer = false;
+  int AdapterIndex = -1;
+  std::string AdapterSubstring = "";
```
farzonl: Not sure what the right name is for this. It can techically be the whole Device Description name. It can also be a subset like the Vendor name i.e (Intel, Nvidia, AMD, etc).

hekota: In `DXGI_ADAPTER_DESC1 ` it is called `Description`, so maybe `AdapterDescription` would be a good fit?

farzonl: I wanted a name to indicate it could be a substring

llvm-beanz (commented):
I'm curious why you decide to do the filtering on the creation of the device instead of the selection. Filtering on selected device is API-agnostic, but this implementation needs to be done per API backend.

test/lit.cfg.py
```diff
@@ -41,6 +41,11 @@
 # Tweak the PATH to include the tools dir.
 llvm_config.with_environment("PATH", config.llvm_tools_dir, append_path=True)
 
+# Environment equivalents (useful for ninja):
+#   OFFLOADTEST_GPU_NAME
+GPUName = os.environ.get("OFFLOADTEST_GPU_NAME", "")
```
llvm-beanz: It would be nice if this could also be a CMake-provided value. That could also make it possible to have CMake generate configurations for more than one device from a single configuration tree.

include/API/Device.h
farzonl: The device itself stores the  string from querying the device driver. This is just the string we get from the offloader cli flags so we know what to search for.

PR #648, opened 2026-01-12
Add multiple tests for semantics

Keenuts:
This commits adds multiple tests around user & system semantics. This commit also enables the only graphic tests for Clang+Vulkan as it is now supported.

### DIFF
--- a/test/Feature/Semantics/SemanticTypes.test
+++ b/test/Feature/Semantics/SemanticTypes.test
@@ -71,7 +71,7 @@ Buffers:
     FillValue: 0
   - Name: ResultBuffer_Expected
     Format: Hex32
-# 0xffffffff == IEEE_754_f32(-1)
+# 0xffffffff == -1
 # 0x40600000 == IEEE_754_f32(3.5)
     Data: [ 0xffffffff, 0x2, 0x40600000 ]
 Bindings:


test/Feature/Semantics/ComputeSystemValues.test
```diff
@@ -0,0 +1,39 @@
+#--- system_values.hlsl
+RWStructuredBuffer<uint4> Out : register(u0);
+
+[numthreads(2, 1, 1)]
+void main(uint3 DTID : SV_DispatchThreadID, uint3 GID : SV_GroupID, uint3 GTID : SV_GroupThreadID, uint GI : SV_GroupIndex) {
+    Out[DTID.x] = uint4(DTID.x, GID.x, GTID.x, GI);
```
s-perron: This could be more thorough. You are only getting the `x` element. Could we have 3 or 4 output buffers. Then you can check the results for each buffer.

s-perron (commented):
Generally looks good. I think one test could be improved.

s-perron (commented):
I think the testing is good enough, but a test could be

test/Graphics/ShadowedSemantics.test
s-perron: What is a shadowed semantic? A comment might be nice.

s-perron (commented):
I think we should try to simplify the compute shader test. I also don't understand the non-determinism that you are seeing.

I made a few other comments that can be applied in other places too. For example, all of the tests should be in the semantics folder.

test/Graphics/ArraySemantics.test
```diff
@@ -0,0 +1,74 @@
...
+# Semantics are not implemented in the DXIL backend.
+# XFAIL: Clang && !Vulkan
+
+# RUN: split-file %s %t
+# RUN: %dxc_target -T vs_6_0 -Fo %t-vertex.o %t/vertex.hlsl
+# RUN: %dxc_target -T ps_6_0 -Fo %t-pixel.o %t/pixel.hlsl
+# RUN: %offloader %t/pipeline.yaml %t-vertex.o %t-pixel.o | FileCheck %s
+
+# CHECK: Data:
+# CHECK: [ 0.5, 1, 0, 1 ]
```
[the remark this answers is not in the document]
Keenuts: Because there is a single 4-element array in the pipeline output. Adding the `Name` and check-next instead.

test/Feature/Semantics/SemanticTypes.test
```diff
@@ -0,0 +1,105 @@
...
+# XFAIL: Metal
+
+# RUN: split-file %s %t
+# RUN: %dxc_target -T vs_6_0 -Fo %t-vertex.o %t/vertex.hlsl
+# RUN: %dxc_target -T ps_6_0 -Fo %t-pixel.o %t/pixel.hlsl
+# RUN: %offloader %t/pipeline.yaml %t-vertex.o %t-pixel.o | FileCheck %s
+
+# CHECK: Name: ResultBuffer
+# 1080033280 == 0x40600000 == IEEE_754_f32(3.5)
+# CHECK: [ -1, 2, 1080033280 ]
```
s-perron: Could you update these with the CHECK-NEXT as well?

test/Feature/Semantics/MatrixSemantics.test
```diff
@@ -0,0 +1,103 @@
...
+# See https://github.com/llvm/offload-test-suite/issues/744
+# XFAIL: Metal
+
+# RUN: split-file %s %t
+# RUN: %dxc_target -T vs_6_0 -Fo %t-vertex.o %t/vertex.hlsl
+# RUN: %dxc_target -T ps_6_0 -Fo %t-pixel.o %t/pixel.hlsl
+# RUN: %offloader %t/pipeline.yaml %t-vertex.o %t-pixel.o | FileCheck %s
+
+# CHECK: Name: ResultBuffer
+# CHECK: [ 1, 2, 3, 4, 5, 6, 7 ]
```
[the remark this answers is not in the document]
Keenuts: changed to actual buffer check

PR #651, opened 2026-01-14
Add support for RenderDoc API to offload test suite to capture for Vulkan on Windows

spall:
Add support to get a frame capture for Vulkan on Windows using the RenderDoc API.

### DIFF
--- a/lib/API/VK/Device.cpp
+++ b/lib/API/VK/Device.cpp
@@ -13,6 +13,11 @@
 #include "Support/Pipeline.h"
 #include "llvm/ADT/DenseSet.h"
 #include "llvm/Support/Error.h"
+#include "renderdoc_app.h"
+
+#ifdef _WIN32
+#include <windows.h>
+#endif
 
 #include <memory>
 #include <numeric>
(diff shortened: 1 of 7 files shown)


lib/API/VK/Device.cpp
```diff
@@ -1951,10 +1956,36 @@ class VKDevice : public offloadtest::Device {
...
+    RENDERDOC_API_1_4_1 *RDocAPI = NULL;
+    if (Capture) {
+      // Initialize RenderDoc on Windows
+#if defined(_WIN32)
+      if (HMODULE Mod = GetModuleHandleA("renderdoc.dll")) {
+        pRENDERDOC_GetAPI RENDERDOC_GetAPI =
+	  (pRENDERDOC_GetAPI)GetProcAddress(Mod, "RENDERDOC_GetAPI");
+	int Ret = RENDERDOC_GetAPI(eRENDERDOC_API_Version_1_4_1, (void **)&RDocAPI);
+	assert(Ret == 1);
+	llvm::outs() << "RenderDoc API initialized.\n";
```
damyanp: This doesn't really check the return code from GetAPI - so if the user has, say, an older version of renderdoc.dll then it'll happily say "RenderDoc API initialized" and then not take a capture.

Related to this, I don't think we want to handle runtime errors like this with an assert.

lib/API/VK/Device.cpp
```diff
@@ -1951,10 +1956,36 @@ class VKDevice : public offloadtest::Device {
...
+      } else {
+        DWORD error = GetLastError();
+	llvm::outs() << "Failed to get renderdoc dll: " << error << "\n";
```
damyanp: `llvm::mapLastWindowsError()` might be interesting to look at here.  Looks like it should convert to strings as well.

Kinda hard to find the right balance for polish for a feature like this.

lib/API/VK/Device.cpp
```diff
@@ -1951,10 +1956,36 @@ class VKDevice : public offloadtest::Device {
...
     if (auto Err = createDevice(State))
       return Err;
+
+    // Start capture
+    if (RDocAPI) {
+      RDocAPI->StartFrameCapture(NULL, NULL);
+      llvm::outs() << "RenderDoc frame capture started.\n";
+    } else if(Capture)
+      llvm::outs() << "RenderDoc frame capture NOT started because we failed to \
+                      initialize RenderDoc API. It is only supported on Windows.";
```
damyanp:
```suggestion
    } else if(Capture) {
      llvm::outs() << "RenderDoc frame capture NOT started because we failed to \
                      initialize RenderDoc API. It is only supported on Windows.";
    }
```

I don't know why the diffs on suggestions is so broken these days - anyway suggestion is to add braces around the else if.  

https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements:

> Inconsistent bracing within if/else if/else chains (if one block requires braces, all must)

lib/API/VK/Device.cpp
```diff
@@ -1951,10 +1956,36 @@ class VKDevice : public offloadtest::Device {
     return llvm::Error::success();
   }
 
-  llvm::Error executeProgram(Pipeline &P) override {
+  llvm::Error executeProgram(Pipeline &P, bool Capture = false) override {
+    RENDERDOC_API_1_4_1 *RDocAPI = NULL;
+    if (Capture) {
+      // Initialize RenderDoc on Windows
+#if defined(_WIN32)
+      if (HMODULE Mod = GetModuleHandleA("renderdoc.dll")) {
```
damyanp: This leaks the module.  I don't know if we care about that though.

PR #654, opened 2026-01-16
Add tests for concurrent writes to matrices

Icohedron:
Adds new offload tests as mentioned should be done in https://github.com/llvm/llvm-project/issues/174629

These tests are concurrently writing to a matrix in groupshared memory instead of directly in UAVs because the writing of matrices to UAVs is not currently implemented in Clang.

The XFAIL on Clang with bug https://github.com/llvm/llvm-project/issues/174629 can be removed after https://github.com/llvm/llvm-project/pull/176216 is merged.

### DIFF
--- a/test/WaveOps/GroupSharedMatrixEltComponentDataRace.test
+++ b/test/WaveOps/GroupSharedMatrixEltComponentDataRace.test
@@ -62,9 +62,6 @@ DescriptorSets:
 ...
 #--- end
 
-# Bug: https://github.com/llvm/llvm-project/issues/174629
-# XFAIL: Clang && !WARP
-
 # Bug https://github.com/llvm/llvm-project/issues/172577
 # XFAIL: Clang && Vulkan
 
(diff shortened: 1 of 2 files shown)


bob80905 (commented):
Couple questions:
1. Would it be worth checking for the behavior within a single wave? 
2. Would using the DispatchSize attribute under Shaders be useful for these tests?

Icohedron:
> Couple questions:
> 
> 1. Would it be worth checking for the behavior within a single wave?
> 2. Would using the DispatchSize attribute under Shaders be useful for these tests?

I don't think trying to observe the behavior within individual waves or between thread groups is relevant to the issue this PR is made to address/test.
We just want to observe that data races between threads can occur when writing to independent matrices elements. 
Whether or not the data races occur specifically within warps or between thread groups is not a question that needs to be answered (or at least not at the moment).

bob80905 (commented):
All concerns resolved, LGTM

PR #699, opened 2026-02-02
Remove nan, inf, and denorm from sin.32 and cos.32 tests

Icohedron:
The sin.32 and cos.32 test are [failing on the new Mac Mini](https://github.com/llvm/offload-test-suite/actions/runs/21607660102/job/62268507394#step:12:458) due to not properly handling NaN, Inf, and Denorm floats.
This PR replaces NaN, Inf, and Denorm values of these tests with more relevant floating-point values.

### DIFF
--- a/test/Feature/HLSLLib/cos.32.test
+++ b/test/Feature/HLSLLib/cos.32.test
@@ -25,8 +25,8 @@ Buffers:
   - Name: In
     Format: Float32
     Stride: 16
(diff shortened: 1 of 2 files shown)


spall (approved):

bob80905:
I thought we intended to use `-Gis` for these types of tests. Should we see if they pass with this flag before removing them?

bob80905 (approved):
LGTM

Issue #525, opened 2025-12-01
[NV][Clang] `WaveActiveSum.*` tests fail on Windows D3D12 NVIDIA Clang


Icohedron:
WaveActiveSum tests all fail on NVIDIA when the shader is compiled with Clang for DirectX, except for `WaveOps/WaveActiveSum.test`
```
╭────┬──────────────────────┬─────────────┬────────────────────────────┬────────┬──────────────────────────────────╮
│  # │      timestamp       │   run-id    │          workflow          │ status │               test               │
├────┼──────────────────────┼─────────────┼────────────────────────────┼────────┼──────────────────────────────────┤
│  0 │ 2025-11-28T18:06:07Z │ 19771055797 │ Windows D3D12 NVIDIA Clang │ PASS   │ WaveOps/WaveActiveSum.test       │
│  1 │ 2025-11-28T06:03:54Z │ 19755547346 │ Windows D3D12 NVIDIA DXC   │ PASS   │ WaveOps/WaveActiveSum.fp32.test  │
│  2 │ 2025-11-28T06:03:54Z │ 19755547346 │ Windows D3D12 NVIDIA DXC   │ PASS   │ WaveOps/WaveActiveSum.fp64.test  │
│  3 │ 2025-11-28T06:03:54Z │ 19755547346 │ Windows D3D12 NVIDIA DXC   │ PASS   │ WaveOps/WaveActiveSum.fp16.test  │
│  4 │ 2025-11-28T06:03:54Z │ 19755547346 │ Windows D3D12 NVIDIA DXC   │ PASS   │ WaveOps/WaveActiveSum.test       │
│  5 │ 2025-11-28T06:03:54Z │ 19755547346 │ Windows D3D12 NVIDIA DXC   │ PASS   │ WaveOps/WaveActiveSum.int32.test │
│  6 │ 2025-11-28T06:03:54Z │ 19755547346 │ Windows D3D12 NVIDIA DXC   │ PASS   │ WaveOps/WaveActiveSum.int16.test │
│  7 │ 2025-11-28T06:03:54Z │ 19755547346 │ Windows D3D12 NVIDIA DXC   │ PASS   │ WaveOps/WaveActiveSum.int64.test │
│  8 │ 2025-11-28T18:06:07Z │ 19771055797 │ Windows D3D12 NVIDIA Clang │ XFAIL  │ WaveOps/WaveActiveSum.fp16.test  │
[...]

PR #700, opened 2026-02-02
[WARP] Add O1 to temporarily fix failures due to SimplifyCFG

bob80905:
This PR serves to temporarily resolve the issues facing WaveActiveSum* tests described in this issue:
https://github.com/llvm/offload-test-suite/issues/525
With -O1, the CFG is less aggressively optimized, preventing a switch construct from being generated, and allowing WARP to safely ingest the compiled clang IR.
When the root of the problem is resolved, the plan is to remove -O1.


damyanp:
Is this WARP only?  If so, then (and sorry we've been going back and forth on this a bit), I think we should just treat this like any other IHV bug - XFAIL it for WARP and remove the XFAIL once it is fixed.  I'm not sure why we would treat WARP differently to anything else for this one case.

PR #710, opened 2026-02-03
Update `ByteAddressBuffer` Load/Store XFAILs

kmpeng:
Removes the clang XFAILs after https://github.com/llvm/llvm-project/pull/176058 was merged and replaces them with clang-vulkan XFAILs. Also updates the bool tests to use `or` instead of boolean arithmetic because of the current problem with boolean truncation when loading them from memory.

### DIFF
--- a/test/Feature/ByteAddressBuffer/ByteAddressBuffers-16bit.test
+++ b/test/Feature/ByteAddressBuffer/ByteAddressBuffers-16bit.test
@@ -88,8 +88,8 @@ DescriptorSets:
 ...
 #--- end
 
-# Unimplemented https://github.com/llvm/llvm-project/issues/108058
-# XFAIL: Clang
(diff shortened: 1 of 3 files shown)


Icohedron:
`Feature/ByteAddressBuffers.test` is still failing on DirectX.
https://github.com/llvm/offload-test-suite/actions/runs/21649722416/job/62410889950#step:12:433

I think it needs a new issue associated with it for DirectX.

kmpeng:
> `Feature/ByteAddressBuffers.test` is still failing on DirectX. https://github.com/llvm/offload-test-suite/actions/runs/21649722416/job/62410889950#step:12:433

That was failing because the boolean arithmetic in the bool tests were giving the wrong value. Changing them to `or` should have fixed it.

Icohedron (approved):

PR #719, opened 2026-02-06
Update matrix swizzle tests

farzonl:
This change removes the group thread index from the basic matrix element expression testing. it removes the `# UNSUPPORTED: Clang` tag since  https://github.com/llvm/llvm-project/pull/171225 merged. 

It xfails the original tests until we get fixes in for 

- Vulkan  https://github.com/llvm/llvm-project/issues/180258
- DirectX https://github.com/llvm/llvm-project/issues/180262

### DIFF
--- a/test/Basic/Matrix/matrix_groupthread_swizzle_one_based.test
+++ b/test/Basic/Matrix/matrix_groupthread_swizzle_one_based.test
@@ -111,6 +111,8 @@ DescriptorSets:
 ...
 #--- end
 
+# BUG Vulkan  https://github.com/llvm/llvm-project/issues/180258
+# Bug DirectX https://github.com/llvm/llvm-project/issues/180262
 # XFAIL: Clang
 
 # RUN: split-file %s %t
--- a/test/Basic/Matrix/matrix_groupthread_swizzle_zero_based.test
+++ b/test/Basic/Matrix/matrix_groupthread_swizzle_zero_based.test
@@ -111,6 +111,8 @@ DescriptorSets:
 ...
 #--- end
 
+# BUG Vulkan  https://github.com/llvm/llvm-project/issues/180258
+# Bug DirectX https://github.com/llvm/llvm-project/issues/180262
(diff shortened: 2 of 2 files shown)


test/Basic/Matrix/matrix_groupthread_swizzle_one_based.test
```diff
@@ -0,0 +1,120 @@
...
+      vec2 = B._13_23_33_43;
+      break;
+    case 3:
+      vec1 = A._41_42_43_44;
+      vec2 = B._14_24_34_44;
+      break;
+  }
+  for(int i = 0; i < 4; i++) {
+    RowOut[GI*4+i] = vec1[i];
+    ColOut[i*4+GI] = vec2[i];
```
tex3d: IIUC, from the referenced bugs, these new "groupthread" versions catch a clang compiler crash triggered with a particular code pattern.

That's fine I suppose, but these tests touched by this PR appear to be all the matrix swizzle tests, and for that, the design is very odd.

All tests have the same basic pattern, which seems easily reduceable to eliminate the matrices completely, leaving one scalar value being replicated to contiguous 4 locations in RowOut and 4 strided locations in ColOut, for each of 4 threads (or loops). The purpose of writing with this pattern for RowOut and ColOut escapes me. It seems to only add needless complexity outside the tested scenario.

These tests wouldn't catch a wide variety of very clearly incorrect implementations. Here are a few hypothetical bugs that this test wouldn't catch:
- row/column mixed up
- swizzle locations wrong, even random or overlapping (but stable between read/write with same full swizzle)
- all swizzle locations pointing to the same, single, scalar location, even one scalar for both matrices
- completely broken matrix swizzles outside this degenerate case of a single, in-order row or column
- bad swizzle implementation when accessing non-square matrix (like row/column mixup in bounds check)
- and probably more...

It seems like it would be good to have matrix swizzle tests that would be more sensitive to bad/buggy implementations.

PR #722, opened 2026-02-08
Add Venus docs for Vulkan testing on Linux VMs

farzonl:
Venus is a Vulkan driver that allows Vulkan without vendor drivers in the guest OS.
It is very useful for testing Vulkan workloads without devoting a GPU to your VM the way VFIO requires.

These docs show how we can test Vulkan API forwarding without building any special software.

### DIFF
--- a/docs/Venus.md
+++ b/docs/Venus.md
@@ -1,18 +1,20 @@
 # Enabling GPU API virtualization on UTM VMs with macOS hosts 
 
-> Note: These docs are for mac users wanting on device linux development or a frictionless way to try out Venus.
->
+> Note: These docs are for mac users wanting on device linux development or a
+> frictionless way to try out Venus.
 
 ## 1. What is UTM?
 
-**UTM** is a macOS application for running virtual machines using Apple’s **Hypervisor.framework** and **QEMU**. It supports both:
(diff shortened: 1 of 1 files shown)


docs/Venus.md
```diff
@@ -0,0 +1,154 @@
...
+
+#### Storage Requirements
+
+Allocate **at least 100 GB** of disk space for:
+
+- LLVM build trees
+- DirectXShaderCompiler builds
+- Mesa and Vulkan tooling
+- LLVM and DXC software requirments
+  - (complers, linkers, build systems, etc)
```
bob80905: nit: compilers

bob80905 (commented):
Alex/Josh review:
Question: How popular is venus? Where would it end up in terms of significance, tier 1, tier 2?

farzonl:
> Alex/Josh review: Question: How popular is venus? Where would it end up in terms of significance, tier 1, tier 2?

It is used heavily in Chromos, the Android emulator, and in cloud computing scenarios.  It would be in no tier we won't add any CI for this.  Its an API pass through layer so its behavior is always going to be dependent on the underlying gpu that it doesn't know anything about. 

This is just a suggestion on how to set it up with software that doesn't make you build the mesa driver yourself.

PR #727, opened 2026-02-10
Remove nan, inf and denorm

joaosaffran:
Some tests are calling underspecified/unspecified api, which have different behavior on SPIR-V and DXIL. This doesn't help us know if the intrinsics are implemented correctly.

### DIFF
--- a/test/Feature/HLSLLib/log10.16.test
+++ b/test/Feature/HLSLLib/log10.16.test
@@ -24,16 +24,20 @@ Shaders:
 Buffers:
   - Name: In
     Format: Float16
-    Stride: 8
+    Stride: 0
     Data: [0x8000, 0x0000, 0x3c00, 0xbc00, 0x4900, 0x2e66, 0x5640,]
+    #  [-0.0, 0.0, 1.0, -1.0, 10.0, 0.0960693359375, 100.0]
   - Name: Out
     Format: Float16
-    Stride: 8
+    Stride: 0
     FillSize: 24
   - Name: ExpectedOut # The result we expect
     Format: Float16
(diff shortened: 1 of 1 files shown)


farzonl:
All of these are Mac Specific XFAILs  for vulkan. We don't test these as part of the CI so I don't  think this is going to help us close the gap at all. I don't think it hurts to remove these, but also the xfails I want to get rid of are ones like this one:

https://github.com/llvm/offload-test-suite/blob/928511712c8a666d1307536ac2d17d23e129abaa/test/Feature/HLSLLib/asin.16.test#L64-L74

I want to get rid of these xfails because essentially it means we are only doing Nvidia and Qualcomm testing on vulkan.

The inf, denorm, NaN testing is causing us to many of these vendor specific xfails which is making it hard for us to determine the completeness of these intrinsics. Our current plan is to change these tests to define a floor of correct behavior that should be correct across all  vendors and api versions.

test/Feature/HLSLLib/frac.16.test
```diff
@@ -25,17 +25,15 @@ Buffers:
   - Name: In
     Format: Float16
     Stride: 8
-    Data: [ 0x7e00, 0xfc00, 0x03FF, 0x8000, 0x0000, 0x7c00, 0xbc00, 0x4170, 0x63d1, 0xC764, 0x7e00, 0x7e00,]
-    #  NaN, -Inf, 0.00006097555, -0, 0, Inf, -1, 2.719, 1000.5, -7.390625,
```
farzonl: since the `half` is in hex we do need a comment to represent the numbers. Same below

Issue #541, opened 2025-12-02
[AMD][DirectX][DXC] Test `WaveOps/WaveActiveMax.fp64.test` fails on Windows D3D12 AMD DXC


Icohedron:
Test `WaveOps/WaveActiveMax.fp64.test` fails on Windows D3D12 AMD DXC

```
╭────┬──────────────────────┬─────────────┬──────────────────────────────┬────────┬─────────────────────────────────╮
│  # │      timestamp       │   run-id    │           workflow           │ status │              test               │
├────┼──────────────────────┼─────────────┼──────────────────────────────┼────────┼─────────────────────────────────┤
│  0 │ 2025-12-02T12:01:06Z │ 19857858903 │ Windows D3D12 AMD DXC        │ FAIL   │ WaveOps/WaveActiveMax.fp64.test │
│  1 │ 2025-12-02T16:08:39Z │ 19865261556 │ Windows D3D12 QC DXC         │ FAIL   │ WaveOps/WaveActiveMax.fp64.test │
│  2 │ 2025-12-02T18:02:48Z │ 19868627852 │ Windows Vulkan QC DXC        │ FAIL   │ WaveOps/WaveActiveMax.fp64.test │
│  3 │ 2025-12-02T12:11:34Z │ 19858143265 │ Windows D3D12 Warp DXC       │ PASS   │ WaveOps/WaveActiveMax.fp64.test │
│  4 │ 2025-12-02T16:01:35Z │ 19865037929 │ Windows D3D12 Intel DXC      │ PASS   │ WaveOps/WaveActiveMax.fp64.test │
│  5 │ 2025-12-02T16:03:51Z │ 19865110582 │ Windows D3D12 NVIDIA DXC     │ PASS   │ WaveOps/WaveActiveMax.fp64.test │
│  6 │ 2025-12-02T16:00:59Z │ 19865018538 │ Windows ARM64 D3D12 Warp DXC │ PASS   │ WaveOps/WaveActiveMax.fp64.test │
│  7 │ 2025-12-02T12:07:36Z │ 19858035043 │ Windows Vulkan AMD DXC       │ PASS   │ WaveOps/WaveActiveMax.fp64.test │
│  8 │ 2025-12-02T16:02:15Z │ 19865059004 │ Windows Vulkan Intel DXC     │ PASS   │ WaveOps/WaveActiveMax.fp64.test │
[...]

PR #730, opened 2026-02-11
Fix WaveActiveMax fp64 stride size

bob80905:
This PR doesn't fully fix the issue https://github.com/llvm/offload-test-suite/issues/541, but maximally fixes the state of the test on this side of things.
The PR should enable Out1 - Out4 to pass / have expected results.
Out5 is seeing -inf, which has been filed as a driver bug here: 61038509.
This failure should be fixed with a driver update, and at that point we can just remove the XFAIL.

### DIFF
--- a/test/WaveOps/WaveActiveMax.fp64.test
+++ b/test/WaveOps/WaveActiveMax.fp64.test
@@ -32,7 +32,6 @@ void main(uint3 tid : SV_GroupThreadID)
             }
         }
     }
-    
     // constant folding case
     Out5[0] = WaveActiveMax(double4(1.5,2.5,3.5,4.5));
 }


test/WaveOps/WaveActiveMax.fp64.test
```diff
@@ -91,20 +90,20 @@ Buffers:
...
   - Name: Out1
     Format: Float64
-    Stride: 4
+    Stride: 8
     # 1 double is 8 bytes, * 4 halves for 4 threads, * 4 thread masks, * 2 value sets
     FillSize: 256  
   - Name: Out2
     Format: Float64
-    Stride: 8
+    Stride: 16
```
bogner:
```suggestion
    Channels: 2
```

PR #731, opened 2026-02-11
[AMD] enable inc_counter_array.test for SPIRV on DXC

farzonl:
This test is passing now on Vulkan with DXC. The XFAIL needs to change to account for this.

### DIFF
--- a/test/Feature/StructuredBuffer/inc_counter_array.test
+++ b/test/Feature/StructuredBuffer/inc_counter_array.test
@@ -53,7 +53,7 @@ DescriptorSets:
 # XFAIL: Intel
 
 # Bug https://github.com/llvm/offload-test-suite/issues/337
-# XFAIL: AMD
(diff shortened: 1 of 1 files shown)


test/Feature/StructuredBuffer/inc_counter_array.test
```diff
@@ -52,9 +52,6 @@ DescriptorSets:
 # Bug https://github.com/llvm/offload-test-suite/issues/376
 # XFAIL: Intel
 
-# Bug https://github.com/llvm/offload-test-suite/issues/337
-# XFAIL: AMD
```
farzonl: This may need to become:
```suggestion
# Bug https://github.com/llvm/offload-test-suite/issues/337
# XFAIL: AMD && Clang && DirectX
```

farzonl:
This is very odd. Test seems flaky. Going to close this pr.

PR #732, opened 2026-02-11
Fix stride size errors revealed by enabling the validation layer

bob80905:
An offload test suite bug was discovered due to incorrect stride size. 
@damyanp presented the idea to run the entire offload test suite with the `-validation-layer` option, 
by setting `-DOFFLOADTEST_ENABLE_VALIDATION=ON` in the cmake invocation, to see if that would catch the incorrect stride size, and any other tests with incorrect stride sizes.
There were plenty of other tests with incorrect stride sizes.
This PR address those incorrect stride sizes all in one go. 
Now, on the AMD machine, when running the whole offload test suite, after this change, there are no more validation-layer crashes on incorrect stride size.


bogner (approved):
For the cases where the buffer contains just a `uint4` or `half4` or some other simple vector type I think it'd be better to set `Channels` rather than `Stride` - this is equivalent but the pipeline logic can do the math so if we were to change the type or something there'd be less chance of introducing a bug.

For the cases where we actually have a struct, like `WaveReadLaneAt.mtx.test` or `WaveReadLaneFirst.128Threads.test` using `Stride` as you already have is correct.

PR #752, opened 2026-02-18
Add test for compute shader semantics

Keenuts:
Add testing for compute shader system semantics on a 2x2x2 dispatch.

### DIFF
--- a/test/Feature/Semantics/ComputeSystemValues.test
+++ b/test/Feature/Semantics/ComputeSystemValues.test
@@ -710,3 +710,6 @@ DescriptorSets:
 # get 3 for the x dimension.
 # See https://github.com/llvm/offload-test-suite/issues/735
 # XFAIL: Metal
+#
+# See https://github.com/llvm/offload-test-suite/issues/764
+# XFAIL: Windows && AMD && DirectX


test/Feature/Semantics/ComputeSystemValues.test
```diff
@@ -0,0 +1,125 @@
+#--- system_values.hlsl
+RWStructuredBuffer<uint> OUT_DTID : register(u0);
+RWStructuredBuffer<uint> OUT_GTID : register(u1);
```
s-perron: Add a comment explaining why you don't simply use uint3. I assume it is because we currently get the alignment wrong.

test/Feature/Semantics/ComputeSystemValues.test
```diff
@@ -0,0 +1,125 @@
...
+      VulkanBinding:
+        Binding: 2
+...
+#--- end
+
+# RUN: split-file %s %t
+# RUN: %dxc_target -T cs_6_0 -Fo %t.o %t/system_values.hlsl
+# RUN: %offloader %t/system_values.yaml %t.o
+
+# Metal is reporting DispatchThreadID value ot 8 for the x dimension.
```
s-perron:
```suggestion
# Metal is reporting DispatchThreadID value of 8 for the x dimension.
```

s-perron (commented):
I think the substance of the test looks good, but the comments don't seem correct.

test/Feature/Semantics/ComputeSystemValues.test
```diff
@@ -0,0 +1,707 @@
...
+# DTID  X  Y  Z
+        1, 0, 0,
+# GTID  X  Y  Z
+        1, 0, 0,
+#  GID  X  Y  Z
+        0, 0, 0,
+#  GI   X  0  0
+        1, 0, 0,
+
+# Dispatch x=0 y=0 z=0 | thread x=0 y=1 z=0
```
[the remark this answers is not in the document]
Keenuts: I made a mistake computing the dispatch coordinates in the python script. Should be fixed.
Yes, dispatch on the API is `2, 2, 2`, but those are coordinates. Added a comment above to explain this.

PR #754, opened 2026-02-18
[HLSL][QC] Modify stride on arrays-16bit.test

joaosaffran:
This patch is modifying the stride for the arrays-16bit.test, in order to fix the vulkan validation errors raised by validation layer.


spall:
this also didn't run any of the tests...

test/Feature/CBuffer/arrays-16bit.test
```diff
@@ -39,16 +42,15 @@ Buffers:
...
       0x0005, 0x0006, 0x0007, 0x0008, 0x5A5A, 0x5A5A, 0x5A5A, 0x5A5A,
+      0x0009, 0x5A5A, 0x5A5A, 0x5A5A, 0x5A5A, 0x5A5A, 0x5A5A, 0x5A5A,
     ]
   - Name: Out
     Format: Hex16
-    # Warp doesn't seem to like buffers with non-multiple of 4 strides, so we
-    # add a small amount of buffer at the end here.
     Stride: 28
-    FillSize: 28
+    FillSize: 64
```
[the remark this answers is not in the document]
joaosaffran: When running the Vulkan validation layer, it is outputting the following error:

```
Validation Error: [ VUID-VkMappedMemoryRange-size-01389 ] | MessageID = 0xee4872d
vkFlushMappedMemoryRanges(): pMemoryRanges[0].size is VK_WHOLE_SIZE and the mapping end (28 = 0 + 28) not a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize (64) and not equal to the end of the memory object (32).
The Vulkan spec states: If size is equal to VK_WHOLE_SIZE, the end of the current mapping of memory must either be a multiple of VkPhysicalDeviceLimits::nonCoherentAtomSize bytes from the beginning of the memory object, or be equal to the end of the memory object (https://docs.vulkan.org/spec/latest/chapters/memory.html#VUID-VkMappedMemoryRange-size-01389)
Objects: 1
    [0] VkDeviceMemory 0x80000000008
```

To fix that, I need to modify the fill size to match 32 or 64. That causes the extra padding in the output below.

test/Feature/CBuffer/arrays-16bit.test
```diff
@@ -39,26 +39,17 @@ Buffers:
       0x3c00, 0x5A5A, 0x5A5A, 0x5A5A, 0x5A5A, 0x5A5A, 0x5A5A, 0x5A5A,
       0x0001, 0x0002, 0x0003, 0x0004, 0x5A5A, 0x5A5A, 0x5A5A, 0x5A5A,
       0x0005, 0x0006, 0x0007, 0x0008, 0x5A5A, 0x5A5A, 0x5A5A, 0x5A5A,
+      0x0009, 0x5A5A, 0x5A5A, 0x5A5A, 0x5A5A, 0x5A5A, 0x5A5A, 0x5A5A,
```
damyanp: Should this last line be here?  This is beyond the end of the data covered by CBArrays, isn't it?

If we need to pad this buffer to be a multiple of something then we should make the whole thing be `0x5A5A`s to be consistent with the values used for the other padding in this test.

test/Feature/CBuffer/arrays-16bit.test
```diff
@@ -42,23 +42,13 @@ Buffers:
     ]
   - Name: Out
     Format: Hex16
-    # Warp doesn't seem to like buffers with non-multiple of 4 strides, so we
-    # add a small amount of buffer at the end here.
     Stride: 28
-    FillSize: 28
+    FillSize: 32
```
spall: it sounds like @damyanp suggestion of fixing this issue in the C++ would be preferable because other tests might have this problem and it isn't clear what the padding is for.

joaosaffran:
I will close this issue, since this is a offloader bug

spall:
> I will close this issue, since this is a offloader bug

By issue you mean PR? the stride still needs to be fixed right?

PR #767, opened 2026-02-25
Remove nan, inf denorm from inputs in saturate.16

joaosaffran:
This patch removes nan, inf and denorm values from inputs to saturate intrinsic call on saturate.16 test. 
fix: https://github.com/llvm/offload-test-suite/issues/558

### DIFF
--- a/test/Feature/HLSLLib/saturate.16.test
+++ b/test/Feature/HLSLLib/saturate.16.test
@@ -34,12 +34,12 @@ Buffers:
   - Name: ExpectedOut # The result we expect
     Format: Float16
     Stride: 8
-    Data: [ 0x0, 0x0, 0x3C00, 0x0, 0x3800, 0x1D1F, 0x3BFE, 0x3C00, 0x3C00, 0x3C00, 0x3C00, 0x3C00 ]
-    #  0, 0, 1, 0, 0.5, 0.005002, 0.9991, 1, 1, 1, 1, 1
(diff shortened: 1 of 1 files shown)


test/Feature/HLSLLib/saturate.16.test
```diff
@@ -25,17 +25,17 @@ Buffers:
...
     Format: Float16
     Stride: 8
     FillSize: 24
   - Name: ExpectedOut # The result we expect
     Format: Float16
     Stride: 8
-    Data: [ 0x0, 0x0, 0x0, 0x0000, 0x0000, 0x0001, 0x3c00, 0x3c00, 0x0000, 0x3800, 0x1d1f, 0x3bfe,]
-    #  0, 0, 0, 0, 0, denorm, 1, 1, 0, 0.5, 0.005, 0.999,
+    Data: [ 0x0, 0x0, 0x3C00, 0x0, 0x3800, 0x1D1F, 0x3BFE, 0x3C00, 0x3C00, 0x3C00, 0x3C00, 0x3C00 ]
+    #  0, 0, 1, 0, 0.5, 0.005002, 0.9991, 1, 1, 1, 1, 1
```
bob80905: I think the proper fix is actually to keep the values between 0 and 1 the same, and instead of adjusting these values to perfectly match, to raise the epsilon to something like .0002. That should account for the slight difference in values. But it doesn't make sense to *expect* these different values. We should expect 0, 0.5, 0.005, 0.999.

PR #930, opened 2026-03-04
Add tests for WaveActiveAllEqual

bob80905:
This PR adds tests for WaveActiveAllEqual.
It does not test matrices.
It tests ints and int2s for wavesizes 32, 64, and 128.
Fixes https://github.com/llvm/offload-test-suite/issues/915

### DIFF
--- a/test/WaveOps/WaveActiveAllEqual.32.test
+++ b/test/WaveOps/WaveActiveAllEqual.32.test
@@ -18,8 +18,6 @@ RWStructuredBuffer<int> Out : register(u12);
 RWStructuredBuffer<int> UOut : register(u13);
 RWStructuredBuffer<int> FOut : register(u14);
 
-// Expect all 4s, all elements will be the same.
-
 [numthreads(4,1,1)]
 void main(uint3 TID : SV_GroupThreadID)
 {
--- a/test/WaveOps/WaveActiveAllEqual.fp64.test
+++ b/test/WaveOps/WaveActiveAllEqual.fp64.test
@@ -6,8 +6,6 @@ StructuredBuffer<double4> In4  : register(t3);
 
 RWStructuredBuffer<int> Out : register(u4);
 
-// Expect all 4s, all elements will be the same.
-
 [numthreads(4,1,1)]
 void main(uint3 TID : SV_GroupThreadID)
 {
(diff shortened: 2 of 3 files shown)


test/WaveOps/WaveActiveAllEqual.32.test
```diff
@@ -0,0 +1,317 @@
...
+// lane writes its result value anded with the existing
+// result in the output buffer.
+// Since we expect all results to be true, output buffers
+// should remain all true at the end.
+
+[numthreads(4,1,1)]
+void main(uint3 TID : SV_GroupThreadID)
+{
+    unsigned int index = 0;
+    Out[index++] &= WaveActiveAllEqual(In[TID.x]);
```
tex3d: Note that this `Out[index++] &=` pattern creates a data race between threads for the UAV accesses. Not that it matters because the shader isn't ever changing the contents of the output buffer, but I suppose, if a thread were to return false, instead of true, that result could be lost in the data race (in theory).

inbelic (approved):

test/WaveOps/WaveActiveAllEqual.32.test
```diff
@@ -0,0 +1,372 @@
...
+StructuredBuffer<float> FIn  : register(t8);
+StructuredBuffer<float2> FIn2  : register(t9);
+StructuredBuffer<float3> FIn3  : register(t10);
+StructuredBuffer<float4> FIn4  : register(t11);
+
+RWStructuredBuffer<int> Out : register(u12);
+RWStructuredBuffer<int> UOut : register(u13);
+RWStructuredBuffer<int> FOut : register(u14);
+
+// Expect all 4s, all elements will be the same.
```
tex3d: NIT: Comment out of date.
```suggestion
// Expect all 1s, all elements will be the same.
```

test/WaveOps/WaveActiveAllEqual.fp64.test
```diff
@@ -0,0 +1,168 @@
+#--- source.hlsl
+StructuredBuffer<double> In  : register(t0);
+StructuredBuffer<double2> In2  : register(t1);
+StructuredBuffer<double3> In3  : register(t2);
+StructuredBuffer<double4> In4  : register(t3);
+
+RWStructuredBuffer<int> Out : register(u4);
+
+// Expect all 4s, all elements will be the same.
```
tex3d: NIT: Comment out of date.

```suggestion
// Expect all 1s, all elements will be the same.
```

PR #935, opened 2026-03-04
Add reconvergence tests

luciechoi:
Part of https://github.com/llvm/llvm-project/issues/136930

The plan is to trigger the pipeline and check all the failing tests to add `XFAIL` statements.

The failing tests should be investigated, added to the `FailedTests.yaml`, and have appropriate TODOs. Until this work is done, I plan to have the test pipeline pass regardless of the reconvergence test results.


Keenuts (changes requested):
Need to spend more time on the ballot and generation implem. Initial comments to begin with.

PR #942, opened 2026-03-05
[Matrix] Add Matrix Subscript tests for cbuffer

### LINKED ISSUE #745: [HLSL][Cbuffer][Matrix] Test `EmitMatrixSubscriptExpr` padding
This is just to track a pr to cover `MatrixSubscriptExpr` padding

farzonl:
fixes #745

### DIFF
--- a/test/Feature/CBuffer/Matrix/MatrixSubscript/row_major_flag_matrix_subscript.f32.test
+++ b/test/Feature/CBuffer/Matrix/MatrixSubscript/row_major_flag_matrix_subscript.f32.test
@@ -1,0 +1,207 @@
+#--- source.hlsl
+RWStructuredBuffer<float> Out_f4x1 : register(u0);
+RWStructuredBuffer<float> Out_f4x2 : register(u1);
+RWStructuredBuffer<float> Out_f2x3 : register(u2);
+RWStructuredBuffer<float> Out_f2x4 : register(u3);
+
+cbuffer example_f4x1 : register(b1) {
+    float4x1 M_f4x1;
+};
+
+cbuffer example_f4x2 : register(b2) {
+    float4x2 M_f4x2;
+};
+
+cbuffer example_f3x2 : register(b3) {
+    float2x3 M_f2x3;
+};
+
+cbuffer example_f2x4 : register(b4) {
+    float2x4 M_f2x4;
+};
+
+[numthreads(1,1,1)]
+void main() {
+  // float4x1
+  Out_f4x1[0] = M_f4x1[0][0];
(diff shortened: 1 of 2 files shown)


test/Feature/CBuffer/Matrix/MatrixSubscript/matrix_subscript.f32.test
```diff
@@ -0,0 +1,207 @@
...
+# Note: NV crashes AMD and Intel Output buffers are empty.
+# Bug https://github.com/microsoft/DirectXShaderCompiler/issues/8218
+# XFAIL: DXC && Vulkan
+
+# Note: clang Vulkan crashes
+# BUG https://github.com/llvm/llvm-project/issues/179879
+# XFAIL: Clang && Vulkan
+
+# RUN: split-file %s %t
+# RUN: %dxc_target -T cs_6_0 -Fo %t.o %t/source.hlsl
```
alsepkow: Wondering if it would be worth testing this with row major and col major layouts?

test/Feature/CBuffer/Matrix/MatrixSubscript/matrix_subscript.f32.test
```diff
@@ -0,0 +1,207 @@
+#--- source.hlsl
+RWStructuredBuffer<float> Out_f1x4 : register(u0);
+RWStructuredBuffer<float> Out_f2x4 : register(u1);
+RWStructuredBuffer<float> Out_f3x2 : register(u2);
+RWStructuredBuffer<float> Out_f4x2 : register(u3);
+
+cbuffer example_f1x4 : register(b1) {
```
alsepkow: Also wondering if we would want to annotate some matrices to be row major and some to be col major?

alsepkow (commented):
@inbelic /Alex :  Leaving a couple comments

test/Feature/CBuffer/Matrix/MatrixSubscript/matrix_subscript.f32.test
farzonl: Yeah doing it at the run line would be easier in another file so went ahead and did that.

Icohedron (approved):

test/Feature/CBuffer/Matrix/MatrixSubscript/matrix_subscript.f32.test
farzonl: I think we can add them it just needs to be understood that we don't plan to turn them on for clang until closer to the end of the year atleast in the case of the keyword cases.

For the flag cases those should work but they should be their own seperate test files.

PR #953, opened 2026-03-11
Add tests for fma 

### LINKED ISSUE #916: Add test for `fma`
- [ ] Test for 16 bit int types (`fma.int16.test`)
- [ ] Test for Half type (`fma.fp16.test`)
- [ ] Test for 32 bit types (`fma.32.test`)
- [ ] Test for 64 bit int types (`fma.int64.test`)
- [ ] Test for Double type (`fma.fp64.test`)

Matrix tests are also needed:
<br>

A .test file should contain 
```
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# RUN: split-file %s %t
```

NeKon69:
This PR adds double tests for `fma` function. Only double is applicable to this function as per [official api](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-fma)

Closes #916
Related to [llvm/llvm-project/#185304](https://github.com/llvm/llvm-project/pull/185304)

Assisted-by: GPT-5.4 (opencode)


farzonl:
> Also some of the tests seem to fail because there isn't an `fma` function. Don't really know what to do with that

Sorry will do a review of this in the morning. But yeah xfail for clang for now

test/Feature/HLSLLib/fma.fp64.test
```diff
@@ -0,0 +1,175 @@
+#--- source.hlsl
+
+StructuredBuffer<double4> A : register(t0);
+StructuredBuffer<double4> B : register(t1);
+StructuredBuffer<double4> C : register(t2);
+
+RWStructuredBuffer<double> Out : register(u3);
```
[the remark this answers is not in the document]
kmpeng: I think it's fine to have only one Out buffer if we follow the same format as other tests that have been written. See my other comment.

kmpeng (commented):
You should take [Matrix] out of the title since this PR is covering more than just the matrix tests.

kmpeng (commented):
The structure of the tests lgtm, but I have some problems with the values being used. I'm also not very familiar with matrices, so I'll defer to other people for more feedback on those (@farzonl ?)

FYI a lot of the failing checks on this PR have been fixed recently. Could you update the branch so it's easier to see the ones this PR is affecting?

joaosaffran (changes requested):

test/Feature/HLSLLib/fma.test
```diff
@@ -0,0 +1,104 @@
...
+      VulkanBinding:
+        Binding: 3
+#--- end
+
+# Unimplemented https://github.com/llvm/llvm-project/issues/99117
+# XFAIL: Clang
+
+# Bug: https://github.com/llvm/offload-test-suite/issues/1000
+# XFAIL: arm64 && WARP && DXC
+
```
kmpeng: The constant folding test is failing for DXC Vulkan on Intel. Could you file an issue against this repo and XFAIL it?
```
# Bug <link-to-issue>
# XFAIL: Intel && Vulkan && DXC
```

kmpeng (approved):
LGTM, but again you should probably get more feedback on the matrix test

test/Feature/HLSLLib/fma.matrix.test
```diff
@@ -0,0 +1,144 @@
...
+        Space: 0
+      VulkanBinding:
+        Binding: 3
+#--- end
+
+# Unimplemented https://github.com/llvm/llvm-project/issues/99117
+# XFAIL: Clang
+
+# Bug: https://github.com/llvm/offload-test-suite/issues/1000
+# XFAIL: arm64 && WARP && DXC
```
kmpeng: Seems like QC WARP still fails on Clang even with the implementation in, so I'm guessing this bug actually applies to both DXC and Clang.

To confirm, we probably want to do a run with just the Clang XFAIL removed so we can get the QC clang-warp-d3d12 failing logs. Then if it turns out to be the same bug, we can update the issue and remove `&& DXC` from this XFAIL.

test/Feature/HLSLLib/fma.test
```diff
@@ -0,0 +1,107 @@
...
+      Kind: RWStructuredBuffer
+      DirectXBinding:
+        Register: 3
+        Space: 0
+      VulkanBinding:
+        Binding: 3
+#--- end
+
+# Unimplemented https://github.com/llvm/llvm-project/issues/99117
+# XFAIL: Clang
```
kmpeng: Same as above

test/Feature/HLSLLib/fma.test
```diff
@@ -0,0 +1,107 @@
...
+
+# Bug https://github.com/llvm/offload-test-suite/issues/1000
+# XFAIL: arm64 && WARP && DXC
```
kmpeng: Same as above

test/Feature/HLSLLib/fma.test
```diff
@@ -0,0 +1,104 @@
...
+
+# Bug https://github.com/llvm/offload-test-suite/issues/1000
+# XFAIL: arm64 && WARP
+
+# Bug https://github.com/llvm/offload-test-suite/issues/1026
+# XFAIL: Intel && Vulkan && DXC
+
+# REQUIRES: Double
+# RUN: split-file %s %t
+# RUN: %dxc_target -Gis -HV 202x -T cs_6_5 -Fo %t.o %t/source.hlsl
```
kmpeng: Shouldn't need this since we're not testing NaN and INF values
```suggestion
# RUN: %dxc_target -HV 202x -T cs_6_5 -Fo %t.o %t/source.hlsl
```

farzonl (approved):

joaosaffran (approved):

kmpeng:
@NeKon69 This should be ready to go. Do you need someone to merge it in for you?

You should probably also update the branch to make sure nothing has broken as it's been a while

NeKon69:
Yes I need someone to merge this for me, just updated the branch

PR #957, opened 2026-03-12
Added WaveOps tests for QuadReadAcrossX

kcloudy0717:
PR for `QuadReadAcrossX` tests, resolves https://github.com/llvm/offload-test-suite/issues/882.

### DIFF
--- a/test/WaveOps/QuadReadAcrossX.32.test
+++ b/test/WaveOps/QuadReadAcrossX.32.test
@@ -343,6 +343,9 @@ DescriptorSets:
 # Bug: https://github.com/llvm/offload-test-suite/issues/986
 # XFAIL: Intel && Vulkan
 
+# Bug: https://github.com/llvm/offload-test-suite/issues/989
+# XFAIL: Metal
+
 # RUN: split-file %s %t
 # RUN: %dxc_target -T cs_6_5 -Fo %t.o %t/source.hlsl
 # RUN: %offloader %t/pipeline.yaml %t.o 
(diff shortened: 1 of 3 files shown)


test/WaveOps/QuadReadAcrossX.test
```diff
@@ -0,0 +1,60 @@
+#--- source.hlsl
+StructuredBuffer<float> In : register(t0);
+RWStructuredBuffer<float> Out : register(u1);
+
+[numthreads(2, 2, 1)]
+void main(uint3 dtid : SV_DispatchThreadID) {
+  uint index = dtid.y * 2 + dtid.x;
```
bob80905: Thanks, the data type tests are looking great.
This extra `.test` file that isn't formally mentioned in the issues was added to several of the past wave builtins because of their convergent attributes, to test convergence in control flow. I would argue this test doesn't contribute much more to coverage than the previous tests. 
If you would, please update this test so that it tests QuadReadAcrossX on control flow cases that alter lane activity
e.g., if (index < 2) QuadReadAcrossX ... 
Note, the spec mentions that QuadReadAcrossX will return undefined results if the lane across from it in the X direction is not active. So you don't want index < 1 or index < 3. 
Maybe an index >= 2 condition as well, to verify that works too.

kcloudy0717: I was wondering why some of the previous wave intrinsic tests had this extra test, now it makes sense that it tests convergence in control flows. Thanks for the clarification! I'll go ahead and update this test according to the feedback.

PR #971, opened 2026-03-13
Add WaveActiveBitXor tests

bob80905:
This PR adds tests for WaveActiveBitXor.
WaveActiveBitXor only accepts uint and uint64 types.
The PR also adds a control flow test, since this operation is convergent.
Fixes https://github.com/llvm/offload-test-suite/issues/896


tex3d (approved):
Looks fine, but had a suggestion to make the constant case less of a no-op test, and suggest trying 3 active threads for a non-constant scalar.

PR #973, opened 2026-03-13
Add WaveActiveBitOr tests

bob80905:
This PR adds tests for WaveActiveBitOr.
WaveActiveBitOr only accepts uint and uint64 types.
The PR also adds a control flow test, since this operation is convergent.
Fixes https://github.com/llvm/offload-test-suite/issues/92
Assisted by: Claude Opus 4.6


bogner (approved):
A lot of the variables in the tests aren’t following the llvm capitalization conventions (TID, V, R rather than tid, v, r), so it would be nice to make that consistent. Otherwise, LGTM!

PR #975, opened 2026-03-14
Make element-alignment XFAILs specific to Vulkan

bogner:
Now that llvm/llvm-project#123968 is resolved, these all pass with clang for DirectX.


kmpeng (approved):
FYI you didn't actually link the issue in the description (guessing it's in llvm-project)

PR #997, opened 2026-03-20
[Metal] Synchronize managed resources in Metal backend

kcloudy0717:
I'm investigating some tests that are failing on Metal backend where majority of the outputs of the tests is empty (all zeroes), and I noticed buffer/texture buffer creations in the Metal backend are using [MTLStorageModeManaged](https://developer.apple.com/documentation/metal/mtlstoragemode/managed).

In this [documentation](https://developer.apple.com/documentation/metal/synchronizing-a-managed-resource-in-macos):
> After encoding a GPU modification, encode a [synchronize(resource:)](https://developer.apple.com/documentation/metal/mtlblitcommandencoder/synchronize(resource:)) command. This command updates the entire buffer and keeps it synchronized. Before executing this command, the modified buffer’s data on the CPU is in an undefined state.

it says GPU modifications to managed resource should use `MTLBlitCommandEncoder::synchronizeResource` function to properly ensure the CPU sees the change as well.

My local mac machine isn't properly configured to test this change so opening a PR to see how workflow reacts.

### DIFF
--- a/lib/API/MTL/MTLDevice.cpp
+++ b/lib/API/MTL/MTLDevice.cpp
@@ -462,6 +462,21 @@ class MTLDevice : public offloadtest::Device {
   }
 
   llvm::Error executeCommands(InvocationState &IS) {
+    // Synchronize managed resources back to the CPU so that any updates we made
+    // on the GPU is visible.
+    // See:
+    // https://developer.apple.com/documentation/metal/synchronizing-a-managed-resource-in-macos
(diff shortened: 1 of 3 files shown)


bogner:
Sorry, I somehow missed this PR entirely. I think it would be fine to do this, but I also can't really see us caring about running the offload-test-suite on intel macs, so it's extremely low value. Given that, and the fact that all of the code around here has changed significantly so we'd need to redo it anyway, I'm going to go ahead and just close this PR.

PR #1007, opened 2026-03-24
Implement an abstract Fence type to expose waiting for command buffer completion.

manon-traverse:
The Fence type is modeled around the DX12 Fence. The closest equivalent in Vulkan is a Timeline Semaphore. These are widely supported and included in Vulkan 1.2. Metal provides the same functionality through the SharedEvent type.

The abstract interface allows us to wait for GPU work to complete by waiting on a signal value using:
```C++
Fence->waitForCompletion(SignalValue);
```

Signaling the value still requires downcasting because queue submission is not done via an abstract interface yet.

### DIFF
--- a/lib/API/VK/Device.cpp
+++ b/lib/API/VK/Device.cpp
@@ -424,7 +424,7 @@ public:
     uint64_t Value = 0;
     [[maybe_unused]] const VkResult Ret =
         vkGetSemaphoreCounterValue(Device, Semaphore, &Value);
-    assert(!Ret);
+    assert(!Ret && "vkGetSemaphoreCounterValue failed but should never fail.");
     return Value;
   }
 


lib/API/VK/Device.cpp
```diff
@@ -1110,32 +1178,41 @@ class VulkanDevice : public offloadtest::Device {
...
+    TimelineSubmitInfo.sType = VK_STRUCTURE_TYPE_TIMELINE_SEMAPHORE_SUBMIT_INFO;
+    TimelineSubmitInfo.signalSemaphoreValueCount = 1;
+    TimelineSubmitInfo.pSignalSemaphoreValues = &CurrentCounter;
+
     VkSubmitInfo SubmitInfo = {};
     SubmitInfo.sType = VK_STRUCTURE_TYPE_SUBMIT_INFO;
+    SubmitInfo.pNext = &TimelineSubmitInfo;
     SubmitInfo.commandBufferCount = 1;
     SubmitInfo.pCommandBuffers = &IS.CmdBuffer;
     SubmitInfo.pWaitDstStageMask = &WaitMask;
```
MarijnS95: Random find: `pWaitDstStageMask` is used without `pWaitSemaphores`/`waitSemaphoreCount`, so this has no effect 😛 

Maybe that adds unexpected and implicit synchronisation issues because this command won't actually "wait for `WaitMask`", as soon as we drop the wait for idle and have in-flight command buffers?

lib/API/DX/Device.cpp
```diff
@@ -342,12 +386,7 @@ class DXDevice : public offloadtest::Device {
     ComPtr<ID3D12PipelineState> PSO;
     ComPtr<ID3D12CommandAllocator> Allocator;
     ComPtr<ID3D12GraphicsCommandList> CmdList;
-    ComPtr<ID3D12Fence> Fence;
-#ifdef _WIN32
-    HANDLE Event;
-#else // WSL
-    int Event;
-#endif
+    std::shared_ptr<offloadtest::Fence> Fence;
```
bogner: Does this need to be `shared_ptr`? As far as I can tell each fence is uniquely owned by a particular DXQueue.

lib/API/DX/Device.cpp
```diff
@@ -289,6 +289,50 @@ class DXBuffer : public offloadtest::Buffer {
       : Buffer(Buffer), Name(Name), Desc(Desc), SizeInBytes(SizeInBytes) {}
 };
 
+class DXFence : public offloadtest::Fence {
+public:
+  std::string Name;
+  ComPtr<ID3D12Fence> Fence;
+  HANDLE Event;
```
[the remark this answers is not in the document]
MarijnS95: @bogner on a related note, is the CI build-testing the WSL path (Linux with D3D12 headers available)?

Issue #611, opened 2025-12-08
Fix Vulkan offload test suite setup to allow for wave size specification


bob80905:
Right now, Vulkan ignores the specified WaveSize attribute in the HLSL.
In the device creation code, there is no mention of VK_EXT_subgroup_size_control to set the wave size based on the attribute in the HLSL.
So, when Vulkan is the IR target, the device will just use the default wave size. 

This issue tracks updating the test infrastructure so that the wave size specified in the HLSL is properly referenced, or an error is raised if the device doesn't support it.

---

We have [various tests](https://github.com/search?q=repo%3Allvm%2Foffload-test-suite%20WaveSize%20path%3A*.test&type=code) that rely on the WaveSize attribute to control the wave size ("subgroup size" in Vulkan terms). Vulkan/SPIRV doesn't have an equivalent to this attribute, but instead has the VK_EXT_subgroup_size_control extension that can be used for the same purpose.

For this task we should:
* find a way to expose this via the pipeline yaml
* expose supported subgroup sizes that can be used in REQUIRES clauses.
* update the existing tests to use it

PR #1012, opened 2026-03-25
Mark tests that use WaveSize attribute as unsupported on Vulkan & Metal

damyanp:
The WaveSize attribute is currently only fully supported on D3D. Some tests may be passing, or not running, by accident due to other requirements. This change explicitly marks these tests as unsupported on Metal and Vulkan.

#611 tracks adding the required support to Vulkan to be able to enable these tests.

Assisted-by: Copilot

### DIFF
--- a/test/WaveOps/WaveActiveAllEqual.Wave128.test
+++ b/test/WaveOps/WaveActiveAllEqual.Wave128.test
@@ -325,6 +325,7 @@ DescriptorSets:
 #--- end
 
 # REQUIRES: WaveSize_128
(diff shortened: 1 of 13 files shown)


bogner (commented):
There's some argument that it's obvious, since the WaveSize `REQUIRES` immediately precedes each `UNSUPPORTED`, but I think our policy of including a comment for any `UNSUPPORTED` or `XFAIL` should apply here and we should explicitly comment that these two platforms don't do WaveSize.

PR #1014, opened 2026-03-25
XFAIL for QuadReadAcrossY.fp64.test on WARP D3D12 DXC config

kcloudy0717:
WARP D3D12 DXC tests are failing for QuadReadAcrossY.fp64.test. I filed an issue for it, opening a PR for add XFAIL to the test.


inbelic (approved):

Icohedron (approved):

inbelic:
It looks like if you are targeting a `WARP` device then the machine type doesn't get set. So `WARP && AMD` is never true. Will fix this tomorrow.

Issue #1013, opened 2026-03-25
[WARP][DXC][D3D12] `QuadReadAcrossY.fp64.test` failure for WARP on x64 architecture


kcloudy0717:
`WaveOps/QuadReadAcrossY.fp64.test` is failing on `WARP-AMD` for both dxc and clang. In both cases the behaviour is the same: it returns all zeros (appears to do nothing).

This is due to the following error, visible when run with `-debug-layer -validation-layer` specified:

```
gpu-exec: error: Failed to map result.
```

[DXC log](https://github.com/llvm/offload-test-suite/issues/1013). [Clang log](https://github.com/llvm/offload-test-suite/issues/1013).

```
******************** TEST 'OffloadTest-warp-d3d12 :: WaveOps/QuadReadAcrossY.fp64.test' FAILED ********************
Exit Code: 1

Command Output (stdout):
--
# RUN: at line 134
c:\actions-runner\_work\offload-test-suite\offload-test-suite\llvm-project\build\bin\split-file.exe C:\actions-runner\_work\offload-test-suite\offload-test-suite\OffloadTest\test\WaveOps\QuadReadAcrossY.fp64.test C:\actions-runner\_work\offload-test-suite\offload-test-suite\llvm-project\build\tools\OffloadTest\test\warp-d3d12\WaveOps\Output\QuadReadAcrossY.fp64.test.tmp
# executed command: 'c:\actions-runner\_work\offload-test-suite\offload-test-suite\llvm-project\build\bin\split-file.exe' 'C:\actions-runner\_work\offload-test-suite\offload-test-suite\OffloadTest\test\WaveOps\QuadReadAcrossY.fp64.test' 'C:\actions-runner\_work\offload-test-suite\offload-test-suite\llvm-project\build\tools\OffloadTest\test\warp-d3d12\WaveOps\Output\QuadReadAcrossY.fp64.test.tmp'
# RUN: at line 135
[...]

PR #1017, opened 2026-03-25
Modify XFail for firstbit low and high on clang

joaosaffran:
Modify XFAIL on firstbit high and low


inbelic:
This is passing on main now too: https://github.com/llvm/offload-test-suite/issues/1013. It would be good to get this into review soong.

farzonl (approved):

PR #1020, opened 2026-03-26
Introduce a new `Texture` type with DX,VK,MTL implementations.

EmilioLaiso:
## Summary

- **Introduced `Texture.h`** with a shared `TextureCreateDesc`, `TextureFormat` enum, `TextureUsage` flags, `ClearValue` variant (`ClearColor` / `ClearDepthStencil`), and `validateTextureCreateDesc` returning specific `llvm::Error` messages
- **Added `createTexture` to all three backends** (DX, Vulkan, Metal) as a virtual method on `Device`, handling resource creation, optimized clear values, and format/usage validation
- **Unified render target creation** via `createRenderTargetFromCPUBuffer(Device&, CPUBuffer&)` free function — all backends now route through `createTexture` instead of manually calling API-specific resource creation
- **Unified depth/stencil creation** via `createDefaultDepthStencilTarget(Device&, Width, Height)` free function with `D32FloatS8Uint` format — added depth testing to DX and Metal backends (previously only Vulkan had it)
- **Readback buffers** created through the virtual `createBuffer` API with `MemoryLocation::GpuToCpu`; Metal refactored from `StorageModeShared` texture with `getBytes()` to `StorageModePrivate` texture + blit copy to readback buffer (this is to align functionality between backends in preparation of api-agnostic code-flow. Sacrifices had to be made 😛)
- **Views on textures** — RTV/DSV descriptor heaps (DX) and `VkImageView` (VK) are now created at `createTexture` time for `RenderTarget`/`DepthStencil` usage, removing view creation boilerplate from pipeline setup
- Render target and readback buffer are stored directly as `shared_ptr<VulkanTexture>` / `shared_ptr<VulkanBuffer>` on `InvocationState`
- **Clear values read from texture descriptors** in all backends instead of being hardcoded
- **Added `toTextureFormat(DataFormat, Channels)`** as a bridge function while refactoring, with `validateTextureDescMatchesCPUBuffer` to catch mismatches between the old and new description systems
- **Split Metal storage mode helpers** into `getMetalTextureStorageMode` and `getMetalBufferResourceOptions` because `GpuToCpu` maps differently for textures (Managed) vs buffers (Shared)

## Future direction
Added for context to better understand some of the changes in this PR.
- **SRV/UAV views on textures** — currently not stored on the texture because they would require a descriptor heap allocation scheme
- Work towards eliminating `DataFormat + Channels` dual description. 
- **Route VK vertex buffer through `createBuffer`** — still uses the old `ResourceRef`/`BufferRef` system
- Abstract readback copy
- **Unify `InvocationState`** across backends

### DIFF
--- a/include/API/FormatConversion.h
+++ b/include/API/FormatConversion.h
@@ -137,7 +137,7 @@ validateTextureDescMatchesCPUBuffer(const TextureCreateDesc &Desc,
         "TextureCreateDesc mip levels %u does not match CPUBuffer mip "
         "levels %d.",
         Desc.MipLevels, Buf.OutputProps.MipLevels);
-  const uint32_t TexelSize = getFormatSize(Desc.Format);
+  const uint32_t TexelSize = getFormatSizeInBytes(Desc.Format);
   if (Buf.Stride > 0 && static_cast<uint32_t>(Buf.Stride) != TexelSize)
     return llvm::createStringError(
         std::errc::invalid_argument,
--- a/include/API/Resources.h
+++ b/include/API/Resources.h
@@ -92,7 +92,7 @@ inline llvm::StringRef getFormatName(Format Format) {
 }
 
 // Returns the size in bytes of a single texel/element for the given format.
-inline uint32_t getFormatSize(Format Format) {
+inline uint32_t getFormatSizeInBytes(Format Format) {
   switch (Format) {
   case Format::R16Sint:
   case Format::R16Uint:
--- a/lib/API/MTL/MTLDevice.cpp
+++ b/lib/API/MTL/MTLDevice.cpp
@@ -629,7 +629,8 @@ class MTLDevice : public offloadtest::Device {
 
     // Blit the render target into the readback buffer for CPU access.
     MTL::BlitCommandEncoder *Blit = IS.CB->CmdBuffer->blitCommandEncoder();
-    const size_t ElemSize = getFormatSize(IS.FrameBufferTexture->Desc.Format);
+    const size_t ElemSize =
+        getFormatSizeInBytes(IS.FrameBufferTexture->Desc.Format);
     const size_t RowBytes = Width * ElemSize;
     Blit->copyFromTexture(IS.FrameBufferTexture->Tex, 0, 0,
                           MTL::Origin(0, 0, 0), MTL::Size(Width, Height, 1),


lib/API/DX/Device.cpp
```diff
@@ -457,38 +480,116 @@ class DXDevice : public offloadtest::Device {
...
-      HeapType = D3D12_HEAP_TYPE_READBACK;
-      break;
-    }
-
+    // Readback heaps do not support UAV access.
     const D3D12_RESOURCE_FLAGS Flags =
-        D3D12_RESOURCE_FLAG_ALLOW_UNORDERED_ACCESS;
+        HeapType == D3D12_HEAP_TYPE_READBACK
+            ? D3D12_RESOURCE_FLAG_NONE
+            : D3D12_RESOURCE_FLAG_ALLOW_UNORDERED_ACCESS;
```
MarijnS95: Fwiw in our codebase we use a custom heap that might not place this restriction on buffers.

lib/API/VK/Device.cpp
```diff
@@ -543,9 +570,9 @@ class VulkanDevice : public offloadtest::Device {
 
     // FrameBuffer associated data for offscreen rendering.
     VkFramebuffer FrameBuffer = VK_NULL_HANDLE;
-    ResourceBundle FrameBufferResource = {VK_DESCRIPTOR_TYPE_STORAGE_IMAGE, 0,
-                                          nullptr};
-    ImageRef DepthStencil = {0, 0, 0};
+    std::shared_ptr<VulkanTexture> RenderTarget;
+    std::shared_ptr<VulkanBuffer> RTReadback;
+    std::shared_ptr<VulkanTexture> DepthStencil;
```
manon-traverse, line 624: Note: we should try to store the platform-agnostic pointer here and downcast where needed, as the goal is to use these only in the platform-abstract way.

I am fine with keeping it this way for this PR, but it's something we should do in the future. Perhaps once we have an agnostic way of binding resources?

include/API/FormatConversion.h
```diff
@@ -0,0 +1,159 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// Transitional helpers for converting between the legacy DataFormat + Channels
+// description system and the unified Format enum. This file should be deleted
+// once the pipeline is fully migrated to use Format directly.
```
MarijnS95, line 11: Do we have a tracking issue for this conversion and removal?

lib/API/MTL/MTLResources.h
```diff
@@ -0,0 +1,152 @@
...
+#include "Metal/Metal.hpp"
+
+namespace offloadtest {
+
+// Metal requires different storage modes for textures and buffers.
+// Textures use Managed for GpuToCpu because Shared textures are not available
+// on discrete GPUs and lack hardware compression/tiling optimizations.
+// Buffers use Shared for GpuToCpu because they are linear memory with no
+// layout concerns, and Shared avoids the explicit synchronizeResource step
+// that Managed requires.
```
MarijnS95: Move this comment to the function that returns the Buffer resource storage mode?

I was really conflicted reading this with only the function below in mind, having to read it from the start again to realize that this comment belongs on the function below.

lib/API/MTL/MTLResources.h
```diff
@@ -0,0 +1,152 @@
...
+  case Format::RG32Sint:
+    return MTL::PixelFormatRG32Sint;
+  case Format::RG32Uint:
+    return MTL::PixelFormatRG32Uint;
+  case Format::RG32Float:
+    return MTL::PixelFormatRG32Float;
+  // Metal has no 3-component pixel format.
+  // RGB32Float is only valid as a vertex format.
+  case Format::RGB32Float:
+    llvm_unreachable("RGB32Float has no Metal pixel format equivalent");
```
MarijnS95, line 81: Unreachable or an actual error? Perhaps because this returns "false" out of `isTextureCompatible()` it should "never" be used in any call site to/through `getMetalPixelFormat()`?

lib/API/VK/VKResources.h
```diff
@@ -0,0 +1,93 @@
...
+inline VkMemoryPropertyFlags getVulkanMemoryFlags(MemoryLocation Location) {
+  switch (Location) {
+  case MemoryLocation::GpuOnly:
+    return VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT;
+  case MemoryLocation::CpuToGpu:
+    return VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT |
+           VK_MEMORY_PROPERTY_HOST_COHERENT_BIT;
+  case MemoryLocation::GpuToCpu:
+    return VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT |
+           VK_MEMORY_PROPERTY_HOST_CACHED_BIT;
```
MarijnS95, line 30: So we'll always do `vkInvalidateMappedMemoryRanges()` before reading back `GpuToCpu` memory? Otherwise this heap is required to have `VK_MEMORY_PROPERTY_HOST_COHERENT_BIT` too.

lib/API/Device.cpp
```diff
@@ -45,3 +46,26 @@ offloadtest::initializeDevices(const DeviceConfig Config) {
...
+  Desc.Format = *TexFmtOrErr;
+  Desc.Width = Buf.OutputProps.Width;
+  Desc.Height = Buf.OutputProps.Height;
+  Desc.MipLevels = 1;
+  Desc.OptimizedClearValue = ClearColor{};
+
+  if (auto Err = validateTextureDescMatchesCPUBuffer(Desc, Buf))
+    return Err;
+
+  return Dev.createTexture("RenderTarget", Desc);
```
MarijnS95, line 79: Nit: feels like the name should come from the caller?

lib/API/DX/Device.cpp
```diff
@@ -293,6 +295,27 @@ class DXBuffer : public offloadtest::Buffer {
       : Buffer(Buffer), Name(Name), Desc(Desc), SizeInBytes(SizeInBytes) {}
 };
 
+class DXTexture : public offloadtest::Texture {
+public:
+  ComPtr<ID3D12Resource> Resource;
+  // TODO:
+  // RTV/DSV views own a dedicated single-descriptor heap and are created at
```
MarijnS95: Nit: RTV/DSV already contains "view" making this hard to "read out loud" 😉

lib/API/DX/Device.cpp
```diff
@@ -479,33 +479,32 @@ class DXDevice : public offloadtest::Device {
...
-      HeapType = D3D12_HEAP_TYPE_UPLOAD;
-      break;
-    case MemoryLocation::GpuToCpu:
-      HeapType = D3D12_HEAP_TYPE_READBACK;
-      break;
-    }
-
+    // As per the readback heap docs
+    // > Resources in this heap must be created with
+    // > D3D12_RESOURCE_STATE_COPY_DEST, and cannot be changed away from this.
```
MarijnS95: Nit: looks like this comment belongs ~10 lines down, where you actually set `COPY_DEST`. Here you decide on the resource flags instead.

lib/API/VK/Device.cpp
```diff
@@ -2400,17 +2446,15 @@ class VulkanDevice : public offloadtest::Device {
...
       Range.size = VK_WHOLE_SIZE;
-      const ResourceRef &ResRef = IS.FrameBufferResource.ResourceRefs[0];
+      Range.memory = IS.RTReadback->Memory;
 
       void *Mapped = nullptr; // NOLINT(misc-const-correctness)
-      vkMapMemory(Device, ResRef.Host.Memory, 0, VK_WHOLE_SIZE, 0, &Mapped);
-
-      Range.memory = ResRef.Host.Memory;
+      vkMapMemory(Device, IS.RTReadback->Memory, 0, VK_WHOLE_SIZE, 0, &Mapped);
       vkInvalidateMappedMemoryRanges(Device, 1, &Range);
```
MarijnS95, line 2421: Nice

lib/API/Device.cpp
```diff
@@ -69,3 +69,17 @@ offloadtest::createRenderTargetFromCPUBuffer(Device &Dev,
...
+  TextureCreateDesc Desc = {};
+  Desc.Location = MemoryLocation::GpuOnly;
+  Desc.Usage = TextureUsage::DepthStencil;
+  Desc.Format = Format::D32FloatS8Uint;
+  Desc.Width = Width;
+  Desc.Height = Height;
+  Desc.MipLevels = 1;
+  Desc.OptimizedClearValue = ClearDepthStencil{1.0f, 0};
+
+  return Dev.createTexture("DepthStencil", Desc);
```
MarijnS95, line 94: Same, could the caller pass this string? (EDIT: Unless these helpers are scheduled to disappear soon too)

lib/API/MTL/MTLResources.h
EmilioLaiso: Indeed, `llvm_unreachable` because "you shouldn't get to that point"

lib/API/Device.cpp
EmilioLaiso: This function is basically just wrapping some hardcoded code in the backend. Soon-ish it will disappear altogether.

lib/API/Device.cpp
EmilioLaiso: Indeed scheduled to disappear, they're just wrapping some code that was inlined to help create a more readable flow during refactoring and to achieve "flow alignment" between the 3 backends

lib/API/VK/VKResources.h
manon-traverse: According to the spec, there is always a `HOST_COHERENT | HOST_VISIBLE` memory type, so we should definitely enforce using the `HOST_COHERENT` bit.

include/API/Device.h
```diff
@@ -117,6 +103,16 @@ initializeMetalDevices(const DeviceConfig Config,
 llvm::Expected<llvm::SmallVector<std::unique_ptr<Device>>>
 initializeDevices(const DeviceConfig Config);
 
+// Creates a render target texture using the format and dimensions from a
+// CPUBuffer. Does not upload the buffer's data — only uses its description to
+// configure the texture.
+llvm::Expected<std::shared_ptr<Texture>>
```
s-perron, line 114: Do we want to use `std::shared_ptr`? I know a lot of projects avoid using it because it is slow.

include/API/Resources.h
```diff
@@ -0,0 +1,237 @@
...
+  case Format::D32Float:
+    return "D32Float";
+  case Format::D32FloatS8Uint:
+    return "D32FloatS8Uint";
+  }
+  llvm_unreachable("All Format cases handled");
+}
+
+// Returns the size in bytes of a single texel/element for the given format.
+inline uint32_t getFormatSize(Format Format) {
```
s-perron: Could you add the unit in the function name, not just the comments? Otherwise, I will not know the units when I see a use, and I'll have to look up the comment to find it out.


```suggestion
inline uint32_t getFormatSizeInBytes(Format Format) {
```

lib/API/VK/Device.cpp
```diff
@@ -592,9 +619,9 @@ class VulkanDevice : public offloadtest::Device {
 
     // FrameBuffer associated data for offscreen rendering.
     VkFramebuffer FrameBuffer = VK_NULL_HANDLE;
-    ResourceBundle FrameBufferResource = {VK_DESCRIPTOR_TYPE_STORAGE_IMAGE, 0,
-                                          nullptr};
-    ImageRef DepthStencil = {0, 0, 0};
+    std::shared_ptr<VulkanTexture> RenderTarget;
+    std::shared_ptr<VulkanBuffer> RTReadback;
+    std::shared_ptr<VulkanTexture> DepthStencil;
```
s-perron, line 624: I'm use to the [Google style guide](https://google.github.io/styleguide/cppguide.html#Ownership_and_Smart_Pointers). We generally do not allow std::shared_ptr unless there is a very good reason for it.

I've only looked at the code quickly. However, would it be possible to make the `std::unqiue_ptr` to say that they are owned by the `InvocationState`?

I'll don't have strong opinions for this project, but we should make a conscience decision.

include/API/Device.h
manon-traverse: I would recommend using the shared_ptr. We want to be able to use reference counting to determine if resources need to be kept alive so that we can free them once they are no longer being used on the GPU timeline.

IMO the overhead of shared_ptr is a bit overblown (an atomic increment on copy). If we ever run into a case where we need thousands of textures that need to be kept alive we can look at alternative solutions for those use cases. However, in my experience, that case is only there for bindless resources where you are going to need an alternative anyway.

manon-traverse:
This PR fails some CI tests, but only after changing the name of a getter (unrelated to any errors). It fails on Mac because that machine is out of storage. Intel-clang failure is one that also happens on main, I am going to assume this is because of a change in the llvm-project repo.

Because these reasons have nothing to do with the PR and enough reviewing and applying feedback has happened I am gonig to merge this PR so we can move forward. If any more issues arise, we can address those in follow-up PRs.

include/API/Device.h
s-perron: I think we need to revisit this. I'm not sure how you count "thousands of textures", but, as the number of tests increases, we could easily reach that over all of the tests. I'll let others like @bogner and @llvm-beanz give their opinions.

I'll go along with what the community thinks.

damyanp: @manon-traverse - can you file an issue to follow up on this please?  Ideally, we'd get consensus before the PR is merged, but we as long as we figure it all out it's ok to that post-merge.

bogner: I agree with Steven here that shared_ptr should be avoided unless absolutely necessary. Regardless of whether or not the overhead of shared_ptr will become an issue for us here, relying on reference counting obscures object lifetimes and encourages messy design patterns. It is usually possible to have clear ownership of the object, which makes the code easier to understand and reason about, and generally leads to a more robust architecture.

PR #1033, opened 2026-03-30
Add CommandBuffer abstraction with backend-specific downcasting

MarijnS95:
Command buffer creation and management was previously spread across each backend's `executeProgram()` with no shared interface, making it impossible to manage command buffers from backend-agnostic code. This introduces a `CommandBuffer` base class on `Device` so that higher-level code can create and pass around command buffers without knowing the backend. Per-object allocator/pool ownership also prepares for future async execution where multiple command buffers may be in-flight with independent lifetimes.

- DX: `DXCommandBuffer` owns Allocator, CmdList, Fence, Event
- VK: `VKCommandBuffer` owns CmdPool, CmdBuffer; each submission creates a new CommandBuffer for independent lifetime management
- MTL: `MTLCommandBuffer` wraps `MTL::CommandBuffer`

`Device::createCommandBuffer()` returns `Expected<unique_ptr<CommandBuffer>>` with a default "not supported" implementation.

## Test plan
- [ ] Verify DX backend compiles and passes existing tests
- [x] Verify VK backend compiles and passes existing tests
- [x] Verify MTL backend compiles and passes existing tests

### DIFF
--- a/include/API/CommandBuffer.h
+++ b/include/API/CommandBuffer.h
@@ -1,0 +1,44 @@
+//===- CommandBuffer.h - Offload Command Buffer API -----------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef OFFLOADTEST_API_COMMANDBUFFER_H
+#define OFFLOADTEST_API_COMMANDBUFFER_H
+
+#include "API/API.h"
+
+#include <cassert>
+
+namespace offloadtest {
+
+class CommandBuffer {
+  GPUAPI API;
+
+public:
+  explicit CommandBuffer(GPUAPI API) : API(API) {}
+  virtual ~CommandBuffer() = default;
+  CommandBuffer(const CommandBuffer &) = delete;
+  CommandBuffer &operator=(const CommandBuffer &) = delete;
+
+  GPUAPI getAPI() const { return API; }
+
+  template <typename T> T &as() {
+    assert(API == T::BackendAPI && "CommandBuffer backend mismatch");
+    return static_cast<T &>(*this);
+  }
+  template <typename T> const T &as() const {
+    assert(API == T::BackendAPI && "CommandBuffer backend mismatch");
+    return static_cast<const T &>(*this);
+  }
+};
+
+} // namespace offloadtest
+
+#endif // OFFLOADTEST_API_COMMANDBUFFER_H
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -16,8 +16,10 @@
 
 #include "API/API.h"
 #include "API/Capabilities.h"
+#include "API/CommandBuffer.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/ADT/iterator_range.h"
+#include "llvm/Support/Error.h"
 
 #include <memory>
 #include <string>
@@ -98,6 +100,9 @@ public:
   createBuffer(std::string Name, BufferCreateDesc &Desc,
                size_t SizeInBytes) = 0;
   virtual void printExtra(llvm::raw_ostream &OS) {}
+
+  virtual llvm::Expected<std::unique_ptr<CommandBuffer>>
+  createCommandBuffer() = 0;
 
   virtual ~Device() = 0;
 
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -389,6 +389,36 @@ public:
   }
 };
 
+class DXCommandBuffer : public offloadtest::CommandBuffer {
+public:
+  static constexpr GPUAPI BackendAPI = GPUAPI::DirectX;
+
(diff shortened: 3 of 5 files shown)


lib/API/DX/Device.cpp
```diff
@@ -309,6 +309,66 @@ class DXQueue : public offloadtest::Queue {
...
+  // management of each command buffer are independently safe without external
+  // synchronization.
+  ComPtr<ID3D12CommandAllocator> Allocator;
+  ComPtr<ID3D12GraphicsCommandList> CmdList;
+  ComPtr<ID3D12Fence> Fence;
+#ifdef _WIN32
+  HANDLE Event = nullptr;
+#else // WSL
+  int Event = -1;
+#endif
```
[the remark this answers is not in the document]
manon-traverse: #1007 has hit main ^.^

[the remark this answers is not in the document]
MarijnS95: It shouldn't even be here at all, the fence is already in the `InvocationState`. It should only start being signaled/returned in the `submit()` PR.

include/API/CommandBuffer.h
```diff
@@ -0,0 +1,44 @@
+//===- CommandBuffer.h - Offload Command Buffer API -----------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+//
+//===----------------------------------------------------------------------===//
```
bogner, line 10: This bit of header comment isn't really necessary if we don't have any description of the module inside of it.

include/API/CommandBuffer.h
```diff
@@ -0,0 +1,44 @@
...
+  GPUAPI getAPI() const { return API; }
+
+  template <typename T> T &as() {
+    assert(API == T::BackendAPI && "CommandBuffer backend mismatch");
+    return static_cast<T &>(*this);
+  }
+  template <typename T> const T &as() const {
+    assert(API == T::BackendAPI && "CommandBuffer backend mismatch");
+    return static_cast<const T &>(*this);
+  }
```
bogner, line 39: We should probably just set these classes up using LLVM-style RTTI rather than hand rolling it here: https://llvm.org/docs/HowToSetUpLLVMStyleRTTI.html

include/API/Device.h
```diff
@@ -99,6 +101,9 @@ class Device {
                size_t SizeInBytes) = 0;
   virtual void printExtra(llvm::raw_ostream &OS) {}
 
+  virtual llvm::Expected<std::unique_ptr<CommandBuffer>>
+  createCommandBuffer() = 0;
```
bogner, line 105: Too late to do anything about it, but note that the commit message claims:
> Device::createCommandBuffer() returns Expected<unique_ptr<CommandBuffer>> with a default "not supported" implementation.

However, this is a pure virtual function - there is no default implementation at all.

include/API/CommandBuffer.h
```diff
@@ -0,0 +1,44 @@
...
+#include <cassert>
+
+namespace offloadtest {
+
+class CommandBuffer {
+  GPUAPI API;
+
+public:
+  explicit CommandBuffer(GPUAPI API) : API(API) {}
+  virtual ~CommandBuffer() = default;
```
bogner, line 26: Classes with vtables should have an anchor to avoid object size and relocation bloat. See https://llvm.org/docs/CodingStandards.html#provide-a-virtual-method-anchor-for-classes-in-headers

lib/API/DX/Device.cpp
```diff
@@ -389,6 +389,36 @@ class DXQueue : public offloadtest::Queue {
   }
 };
 
+class DXCommandBuffer : public offloadtest::CommandBuffer {
+public:
+  static constexpr GPUAPI BackendAPI = GPUAPI::DirectX;
```
bogner, line 394: Seems like a false positive, but clang is emitting a warning for these:
```
E:/code/HLSLTest/lib/API/DX/Device.cpp(395,27): error: unused variable 'BackendAPI' [-Werror,-Wunused-const-variable]
  395 |   static constexpr GPUAPI BackendAPI = GPUAPI::DirectX;
      |                           ^~~~~~~~~~
```

Apparently we don't have -Werror for the pre-checkin CI jobs here, because it also showed up in those runs: https://github.com/llvm/offload-test-suite/actions/runs/24345003263/job/71083354829#step:11:4368

In any case, switching to LLVM-style RTTI will avoid the issue.

include/API/CommandBuffer.h
MarijnS95: Ah, I saw it in the other PRs as well and thought that was introduced by `clang-tidy` as a default. Will remove it or add a description in the next PR.

include/API/Device.h
MarijnS95: You're right - I fixed this last minute after realizing there were not actually any backends that leave this unimplemented, but forgot to update the commit message.

lib/API/DX/Device.cpp
MarijnS95: Correct, this was mostly to have backend-specific downcasting readily available when needed, for special-case tests as well as backend logic like `Queue` receiving generic `CommandBuffer` structures.

I'll look at LLVM-style RTTI as well as checking if the CI jobs can be extended with `-Werror` going forward as the build was always warning-free until this PR 😅

PR #1057, opened 2026-04-02
Add `Queue::submit()` for command buffer submission

MarijnS95:
~Depends on #1033~

Move command buffer submission logic from each backend's `Device` into `Queue::submit()`, which takes ownership of the command buffers. Each backend uses its `Fence` abstraction (#1007) for GPU synchronization:

- Metal: `commit()` + `waitUntilCompleted()`
- Vulkan: `vkQueueSubmit()` signaling a timeline semaphore (`VulkanFence`), then `VulkanFence::waitForCompletion()`
- DX12: `ExecuteCommandLists()` + `Queue::Signal()` on the queue-owned `DXFence`, then `DXFence::waitForCompletion()`

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -58,6 +58,19 @@ class Queue {
 public:
   virtual ~Queue() = 0;
 
+  /// Submit command buffers for execution and block until completion.
+  /// Command buffers execute in array order, but dependencies between them
+  /// require appropriate barriers within the command buffers themselves.
+  virtual llvm::Error
+  submit(llvm::SmallVectorImpl<std::unique_ptr<CommandBuffer>> &&CBs) = 0;
+
+  /// Convenience overload for submitting a single command buffer.
+  llvm::Error submit(std::unique_ptr<CommandBuffer> CB) {
+    llvm::SmallVector<std::unique_ptr<CommandBuffer>, 1> CBs;
+    CBs.push_back(std::move(CB));
+    return submit(std::move(CBs));
+  }
+
 protected:
   Queue() = default;
 };
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -433,22 +433,36 @@ public:
 
 class DXQueue : public offloadtest::Queue {
 public:
-  ComPtr<ID3D12CommandQueue> Queue;
+  using Queue::submit;
 
-  DXQueue(ComPtr<ID3D12CommandQueue> Queue) : Queue(Queue) {}
-  virtual ~DXQueue() {}
+  ComPtr<ID3D12CommandQueue> Queue;
+  std::unique_ptr<DXFence> SubmitFence;
+  uint64_t FenceCounter = 0;
+
+  DXQueue(ComPtr<ID3D12CommandQueue> Queue,
+          std::unique_ptr<DXFence> SubmitFence)
+      : Queue(Queue), SubmitFence(std::move(SubmitFence)) {}
+  DXQueue(DXQueue &&) = default;
+  ~DXQueue() override {}
 
   static llvm::Expected<DXQueue>
   createGraphicsQueue(ComPtr<ID3D12Device> Device) {
     const D3D12_COMMAND_QUEUE_DESC Desc = {D3D12_COMMAND_LIST_TYPE_DIRECT, 0,
                                            D3D12_COMMAND_QUEUE_FLAG_NONE, 0};
-    ComPtr<ID3D12CommandQueue> Queue;
-    if (auto Err =
-            HR::toError(Device->CreateCommandQueue(&Desc, IID_PPV_ARGS(&Queue)),
-                        "Failed to create command queue."))
+    ComPtr<ID3D12CommandQueue> CmdQueue;
+    if (auto Err = HR::toError(
+            Device->CreateCommandQueue(&Desc, IID_PPV_ARGS(&CmdQueue)),
+            "Failed to create command queue."))
       return Err;
-    return DXQueue(Queue);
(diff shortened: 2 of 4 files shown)


MarijnS95:
Looks like this needs some more TLC, from CI:

```
# | D3D: ID3D12CommandAllocator::LUCPrepareForDestruction: A command allocator 0x00000136B8B29B10:'Unnamed ID3D12CommandAllocator Object' is being reset before previous executions associated with the allocator have completed.
# | Exception Code: 0x0000087A
...
# | #11 0x00007ffa7089c1e4 (C:\WINDOWS\SYSTEM32\D3D12SDKLayers.dll+0x3c1e4)
# | #12 0x00007ff6c4440df9 `anonymous namespace'::DXCommandBuffer::~DXCommandBuffer E:\actions-runner\_work\offload-test-suite\offload-test-suite\OffloadTest\lib\API\DX\Device.cpp:364:0
```

MarijnS95:
This PR is in a bit of an odd interim state. It aims to at least generalize the submit+wait paths behind the common `CommandBuffer` abstraction, but there are still two points of contention:

1. Each backends' `InvocationState` still stores a "downcast" `CommandBuffer`. Storing the "trait" `offloadtest::CommandBuffer` requires extra casting in many places, or a casting helper that Claude generated here: https://github.com/Traverse-Research/offload-test-suite/commit/8ddb409f5ab4e2bbffa6ca4dcdd460b402646621.
2. This function is subject to change pending future plans. Right now it still waits for completion, but in the near future this function should default to submitting asynchronously and instead returning the relevant fence and "to-be-signalled value".
   That however also relies on a keepalive mechanism and less strictly defined resource ownership and destruction, something that was brought up for debate previously and likely requires some more thought and confirmation before committing to that pattern.

lib/API/DX/Device.cpp
```diff
@@ -444,6 +458,42 @@ class DXCommandBuffer : public offloadtest::CommandBuffer {
...
+
+  const uint64_t CurrentCounter = ++FenceCounter;
+  if (auto Err =
+          HR::toError(Queue->Signal(SubmitFence->Fence.Get(), CurrentCounter),
+                      "Failed to add signal."))
+    return Err;
+
+  // TODO: Return a Fence+value with keepalive lists instead of blocking here.
+  if (auto Err = SubmitFence->waitForCompletion(CurrentCounter))
+    return Err;
```
manon-traverse, line 493: Not a fan of blocking inside the submit function, but I think it's a fine step for now since that is the current behaviour. But let's look at this properly once more of the puzzle pieces are in place!

manon-traverse (approved):

lib/API/DX/Device.cpp
MarijnS95: That not only depends on simply returning the fence and to-be-signalled value as outlined above, but also a keepalive system that keeps the resources with the command buffer. Both are bigger refactors that are planned and done, but were explicitly excluded to keep PRs small and incremental.

Do I need to squash it into here after all?

manon-traverse: To clarify, I think going this route for this PR is totally fine! It was more of an 'I agree with this TODO, than a critique point'.
Once we have the keep-alive system and more of the command buffer abstraction in place we can improve it in a future PR.

include/API/Device.h
```diff
@@ -58,6 +58,19 @@ class Queue {
 public:
   virtual ~Queue() = 0;
 
+  /// Submit command buffers for execution and block until completion.
+  /// Command buffers execute in array order, but dependencies between them
+  /// require appropriate barriers within the command buffers themselves.
+  virtual llvm::Error
+  submit(llvm::SmallVector<std::unique_ptr<CommandBuffer>> CBs) = 0;
```
bogner: This could take a `llvm::SmallVectorImpl<std::unique_ptr<CommandBuffer>> &&`, which would allow it to be called regardless of the actual size of the caller's SmallVector.

include/API/Device.h
```diff
@@ -58,6 +58,19 @@ class Queue {
...
+
+  /// Convenience overload for submitting a single command buffer.
+  llvm::Error submit(std::unique_ptr<CommandBuffer> CB) {
+    llvm::SmallVector<std::unique_ptr<CommandBuffer>> CBs;
+    CBs.push_back(std::move(CB));
+    return submit(std::move(CBs));
```
bogner: If we do the above, then this SmallVector can have a fixed size of 1, rather than using a bunch of extra stack space here.

```suggestion
    return submit(
        llvm::SmallVector<std::unique_ptr<CommandBuffer>, 1>{std::move(CB)});
```

bogner (approved):

include/API/Device.h
MarijnS95: Sounds good to explicitly specify the size of `N=1`. However this move constructor doesn't seem to exist for `unique_ptr`:


     ...
     include/llvm/ADT/SmallVector.h:1254:11: note: in instantiation of member function
     'llvm::SmallVectorImpl<std::unique_ptr<offloadtest::CommandBuffer>>::append' requested here
      1254 |     this->append(IL);
           |           ^
     include/API/Device.h:70:9: note: in instantiation of member function
     'llvm::SmallVector<std::unique_ptr<offloadtest::CommandBuffer>, 1>::SmallVector' requested here
        70 |         llvm::SmallVector<std::unique_ptr<CommandBuffer>, 1>{std::move(CB)});
           |         ^
     include/c++/15.2.1/bits/unique_ptr.h:524:7: note: 'unique_ptr' has
     been explicitly marked deleted here
       524 |       unique_ptr(const unique_ptr&) = delete;
           |       ^

PR #1067, opened 2026-04-08
Add split build/test callable workflow for validation

### LINKED ISSUE #1081: Split build jobs from test jobs in the offload test suite.
There is an optimization opportunity for the offload test suite in testing different configs. 
By splitting out a job into a build job and a test job, and establishing an artifact transfer system, machines that are using the GPU for testing can also be utilized for building, and machines using the CPU for building can also take a separate testing job, so we can maximally utilize hardware.
Separately, we can take advantage of parallel builds, since all we truly need is for tests to be run on a specific GPU, but any machine with the right architecture can build compilers. We should change the workflow logic so that other vendor machines with say, AMD or Intel compilers, can build the compiler for say, an NVidia run. Since AMD and Nvidia aren't part of the default non-test-all label workflows, using those two machines will help lighten the load on Intel / QC.

bob80905:
Every PR test job holds a GPU runner for the full build+test duration (7–16 min), even though 67–86% of that time is CPU-only DXC/LLVM compilation. With one self-hosted runner per GPU SKU, this means builds block GPU-dependent test work unnecessarily.

This PR adds two new, standalone workflow files — no existing workflows are modified:

build-and-test-split-callable.yaml — A split version of the callable workflow:

 - Build job: Runs on any available runner of the matching architecture ([self-hosted, Windows, X64] for Intel/NVIDIA/AMD, [self-hosted, ARM64, Windows] for QC). Builds DXC + LLVM, uploads artifacts via tar.
 - Test job: Runs on the GPU-specific runner (hlsl-${{ inputs.SKU }}). Downloads build artifacts, extracts with --touch to preserve ninja timestamp freshness, then runs tests.
 - macOS: Unchanged — stays as a single combined job since builds are only ~1 min.

validate-split-build-test.yaml — A workflow_dispatch workflow to validate the split callable per-SKU before migrating production workflows.

Fixes https://github.com/llvm/offload-test-suite/issues/1081

Assisted by: Github Copilot

### DIFF
--- a/.github/workflows/build-and-test-split-callable.yaml
+++ b/.github/workflows/build-and-test-split-callable.yaml
@@ -220,7 +220,7 @@ jobs:
               DXC/build/bin llvm-project/build
       - name: Upload build artifacts
         if: inputs.OS == 'windows'
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: build-${{ inputs.SKU }}-${{ inputs.TestTarget }}
           path: ${{ runner.temp }}/build-artifacts.tar
@@ -269,7 +269,7 @@ jobs:
       # --touch ensures extracted files get current timestamps so ninja
       # sees build outputs as newer than source files (no rebuild).
       - name: Download build artifacts
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
           name: build-${{ inputs.SKU }}-${{ inputs.TestTarget }}
           path: ${{ runner.temp }}
@@ -315,7 +315,7 @@ jobs:
           Write-Host "DxDiag report saved to $output"
       - name: Upload dxdiag artifact
         if: failure()
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: dxdiag-${{ inputs.SKU }}-${{inputs.TestTarget}}.txt
           path: ${{ runner.temp }}/dxdiag.txt
--- a/.github/workflows/build-and-test-split-callable.yaml
+++ b/.github/workflows/build-and-test-split-callable.yaml
@@ -1,0 +1,326 @@
+name: HLSL Test (Split Build/Test)
+
+permissions:
+  contents: read
+
+on:
+  workflow_dispatch:
+    inputs:
+      OffloadTest-branch:
+        description: 'Test Suite Branch'
+        required: false
+        default: 'main'
+        type: string
+      LLVM-branch:
+        description: 'LLVM Branch'
+        required: false
+        default: 'main'
+        type: string
+      LLVM-fork:
+        description: 'LLVM fork'
+        required: false
+        default: 'llvm'
+        type: string
+      DXC-branch:
+        description: 'DXC Branch'
+        required: false
+        default: 'main'
+        type: string
+      BuildType:
+        description: 'Build Type'
+        required: false
+        default: 'RelWithDebInfo'
+        type: choice
+        options:
+        - Release
+        - RelWithDebInfo
+        - Debug
+      TestTarget:
+        required: false
+        default: 'check-hlsl'
+        type: string
+      OS:
+        required: true
+        type: choice
+        options:
+        - macOS
+        - windows
+      SKU:
+        required: true
+        type: choice
+        options:
+        - macos
+        - windows-amd
+        - windows-intel
+        - windows-nvidia
+        - windows-qc
+      LLVM-ExtraCMakeArgs:
+        description: 'Extra CMake Args for LLVM'
+        required: false
+        default: ''
+        type: string
+  workflow_call:
+    inputs:
+      OffloadTest-branch:
+        description: 'Test Suite Branch'
+        required: false
+        default: 'main'
+        type: string
+      LLVM-branch:
+        description: 'LLVM Branch'
+        required: false
+        default: 'main'
+        type: string
+      LLVM-fork:
+        description: 'LLVM For'
+        required: false
+        default: 'llvm'
+        type: string
+      DXC-branch:
+        description: 'DXC Branch'
+        required: false
+        default: 'main'
+        type: string
+      OS:
+        required: true
+        type: string
+      SKU:
+        required: true
+        type: string
+      BuildType:
+        description: 'Build Config'
+        required: false
+        default: 'RelWithDebInfo'
+        type: string
+      TestTarget:
+        required: false
+        default: 'check-hlsl'
+        type: string
+      LLVM-ExtraCMakeArgs:
+        description: 'Extra CMake Args for LLVM'
+        required: false
+        default: ''
+        type: string
+
+# ============================================================================
+# Split Build/Test Architecture:
+#
+# Windows: Build job runs on ANY available runner of the right architecture
+#   (load-balanced via generic labels), then test job runs on the specific
+#   GPU runner. This frees GPU runners during the build phase.
+#
+# macOS: Single combined job (builds are fast ~1 min, split not worth it).
+# ============================================================================
+
+jobs:
+  build:
+    permissions:
+      checks: write
+    # macOS: use the specific macOS runner (single combined job)
+    # Windows QC (ARM64): use any ARM64 Windows runner
+    # Windows x64: use any x64 Windows runner (load-balanced across Intel/NVIDIA/AMD)
+    runs-on: ${{ fromJSON(inputs.OS == 'macOS' && '["self-hosted", "hlsl-macos"]' || inputs.SKU == 'windows-qc' && '["self-hosted", "ARM64", "Windows"]' || '["self-hosted", "Windows", "X64"]') }}
+    steps:
+      - name: Checkout DXC
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          repository: Microsoft/DirectXShaderCompiler
+          ref: ${{ inputs.DXC-branch }}
+          path: DXC
+          fetch-depth: 1
+          submodules: true
+      - name: Checkout LLVM
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          repository: ${{ inputs.LLVM-fork }}/llvm-project
+          ref: ${{ inputs.LLVM-branch }}
+          path: llvm-project
+          fetch-depth: 1
+      - name: Checkout OffloadTest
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test-split-callable.yaml
```diff
@@ -0,0 +1,327 @@
...
+            tar cf $RUNNER_TEMP/build-artifacts.tar \
+              --exclude='*.obj' \
+              --exclude='*.o' \
+              --exclude='*.ilk' \
+              --exclude='*.pdb' \
+              --exclude='CMakeFiles' \
+              DXC/build/bin llvm-project/build
+      - name: Upload build artifacts
+        if: inputs.OS == 'windows'
+        uses: actions/upload-artifact@v4
```
Icohedron:
```suggestion
        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
```

.github/workflows/build-and-test-split-callable.yaml
```diff
@@ -0,0 +1,327 @@
...
+        if: failure()
+        shell: powershell
+        run: |
+          $fileName = "dxdiag.txt"
+          $output = Join-Path $env:RUNNER_TEMP $fileName
+          dxdiag /t $output | Out-Null
+          Write-Host "DxDiag report saved to $output"
+      - name: Upload dxdiag artifact
+        if: failure()
+        uses: actions/upload-artifact@v4
```
Icohedron:
```suggestion
        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
```

.github/workflows/build-and-test-split-callable.yaml
```diff
@@ -0,0 +1,327 @@
...
+          repository: llvm/offload-golden-images
+          ref: main
+          path: golden-images
+          fetch-depth: 1
+
+      # Download and extract build artifacts over the source checkouts.
+      # --touch ensures extracted files get current timestamps so ninja
+      # sees build outputs as newer than source files (no rebuild).
+      - name: Download build artifacts
+        uses: actions/download-artifact@v4
```
Icohedron:
```suggestion
        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
```

tex3d (approved):

Icohedron (approved):

.github/workflows/build-and-test-split-callable.yaml
bogner: I find this naming a little bit awkward - the "split" here is only with reference to the version of this that isn't "split". Can we come up with a clearer name for what this is specifically for?

.github/workflows/build-and-test-split-callable.yaml
```diff
@@ -0,0 +1,327 @@
...
+        - windows
+      SKU:
+        required: true
+        type: choice
+        options:
+        - macos
+        - windows-amd
+        - windows-intel
+        - windows-nvidia
+        - windows-qc
```
bogner, line 56: Seems like this is missing in the `build-and-test-callable` you based this on - is that just a bug?

.github/workflows/build-and-test-split-callable.yaml
```diff
@@ -0,0 +1,327 @@
...
+
+# ============================================================================
+# Split Build/Test Architecture:
+#
+# Windows: Build job runs on ANY available runner of the right architecture
+#   (load-balanced via generic labels), then test job runs on the specific
+#   GPU runner. This frees GPU runners during the build phase.
+#
+# macOS: Single combined job (builds are fast ~1 min, split not worth it).
+# ============================================================================
```
bogner, line 113: This comment looks like AI cruft (why explain this randomly in the middle of the file, nowhere near the logic that does it?).

In any case, I really don't like that we're deciding whether to split the build or not based on "is it macOS?" - this should really be the responsibility of the job that uses this to decide whether its split or not. The "builds are fast" argument could easily apply to some arbitrary machine we have in the future that isn't macOS. Can we parametrize this so that the calling job can decide whether or not to split it?

.github/workflows/build-and-test-split-callable.yaml
```diff
@@ -0,0 +1,327 @@
...
+
+jobs:
+  build:
+    permissions:
+      checks: write
+    # macOS: use the specific macOS runner (single combined job)
+    # Windows QC (ARM64): use any ARM64 Windows runner
+    # Windows x64: use any x64 Windows runner (load-balanced across Intel/NVIDIA/AMD)
+    runs-on: ${{ fromJSON(inputs.OS == 'macOS' && '["self-hosted", "hlsl-macos"]' || inputs.SKU == 'windows-qc' && '["self-hosted", "ARM64", "Windows"]' || '["self-hosted", "Windows", "X64"]') }}
```
bogner, line 122: This is probably okay for now, but AFAICT we *could* technically have non-QC GPUs on arm64 (or vice versa). We just don't happen to.

.github/workflows/build-and-test-split-callable.yaml
bogner, line 1: The "split" naming, both here and in the job name, really is only there to differentiate from the existing `build-and-test-callable`. Should I assume we'll rename this once we retire that config?

Notably, it's a bit confusing to call this "split" when it also has logic to not be split at all (currently limited to macOS).

.github/workflows/build-and-test-split-callable.yaml
```diff
@@ -0,0 +1,327 @@
...
+        run: |
+            cd llvm-project
+            mkdir build
+            cd build
+            cmake -G Ninja ${{ inputs.LLVM-ExtraCMakeArgs }} -DCMAKE_BUILD_TYPE=${{ inputs.BuildType }} -DLLVM_ENABLE_ASSERTIONS=On -C ${{ github.workspace }}/llvm-project/clang/cmake/caches/HLSL.cmake -C ${{ github.workspace }}/OffloadTest/cmake/caches/sccache.cmake -DDXC_DIR=${{ github.workspace }}/DXC/build/bin -DLLVM_EXTERNAL_OFFLOADTEST_SOURCE_DIR=${{ github.workspace }}/OffloadTest -DLLVM_EXTERNAL_PROJECTS="OffloadTest" -DLLVM_LIT_ARGS="--xunit-xml-output=testresults.xunit.xml -v" -DOFFLOADTEST_TEST_CLANG=${{steps.Test-Clang.outputs.TEST_CLANG || 'Off' }} -DGOLDENIMAGE_DIR=${{ github.workspace }}/golden-images ${{ github.workspace }}/llvm-project/llvm/
+            ninja hlsl-test-depends
+
+      # === macOS: run tests in this job (no split) ===
+      - name: Dump GPU Info
+        if: inputs.OS == 'macOS'
```
bogner, line 189: As noted above, I think differentiating based off of macOS/windows here and throughout is a mistake. Also what should we do if we were to set up a runner on linux?

.github/workflows/build-and-test-split-callable.yaml
```diff
@@ -0,0 +1,327 @@
...
+        shell: bash
+        run: |
+            cd $GITHUB_WORKSPACE
+            tar cf $RUNNER_TEMP/build-artifacts.tar \
+              --exclude='*.obj' \
+              --exclude='*.o' \
+              --exclude='*.ilk' \
+              --exclude='*.pdb' \
+              --exclude='CMakeFiles' \
+              DXC/build/bin llvm-project/build
```
bogner, line 220: We should think about how we want to structure this command in a cross platform way.

.github/workflows/build-and-test-split-callable.yaml
```diff
@@ -0,0 +1,327 @@
...
+      - name: Upload build artifacts
+        if: inputs.OS == 'windows'
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: build-${{ inputs.SKU }}-${{ inputs.TestTarget }}
+          path: ${{ runner.temp }}/build-artifacts.tar
+          retention-days: 1
+
+  test:
+    if: inputs.OS != 'macOS'
```
bogner, line 230: Why is this one `inputs.OS != 'macOS'` when the packaging step is `inputs.OS == 'windows'`?

.github/workflows/build-and-test-split-callable.yaml
```diff
@@ -0,0 +1,327 @@
...
+          ref: ${{ inputs.OffloadTest-branch }}
+          path: OffloadTest
+          fetch-depth: 1
+      - name: Checkout Golden Images
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          repository: llvm/offload-golden-images
+          ref: main
+          path: golden-images
+          fetch-depth: 1
```
bogner, line 152: We shouldn't need this for the build step

.github/workflows/build-and-test-split-callable.yaml
```diff
@@ -0,0 +1,327 @@
...
+            cd build
+            ./bin/api-query
+      - name: Run HLSL Tests
+        run: |
+            cd llvm-project
+            cd build
+            ninja check-hlsl-unit
+            ninja ${{ inputs.TestTarget }}
+      - name: Run dxdiag (Windows only)
+        if: failure()
```
bogner, line 309: This should be `if: inputs.OS == 'windows' && failure()`

.github/workflows/build-and-test-split-callable.yaml
```diff
@@ -0,0 +1,327 @@
...
+        shell: powershell
+        run: |
+          $fileName = "dxdiag.txt"
+          $output = Join-Path $env:RUNNER_TEMP $fileName
+          dxdiag /t $output | Out-Null
+          Write-Host "DxDiag report saved to $output"
+      - name: Upload dxdiag artifact
+        if: failure()
```
bogner, line 317: Same here

.github/workflows/build-and-test-split-callable.yaml
```diff
@@ -0,0 +1,327 @@
...
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: dxdiag-${{ inputs.SKU }}-${{inputs.TestTarget}}.txt
+          path: ${{ runner.temp }}/dxdiag.txt
+      #- name: Publish Test Results
+      #  uses: EnricoMi/publish-unit-test-result-action/windows@34d7c956a59aed1bfebf31df77b8de55db9bbaaf # v2.21.0
+      #  if: always() && inputs.OS == 'windows'
+      #  with:
+      #    comment_mode: off
+      #    files: llvm-project/build/**/testresults.xunit.xml
```
bogner, line 327: We shouldn't be adding commented out code like this (even if we're copying it from a place that's already wrong...)

bogner (commented):
I think we should probably try to parametrize the existing build-and-test-callable to optionally be split, rather than duplicating it *and* making it work either way. Several of my comments here lead to that conclusion, though there are a couple of other little things I spotted as well.

PR #1070, opened 2026-04-09
Abstract Pipeline creation (including pipeline layouts and root signatures)

manon-traverse:
Add an API-agnostic way of creating compute and graphics pipelines. This includes the creation of pipeline layouts and root signatures. In Metal, there is no real equivalent due to the use of a top-level argument buffer, so the BindingsDesc variable is ignored.

### DIFF
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -46,16 +46,16 @@ struct DeviceConfig {
   bool EnableValidationLayer = false;
 };
 
-struct DXBinding {
-  uint32_t Register;
-  uint32_t Space;
-};
-
 struct InputLayoutDesc {
   std::string Name;
   Format Format;
   uint32_t OffsetInBytes;
   std::optional<uint32_t> InstanceStepRate;
+};
+
+struct DXBinding {
+  uint32_t Register;
+  uint32_t Space;
 };
 
 struct ResourceBindingDesc {
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -46,7 +46,7 @@ struct DeviceConfig {
   bool EnableValidationLayer = false;
 };
 
-struct DXBinding { // TODO(manon): rename
+struct DXBinding {
   uint32_t Register;
   uint32_t Space;
 };
@@ -58,7 +58,6 @@ struct InputLayoutDesc {
   std::optional<uint32_t> InstanceStepRate;
 };
 
-// TODO(manon): Come up with a good name
 struct ResourceBindingDesc {
   ResourceKind Kind;
   DXBinding DXBinding;
@@ -156,11 +155,11 @@ public:
 
   virtual Queue &getGraphicsQueue() = 0;
 
-  virtual llvm::Expected<std::shared_ptr<PipelineState>>
+  virtual llvm::Expected<std::unique_ptr<PipelineState>>
   createPipelineCs(llvm::StringRef Name, const BindingsDesc &BindingsDesc,
                    ShaderContainer CS) = 0;
 
-  virtual llvm::Expected<std::shared_ptr<PipelineState>>
+  virtual llvm::Expected<std::unique_ptr<PipelineState>>
   createPipelineVsPs(llvm::StringRef Name, const BindingsDesc &BindingsDesc,
                      llvm::ArrayRef<InputLayoutDesc> InputLayout,
                      llvm::ArrayRef<Format> RTFormats,
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -48,7 +48,7 @@ struct DeviceConfig {
 
 struct InputLayoutDesc {
   std::string Name;
-  Format Format;
+  Format Fmt;
   uint32_t OffsetInBytes;
   std::optional<uint32_t> InstanceStepRate;
 };
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -19,12 +19,18 @@
 #include "API/Capabilities.h"
 #include "API/CommandBuffer.h"
 #include "API/Texture.h"
+
 #include "Support/Pipeline.h"
+
+#include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/ADT/iterator_range.h"
 #include "llvm/Support/Error.h"
+#include "llvm/Support/MemoryBuffer.h"
 
+#include <cstdint>
 #include <memory>
+#include <optional>
 #include <string>
 
 namespace llvm {
@@ -38,6 +44,57 @@ struct Pipeline;
 struct DeviceConfig {
   bool EnableDebugLayer = false;
   bool EnableValidationLayer = false;
+};
+
+struct DXBinding { // TODO(manon): rename
+  uint32_t Register;
+  uint32_t Space;
+};
+
+struct InputLayoutDesc {
+  std::string Name;
+  Format Format;
+  uint32_t OffsetInBytes;
+  std::optional<uint32_t> InstanceStepRate;
+};
+
+// TODO(manon): Come up with a good name
+struct ResourceBindingDesc {
+  ResourceKind Kind;
+  DXBinding DXBinding;
+  std::optional<VulkanBinding> VKBinding;
+  uint32_t DescriptorCount;
+};
+
+struct DescriptorSetLayoutDesc {
+  llvm::SmallVector<ResourceBindingDesc> ResourceBindings;
+};
+
+struct PushConstantsRange {
+  uint32_t OffsetInBytes; // Must be multiple of 4
+  uint32_t SizeInBytes;   // Must be multiple of 4
+};
+
+struct BindingsDesc {
+  llvm::SmallVector<DescriptorSetLayoutDesc> DescriptorSetDescs;
+  llvm::SmallVector<PushConstantsRange> PushConstantRanges;
+};
+
+struct ShaderContainer {
+  std::string EntryPoint;
+  const llvm::MemoryBuffer *Shader;
+  llvm::SmallVector<SpecializationConstant> SpecializationConstants;
+};
+
+class PipelineState {
+public:
+  virtual ~PipelineState() = default;
+
+  PipelineState(const Buffer &) = delete;
+  PipelineState &operator=(const Buffer &) = delete;
+
+protected:
+  PipelineState() = default;
 };
 
 class Fence {
@@ -98,6 +155,17 @@ public:
   virtual llvm::Error executeProgram(Pipeline &P) = 0;
 
   virtual Queue &getGraphicsQueue() = 0;
+
+  virtual llvm::Expected<std::shared_ptr<PipelineState>>
+  createPipelineCs(llvm::StringRef Name, const BindingsDesc &BindingsDesc,
+                   ShaderContainer CS) = 0;
+
+  virtual llvm::Expected<std::shared_ptr<PipelineState>>
+  createPipelineVsPs(llvm::StringRef Name, const BindingsDesc &BindingsDesc,
+                     llvm::ArrayRef<InputLayoutDesc> InputLayout,
+                     llvm::ArrayRef<Format> RTFormats,
+                     std::optional<Format> DSFormat, ShaderContainer VS,
+                     ShaderContainer PS) = 0;
 
   virtual llvm::Expected<std::unique_ptr<Fence>>
   createFence(llvm::StringRef Name) = 0;
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -87,13 +87,17 @@ struct ShaderContainer {
 
 class PipelineState {
 public:
+  GPUAPI API;
+
   virtual ~PipelineState() = default;
 
   PipelineState(const Buffer &) = delete;
   PipelineState &operator=(const Buffer &) = delete;
 
+  GPUAPI getAPI() const { return API; }
+
 protected:
-  PipelineState() = default;
+  explicit PipelineState(GPUAPI API) : API(API) {}
 };
 
 class Fence {
--- a/include/API/Enums.h
+++ b/include/API/Enums.h
@@ -1,0 +1,29 @@
+
+#ifndef OFFLOADTEST_API_ENUMS_H
+#define OFFLOADTEST_API_ENUMS_H
+
+namespace offloadtest {
+
+enum class ResourceKind {
+  Buffer,
+  StructuredBuffer,
+  ByteAddressBuffer,
+  Texture2D,
+  RWBuffer,
+  RWStructuredBuffer,
+  RWByteAddressBuffer,
+  RWTexture2D,
+  ConstantBuffer,
+  Sampler,
+  SampledTexture2D,
+};
+
+enum ShaderContainerType {
+  DXIL,
+  SPIRV,
+  Metal,
+};
+
+} // namespace offloadtest
+
+#endif // OFFLOADTEST_API_ENUMS_H
--- a/include/API/Enums.h
+++ b/include/API/Enums.h
@@ -1,3 +1,10 @@
+//===- Enums.h - Offload API Device API -----------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
 
 #ifndef OFFLOADTEST_API_ENUMS_H
 #define OFFLOADTEST_API_ENUMS_H
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -796,16 +796,16 @@ public:
     std::vector<D3D12_INPUT_ELEMENT_DESC> DXInputLayout;
     DXInputLayout.reserve(InputLayout.size());
     for (const InputLayoutDesc &Elem : InputLayout) {
+      assert(!Elem.InstanceStepRate &&
+             "Instance step rate is currently not supported.");
+
       D3D12_INPUT_ELEMENT_DESC ElementDesc = {};
       ElementDesc.SemanticName = Elem.Name.c_str();
       ElementDesc.SemanticIndex = 0;
       ElementDesc.Format = getDXGIFormat(Elem.Format);
       ElementDesc.InputSlot = 0;
       ElementDesc.AlignedByteOffset = Elem.OffsetInBytes;
-      ElementDesc.InputSlotClass =
-          Elem.InstanceStepRate ? D3D12_INPUT_CLASSIFICATION_PER_INSTANCE_DATA
-                                : D3D12_INPUT_CLASSIFICATION_PER_VERTEX_DATA;
-      ElementDesc.InstanceDataStepRate = Elem.InstanceStepRate.value_or(0);
+      ElementDesc.InputSlotClass = D3D12_INPUT_CLASSIFICATION_PER_VERTEX_DATA;
       DXInputLayout.push_back(ElementDesc);
     }
 
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -598,8 +598,7 @@ private:
   struct InvocationState {
     ComPtr<ID3D12DescriptorHeap> DescHeap;
     std::unique_ptr<DXCommandBuffer> CB;
-
-    std::shared_ptr<PipelineState> Pipeline;
+    std::unique_ptr<PipelineState> Pipeline;
 
     // Resources for graphics pipelines.
     std::unique_ptr<offloadtest::Texture> RT;
@@ -753,7 +752,7 @@ public:
                                                OutRootSignature);
   }
 
-  llvm::Expected<std::shared_ptr<PipelineState>>
+  llvm::Expected<std::unique_ptr<PipelineState>>
   createPipelineCs(llvm::StringRef Name, const BindingsDesc &BindingsDesc,
                    ShaderContainer CS) override {
     ComPtr<ID3D12RootSignature> RootSig;
@@ -778,10 +777,10 @@ public:
             "Failed to create PSO."))
       return Err;
 
-    return std::make_shared<DXPipelineState>(Name, RootSig, PSO);
+    return std::make_unique<DXPipelineState>(Name, RootSig, PSO);
   }
 
-  llvm::Expected<std::shared_ptr<PipelineState>>
+  llvm::Expected<std::unique_ptr<PipelineState>>
   createPipelineVsPs(llvm::StringRef Name, const BindingsDesc &BindingsDesc,
                      llvm::ArrayRef<InputLayoutDesc> InputLayout,
                      llvm::ArrayRef<Format> RTFormats,
@@ -839,7 +838,7 @@ public:
             "Failed to create graphics PSO."))
       return Err;
 
-    return std::make_shared<DXPipelineState>(Name, RootSig, PSO);
+    return std::make_unique<DXPipelineState>(Name, RootSig, PSO);
   }
 
   llvm::Expected<std::unique_ptr<offloadtest::Fence>>
@@ -2023,7 +2022,7 @@ public:
           createPipelineCs("Compute Pipeline State", BindingsDesc, CS);
       if (!PipelineStateOrErr)
         return PipelineStateOrErr.takeError();
-      State.Pipeline = *PipelineStateOrErr;
+      State.Pipeline = std::move(*PipelineStateOrErr);
       llvm::outs() << "Compute Pipeline created.\n";
       if (auto Err = createComputeCommands(P, State))
         return Err;
@@ -2083,7 +2082,7 @@ public:
           Format::D32FloatS8Uint, VS, PS);
       if (!PipelineStateOrErr)
         return PipelineStateOrErr.takeError();
-      State.Pipeline = *PipelineStateOrErr;
+      State.Pipeline = std::move(*PipelineStateOrErr);
       llvm::outs() << "Graphics Pipeline created.\n";
       if (auto Err = createGraphicsCommands(P, State))
         return Err;
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -33,6 +33,7 @@
 
 #include "API/Capabilities.h"
 #include "API/Device.h"
+#include "API/FormatConversion.h"
 #include "DXFeatures.h"
 #include "Support/Pipeline.h"
 #include "Support/WinError.h"
@@ -357,6 +358,17 @@ public:
   }
 };
 
+class DXPipelineState : public offloadtest::PipelineState {
+public:
+  std::string Name;
+  ComPtr<ID3D12RootSignature> RootSig;
+  ComPtr<ID3D12PipelineState> PSO;
+
+  DXPipelineState(llvm::StringRef Name, ComPtr<ID3D12RootSignature> RootSig,
+                  ComPtr<ID3D12PipelineState> PSO)
+      : Name(Name), RootSig(RootSig), PSO(PSO) {}
+};
+
 class DXFence : public offloadtest::Fence {
 public:
 #ifdef _WIN32
@@ -584,10 +596,10 @@ private:
   };
 
   struct InvocationState {
-    ComPtr<ID3D12RootSignature> RootSig;
     ComPtr<ID3D12DescriptorHeap> DescHeap;
-    ComPtr<ID3D12PipelineState> PSO;
     std::unique_ptr<DXCommandBuffer> CB;
+
+    std::shared_ptr<PipelineState> Pipeline;
 
     // Resources for graphics pipelines.
     std::unique_ptr<offloadtest::Texture> RT;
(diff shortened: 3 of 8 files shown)


include/API/Enums.h
```diff
@@ -0,0 +1,64 @@
...
+
+enum ShaderContainerType {
+  DXIL,
+  SPIRV,
+  Metal,
+};
+
+enum class Format {
+  R8Sint,
+  R8g8Sint,
```
manon-traverse: https://github.com/llvm/offload-test-suite/pull/1020 already adds a `Format` enum. We should use that one once that PR hits main. For this reason, I have marked this PR as a draft.

include/API/Device.h
```diff
@@ -56,6 +62,56 @@ class Buffer {
...
+struct DXBinding {
+  uint32_t Register;
+  uint32_t Space;
+};
+
+struct InputLayoutDesc {
+  std::string Name;
+  Format Format;
+  uint32_t OffsetInBytes;
+  std::optional<uint32_t> InstanceStepRate;
```
manon-traverse, line 58: This misses some parameters that are not used. Those will be added once we need them.

include/API/Enums.h
```diff
@@ -0,0 +1,29 @@
+
```
bogner, line 8: Needs the standard header comment

lib/API/MTL/MTLDevice.cpp
```diff
@@ -798,17 +793,82 @@ class MTLDevice : public offloadtest::Device {
...
+      auto FormatOrErr = toFormat(P.Bindings.RTargetBufferPtr->Format,
+                                  P.Bindings.RTargetBufferPtr->Channels);
+      if (!FormatOrErr)
+        return FormatOrErr.takeError();
+
+      llvm::SmallVector<Format> RTFormats;
+      RTFormats.push_back(*FormatOrErr);
+
+      auto PipelineStateOrErr = createPipelineVsPs(
+          "Graphics Pipeline State", Bindings, InputLayout, RTFormats, VS, PS);
```
MarijnS95: Forgot the optional DS format here causing a CI failure. To  match the other backends:
```suggestion
          "Graphics Pipeline State", Bindings, InputLayout, RTFormats,
          Format::D32FloatS8Uint, VS, PS);
```

include/API/Device.h
MarijnS95: Nit: can you sort this struct away from being **in between** the structs that are used (transitively) in `DescriptorSetLayoutDesc`?

lib/API/VK/Device.cpp
```diff
@@ -1445,81 +1846,22 @@ class VulkanDevice : public offloadtest::Device {
...
+    const VulkanPipelineState *VulkanPipeline =
+        static_cast<VulkanPipelineState *>(IS.Pipeline.get());
+
     VkDescriptorSetAllocateInfo DSAllocInfo = {};
     DSAllocInfo.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_ALLOCATE_INFO;
     DSAllocInfo.descriptorPool = IS.Pool;
-    DSAllocInfo.descriptorSetCount = IS.DescriptorSetLayouts.size();
-    DSAllocInfo.pSetLayouts = IS.DescriptorSetLayouts.data();
+    DSAllocInfo.descriptorSetCount = VulkanPipeline->SetLayouts.size(); // error
+    DSAllocInfo.pSetLayouts = VulkanPipeline->SetLayouts.data();
```
MarijnS95: `// error`?

lib/API/MTL/MTLDevice.cpp
```diff
@@ -787,6 +636,152 @@ class MTLDevice : public offloadtest::Device {
     return MTLCommandBuffer::create(GraphicsQueue.Queue);
   }
 
+  llvm::Expected<std::unique_ptr<PipelineState>>
+  createPipelineCs(llvm::StringRef Name,
+                   const BindingsDesc & /*unused on metal*/,
```
MarijnS95, line 673: Metal does store this info in reflection, and it does even have a typed builder. But since Metal IRConverter converts this into an arbitrary "buffer of bytes" (top-level argument buffer) that we can just bind, no type info is needed 🎉

lib/API/VK/Device.cpp
manon-traverse: I don't know why that comment is there... I must have wrote it, I don't remember doing so. The code works fine so I"ll just remove that comment.

MarijnS95:
Since this PR mostly seems to move (lots of) code around, I asked Claude to boil down all the **semantic** changes in the diff as these are omitted from the PR description. Let me know if the below is correct and matches the intent, or if it found nonsense, or if certain things weren't intended to be changed like that. It'll help with the review of close to 1200 lines :)

## Semantic changes in moved code

Beyond the refactoring (splitting functions, RAII, new abstract types), these are the behavioral differences I spotted between the old and new code:

### DX/Device.cpp

1. **Depth/stencil disabled by default** in `createPipelineVsPs`: old had `DepthEnable=true`, `DepthWriteMask=ALL`, `DepthFunc=LESS`. New has `DepthEnable=false`, `StencilEnable=false`. The caller now passes `DSFormat` to control this, but the PSO itself no longer enables depth testing.

2. **Input layout supports instancing**: new code checks `InstanceStepRate` to pick `PER_INSTANCE_DATA` vs `PER_VERTEX_DATA`. Old hardcoded `PER_VERTEX_DATA`.

3. **Multiple render targets**: old hardcoded `NumRenderTargets=1`. New loops over `RTFormats`.

### MTL/MTLDevice.cpp

4. **Vertex attribute mapping rewritten**: old reflected on the Metal `Function`'s `vertexAttributes()` and matched by lowercased name + "0" suffix. New ignores shader reflection and assigns attributes sequentially by index (0, 1, 2, ...). Semantics change if shader attribute indices don't match sequential order.

5. **Depth/stencil now conditional** on `DSFormat.has_value()` and stencil only set for `D32FloatS8Uint`. Old unconditionally set both.

6. **Multiple render targets** supported via loop, vs old single-RT.

### VK/Device.cpp

7. **Pipeline cache removed**: old created a `VkPipelineCache`, new passes `VK_NULL_HANDLE`.

8. **Vulkan 1.3 dynamic rendering**: old used `renderPass = IS.RenderPass`. New uses `VkPipelineRenderingCreateInfo` via `pNext` with `renderPass = VK_NULL_HANDLE`. Significant architecture change.

9. **Per-shader specialization constants for graphics**: old only handled spec constants for compute. New builds separate `VkSpecializationInfo` for VS and PS independently.

10. **Blend attachment count** now matches `RTFormats.size()` instead of hardcoded 1.

11. **Better error cleanup** in layout creation: new code destroys already-created set layouts if a later one fails. Old didn't.

lib/API/MTL/MTLDevice.cpp
```diff
@@ -217,178 +220,22 @@ class MTLDevice : public offloadtest::Device {
...
-            "Compute pipeline must have exactly one compute shader.");
-      const llvm::StringRef Program = P.Shaders[0].Shader->getBuffer();
-      dispatch_data_t Data = dispatch_data_create(
-          Program.data(), Program.size(), dispatch_get_main_queue(),
-          ^{
-          });
-      MTL::Library *Lib = Device->newLibrary(Data, &Error);
-      if (Error)
-        return toError(Error);
-      IS.Pool->addObject(Lib);
```
MarijnS95, line 337: Fwiw I'm not entirely confident on this anymore, but we've seen in our own application(s) that not adding objects to autoreleasepools has the adverse effect of them being leaked even when `release()` is called (because of unclear/inconsistent ownership transfer semantics on return), at least on non-main threads where an autoreleasepool isn't "attached" by default (or passed down?).

Probably not very relevant for these run-once executables, but something to keep in mind when building long-running graphics apps.

MarijnS95: Fwiw, the sub-objects allocated inside `createPipelineVsPs` — `VtxDesc` (`VertexDescriptor`), `AttrDesc` (`VertexAttributeDescriptor`), `LDesc` (`VertexBufferLayoutDescriptor`), and `RPCA` (`RenderPipelineColorAttachmentDescriptor`) — are all `alloc()->init()`'d but never `release()`'d and not in any autorelease pool. Same was true in the old code though (they weren't added to `IS.Pool` there either), so it's not a regression — just a pre-existing leak that carried over.

manon-traverse: Thanks for noticing this! I'll pick this up as I think it's good that we do fix these kind of issues as we are finding them.

lib/API/MTL/MTLDevice.cpp
```diff
@@ -798,17 +793,82 @@ class MTLDevice : public offloadtest::Device {
...
+        ResourceBindingDesc ResourceBinding = {};
+        ResourceBinding.Kind = R.Kind;
+        ResourceBinding.DXBinding.Register = R.DXBinding.Register;
+        ResourceBinding.DXBinding.Space = R.DXBinding.Space;
+        ResourceBinding.VKBinding = R.VKBinding;
+        ResourceBinding.DescriptorCount = R.getArraySize();
+        Layout.ResourceBindings.push_back(ResourceBinding);
+      }
+      Bindings.DescriptorSetDescs.push_back(Layout);
+    }
```
MarijnS95, line 871: Perhaps not for this PR, but would it make sense if the **already backend-agnostic `struct Resource`** stored this **new** `struct ResourceBindingDesc`, rather than duplicating this conversion across all backends?

lib/API/MTL/MTLDevice.cpp
```diff
@@ -787,6 +636,152 @@ class MTLDevice : public offloadtest::Device {
...
+          MTL::RenderPipelineColorAttachmentDescriptor::alloc()->init();
+      RPCA->setPixelFormat(getMetalPixelFormat(RTFormats[I]));
+      Desc->colorAttachments()->setObject(RPCA, I);
+    }
+
+    // Configure depth/stencil attachment.
+    if (DSFormat) {
+      const MTL::PixelFormat DSPixelFormat = getMetalPixelFormat(*DSFormat);
+      Desc->setDepthAttachmentPixelFormat(DSPixelFormat);
+      if (*DSFormat == Format::D32FloatS8Uint)
```
MarijnS95: Hmm we have an `isDepthFormat()` but for longevity an `isStencilFormat()` might be useful too, in case a new format is ever added.

lib/API/MTL/MTLDevice.cpp
```diff
@@ -787,6 +636,152 @@ class MTLDevice : public offloadtest::Device {
...
+        AttrDesc->setFormat(getMetalVertexFormat(Elem.Format));
+        VtxDesc->attributes()->setObject(AttrDesc, I);
+        Stride = std::max(Stride, Elem.OffsetInBytes + ElemSize);
+      }
+
+      MTL::VertexBufferLayoutDescriptor *LDesc =
+          MTL::VertexBufferLayoutDescriptor::alloc()->init();
+      LDesc->setStride(Stride);
+      LDesc->setStepRate(1);
+      LDesc->setStepFunction(MTL::VertexStepFunctionPerVertex);
```
MarijnS95, line 815: You added `PerInstance` to Dx12, also add it here for consistency?

lib/API/MTL/MTLDevice.cpp
manon-traverse: Yup, this is something for the future.
The .test files should make describing the buffers and how they are bound as easily as possible. However, the render backend api should be able to get a very detailed description of how things are bound.
Once we have the full abstraction this will just be a single conversion step in the app that runs the .test files.

manon-traverse:
> * **Depth/stencil disabled by default** in `createPipelineVsPs`: old had `DepthEnable=true`, `DepthWriteMask=ALL`, `DepthFunc=LESS`. New has `DepthEnable=false`, `StencilEnable=false`. The caller now passes `DSFormat` to control this, but the PSO itself no longer enables depth testing.
> *

Looks like something went wrong here. It seems like it wasn't enabled in DX12 when I initially worked on this branch but it got enabled in the mean-time. Let me enable depth stencil testing.

lib/API/VK/Device.cpp
```diff
@@ -862,6 +876,393 @@ class VulkanDevice : public offloadtest::Device {
...
+        if (!SeenConstantIDs.insert(SpecConst.ConstantID).second)
+          return llvm::createStringError(
+              std::errc::invalid_argument,
+              "Test configuration contains multiple entries for "
+              "specialization constant ID %u.",
+              SpecConst.ConstantID);
+
+        VkSpecializationMapEntry Entry;
+        if (auto Err = parseSpecializationConstant(SpecConst, Entry, SpecData))
+          return Err;
```
MarijnS95, line 980: Nit (even though it makes things more ugly): these two error paths don't seem to clean up resources like the existing error paths.

lib/API/VK/Device.cpp
```diff
@@ -1909,181 +2233,6 @@ class VulkanDevice : public offloadtest::Device {
...
-    MultisampleStateCI.sType =
-        VK_STRUCTURE_TYPE_PIPELINE_MULTISAMPLE_STATE_CREATE_INFO;
-    MultisampleStateCI.rasterizationSamples = VK_SAMPLE_COUNT_1_BIT;
-
-    const uint32_t Stride = P.Bindings.getVertexStride();
-
-    VkVertexInputBindingDescription VertexInputBinding{};
-    VertexInputBinding.binding = 0;
-    VertexInputBinding.stride = Stride;
-    VertexInputBinding.inputRate = VK_VERTEX_INPUT_RATE_VERTEX;
```
MarijnS95, line 2007: Instance?

lib/API/VK/Device.cpp
manon-traverse: Fixing these!

lib/API/VK/Device.cpp
manon-traverse: I am going to disable instanced rendering for now. It is not used in tests, we should add proper support when we build tests that use it.

lib/API/MTL/MTLDevice.cpp
manon-traverse: Disabling instanced rendering features for now. Added an assert for when we want to use it that we need to properly implement that.

lib/API/MTL/MTLDevice.cpp
manon-traverse: All cleaned up!

lib/API/VK/Device.cpp
```diff
@@ -605,6 +612,30 @@ llvm::Error VulkanQueue::submit(
...
 }
+
+class VulkanPipelineState : public offloadtest::PipelineState {
+public:
+  std::string Name;
+  VkDevice Dev;
+  VkPipeline Pipeline;
+  VkPipelineLayout Layout;
+  llvm::SmallVector<VkDescriptorSetLayout>
+      SetLayouts; // Do we need to keep these alive?
```
MarijnS95: Apparently yes, as long as `vkUpdateDescriptorSets()` can/is going to be called by this code.

lib/API/VK/Device.cpp
```diff
@@ -1445,81 +1867,22 @@ class VulkanDevice : public offloadtest::Device {
...
 
   llvm::Error createDescriptorSets(Pipeline &P, InvocationState &IS) {
-    for (const auto &S : P.Sets) {
-      std::vector<VkDescriptorSetLayoutBinding> Bindings;
-      for (const auto &R : S.Resources) {
-        VkDescriptorSetLayoutBinding Binding = {};
-        if (!R.VKBinding.has_value())
-          return llvm::createStringError(std::errc::invalid_argument,
-                                         "No VulkanBinding provided for '%s'",
-                                         R.Name.c_str());
```
MarijnS95, line 1420: This check didn't seem to survive the code-move, replacing the explicit error with a (supposedly) crash when (optional) `VulkanBinding` was not provided in Yaml?

lib/API/VK/Device.cpp
```diff
@@ -862,6 +876,414 @@ class VulkanDevice : public offloadtest::Device {
...
+    PipelineCI.pVertexInputState = &VertexInputCI;
+    PipelineCI.pInputAssemblyState = &InputAssemblyCI;
+    PipelineCI.pViewportState = &ViewportCI;
+    PipelineCI.pRasterizationState = &RastCI;
+    PipelineCI.pMultisampleState = &MultisampleCI;
+    PipelineCI.pDepthStencilState = &DepthStencilCI;
+    PipelineCI.pColorBlendState = &BlendCI;
+    PipelineCI.pDynamicState = &DynamicCI;
+    PipelineCI.layout = PipelineLayout;
+    PipelineCI.renderPass = VK_NULL_HANDLE;
```
MarijnS95: Claude:

> **VK dynamic rendering** is the biggest risk. The pipeline is now created with `renderPass = VK_NULL_HANDLE` + `VkPipelineRenderingCreateInfo`, but the command encoding path still uses the old `VkRenderPass`-based `vkCmdBeginRenderPass`. Mixing dynamic-rendering pipeline creation with legacy render pass commands is technically invalid per the Vulkan spec — pipelines created with dynamic rendering must be used with `vkCmdBeginRendering`, not `vkCmdBeginRenderPass`. This could silently work on some drivers and fail on others.

MarijnS95: Running with golden images and validation layer enabled (`-DGOLDENIMAGE_DIR=/path/to/golden-images -DOFFLOADTEST_ENABLE_VALIDATION=ON`):

```
  Validation Error: [ VUID-vkCmdDraw-renderPass-02684 ] | MessageID = 0x50685725
  vkCmdDraw(): subpassCount is incompatible between VkRenderPass 0xf000000000f
  (from VkCommandBuffer 0x5625f3c3bf50) and VkRenderPass 0x0 (from VkPipeline
  0x140000000014), 1 != 0
  The Vulkan spec states: The current render pass must be compatible with the
  renderPass member of the VkGraphicsPipelineCreateInfo structure specified when
  creating the VkPipeline bound to VK_PIPELINE_BIND_POINT_GRAPHICS
  (https://docs.vulkan.org/spec/latest/chapters/drawing.html#VUID-vkCmdDraw-renderPass-02684)
  Objects: 3
      [0] VkCommandBuffer 0x5625f3c3bf50
      [1] VkRenderPass 0xf000000000f
      [2] VkPipeline 0x140000000014

  Validation Error: [ VUID-vkCmdDraw-renderPass-02684 ] | MessageID = 0x50685725
  vkCmdDraw(): dependencyCount is incompatible between VkRenderPass 0xf000000000f
  (from VkCommandBuffer 0x5625f3c3bf50) and VkRenderPass 0x0 (from VkPipeline
  0x140000000014), 2 != 0
  The Vulkan spec states: The current render pass must be compatible with the
  renderPass member of the VkGraphicsPipelineCreateInfo structure specified when
  creating the VkPipeline bound to VK_PIPELINE_BIND_POINT_GRAPHICS
  (https://docs.vulkan.org/spec/latest/chapters/drawing.html#VUID-vkCmdDraw-renderPass-02684)
  Objects: 3
      [0] VkCommandBuffer 0x5625f3c3bf50
      [1] VkRenderPass 0xf000000000f
      [2] VkPipeline 0x140000000014
```

MarijnS95 (approved):
Nice, looks like after re-adding render passes (switching to "Dynamic Rendering" is a bigger feat that is better suited to its own isolated PR) this seems to be ready to go!

lib/API/VK/Device.cpp
manon-traverse: I've added the check back but moved it up one layer!

lib/API/VK/Device.cpp
manon-traverse: I have removed the dynamic rendering feature. It's now using classic RenderPass objects.

llvm-beanz (approved):

lib/API/DX/Device.cpp
```diff
@@ -611,6 +627,217 @@ class DXDevice : public offloadtest::Device {
...
+    }
+
+    return llvm::Error::success();
+  }
+
+  llvm::Error createRootSignatureFromBindingsDesc(
+      llvm::StringRef, const BindingsDesc &BindingsDesc, bool IsGraphics,
+      ComPtr<ID3D12RootSignature> &OutRootSignature) {
+    uint32_t DescriptorCount = 0;
+    for (auto &D : BindingsDesc.DescriptorSetDescs)
```
bogner, line 671: Slightly more readable to spell out `DescriptorSetLayoutDesc` here rather than auto. Also shouldn't this be `const`?

lib/API/DX/Device.cpp
```diff
@@ -611,6 +627,217 @@ class DXDevice : public offloadtest::Device {
...
+
+    std::vector<D3D12_ROOT_PARAMETER> RootParams;
+    const std::unique_ptr<D3D12_DESCRIPTOR_RANGE[]> Ranges =
+        std::unique_ptr<D3D12_DESCRIPTOR_RANGE[]>(
+            new D3D12_DESCRIPTOR_RANGE[DescriptorCount]);
+    uint32_t RangeIdx = 0;
+    for (const auto &D : BindingsDesc.DescriptorSetDescs) {
+      uint32_t DescriptorIdx = 0;
+      const uint32_t StartRangeIdx = RangeIdx;
+      for (const auto &R : D.ResourceBindings) {
```
bogner: Let's spell the types for these loops as well

lib/API/DX/Device.cpp
```diff
@@ -611,6 +627,217 @@ class DXDevice : public offloadtest::Device {
...
+      llvm::StringRef, const BindingsDesc &BindingsDesc, bool IsGraphics,
+      ComPtr<ID3D12RootSignature> &OutRootSignature) {
+    uint32_t DescriptorCount = 0;
+    for (auto &D : BindingsDesc.DescriptorSetDescs)
+      DescriptorCount += D.ResourceBindings.size();
+
+    std::vector<D3D12_ROOT_PARAMETER> RootParams;
+    const std::unique_ptr<D3D12_DESCRIPTOR_RANGE[]> Ranges =
+        std::unique_ptr<D3D12_DESCRIPTOR_RANGE[]>(
+            new D3D12_DESCRIPTOR_RANGE[DescriptorCount]);
```
bogner:
```suggestion
    const std::unique_ptr<D3D12_DESCRIPTOR_RANGE[]> Ranges(
        new D3D12_DESCRIPTOR_RANGE[DescriptorCount]);
```

lib/API/DX/Device.cpp
```diff
@@ -1485,8 +1600,10 @@ class DXDevice : public offloadtest::Device {
       IS.CB->CmdList->SetDescriptorHeaps(1, Heaps);
       Handle = IS.DescHeap->GetGPUDescriptorHandleForHeapStart();
     }
-    IS.CB->CmdList->SetComputeRootSignature(IS.RootSig.Get());
-    IS.CB->CmdList->SetPipelineState(IS.PSO.Get());
+    const DXPipelineState *DXPipeline =
+        static_cast<const DXPipelineState *>(IS.Pipeline.get());
```
bogner: We have LLVM-style RTTI set up for the DXPipelineState, so we should be able to use `llvm::cast` here, no? That will assert if we get it wrong.

Note that this also applies later in this file and in MTL and VK's implementations.

lib/API/MTL/MTLDevice.cpp
```diff
@@ -787,6 +642,196 @@ class MTLDevice : public offloadtest::Device {
...
+      llvm::ArrayRef<Format> RTFormats, std::optional<Format> DSFormat,
+      ShaderContainer VS, ShaderContainer PS) override {
+    NS::Error *Error = nullptr;
+
+    // Load vertex shader.
+    const llvm::StringRef VSProgram = VS.Shader->getBuffer();
+    dispatch_data_t VSData = dispatch_data_create(
+        VSProgram.data(), VSProgram.size(), dispatch_get_main_queue(),
+        ^{
+        });
```
bogner, line 715: I guess if this is what clang-format says we should leave it as is for now, but this is a really weird formatting choice. Might be worth filing an issue on clang-format to improve it.

lib/API/MTL/MTLDevice.cpp
```diff
@@ -787,6 +642,196 @@ class MTLDevice : public offloadtest::Device {
...
+
+    // Build vertex descriptor from InputLayout.
+    if (!InputLayout.empty()) {
+      NS::Array *FnAttrs = VSFn->vertexAttributes();
+      // I'm not really sure if there's any valid case for a vertex shader with
+      // no vertex attributes, so we just error if that ever occurs.
+      // NOTE(manon): There is a very valid use case, loading vertices from a
+      // raw buffer using just the index acquried from the index buffer. A lot
+      // of modern hardware have the driver just turns this input layout vertex
+      // loading into shader code anyway.
```
bogner: I find this comment and note a little hard to follow. I guess we could reword it to say that we'd need to update this if we want to support the described use case, but we just error for now?

lib/API/MTL/MTLDevice.cpp
```diff
@@ -787,6 +642,196 @@ class MTLDevice : public offloadtest::Device {
...
+      if (FnAttrs->count() != InputLayout.size())
+        return llvm::createStringError(
+            std::errc::invalid_argument,
+            "Mismatch between vertex shader attribute count and pipeline "
+            "vertex input count.");
+
+      // Collect the attribute indices the shader expects so that we can map the
+      // specified attributes onto the correct indices.
+      llvm::StringMap<uint32_t> ShaderAttrIndices;
+      for (uint32_t Ai = 0; Ai < FnAttrs->count(); ++Ai) {
```
bogner: I guess `Ai` is for "attribute index" here? I guess llvm style would spell that `AI`, but maybe `AttrI`, `AttrIdx`, or even just `I` would be clearer.

lib/API/MTL/MTLDevice.cpp
```diff
@@ -787,6 +642,196 @@ class MTLDevice : public offloadtest::Device {
...
+    // Load pixel/fragment shader.
+    const llvm::StringRef PSProgram = PS.Shader->getBuffer();
+    dispatch_data_t PSData = dispatch_data_create(
+        PSProgram.data(), PSProgram.size(), dispatch_get_main_queue(),
+        ^{
+        });
+    MTL::Library *PSLib = Device->newLibrary(PSData, &Error);
+    if (Error) {
+      VSFn->release();
+      VSLib->release();
```
bogner: It would be better to use an RAII object to manage releasing the live set of metal objects. You can use `llvm::scope_exit` here and manually call `scope_exit::release()` to disengage it in the non-error case.

Note that there are a few different options on how to manage this, and I don't have a strong opinion on which is best:
- A single scope_exit "OnError" that checks which of the objects are non-null and releases each of them. Downside is that we need to delare each of these pointers up front and explicitly null them.
- A scope exit per object that needs to be released. Downside is that we need to call release on each one.
- A SmallVector of scope_exit objects?

lib/API/VK/Device.cpp
```diff
@@ -813,6 +832,415 @@ class VulkanDevice : public offloadtest::Device {
...
+      VkDescriptorSetLayoutCreateInfo SetCI = {};
+      SetCI.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_LAYOUT_CREATE_INFO;
+      SetCI.bindingCount = static_cast<uint32_t>(Binds.size());
+      SetCI.pBindings = Binds.data();
+      VkDescriptorSetLayout SetLayout = VK_NULL_HANDLE;
+      if (auto Err = VK::toError(
+              vkCreateDescriptorSetLayout(Device, &SetCI, nullptr, &SetLayout),
+              "Failed to create descriptor set layout.")) {
+        for (auto *L : SetLayouts)
+          vkDestroyDescriptorSetLayout(Device, L, nullptr);
```
bogner, line 887: Can we add an assert at the beginning of `createPipelineLayout` that `SetLayouts` is empty? If not, this would do the wrong thing here.

lib/API/VK/Device.cpp
```diff
@@ -813,6 +832,415 @@ class VulkanDevice : public offloadtest::Device {
...
+    LayoutCI.sType = VK_STRUCTURE_TYPE_PIPELINE_LAYOUT_CREATE_INFO;
+    LayoutCI.setLayoutCount = static_cast<uint32_t>(SetLayouts.size());
+    LayoutCI.pSetLayouts = SetLayouts.empty() ? nullptr : SetLayouts.data();
+    LayoutCI.pushConstantRangeCount = static_cast<uint32_t>(Ranges.size());
+    LayoutCI.pPushConstantRanges = Ranges.empty() ? nullptr : Ranges.data();
+    if (auto Err = VK::toError(
+            vkCreatePipelineLayout(Device, &LayoutCI, nullptr, &PipelineLayout),
+            "Failed to create pipeline layout.")) {
+      for (auto *L : SetLayouts)
+        vkDestroyDescriptorSetLayout(Device, L, nullptr);
```
bogner, line 911: We should also be using scope_exit for the cleanup of `SetLayouts`, here and in the similar code in the callers of  `createPipelineLayout`

bogner (approved):
A few minor comments. The biggest one will be the comment about the error handling strategy in various places. Generally LGTM though, shouldn't need another review round from me once those are addressed.

lib/API/MTL/MTLDevice.cpp
manon-traverse: I think option two is the simplest and cleanest. I'm opting for that one.
I'm leaving out scope_exit for objects that are release without any branching happening between creation and release.

lib/API/MTL/MTLDevice.cpp
manon-traverse: I'm not sure if clang-format was designed for this apple Objective-C-C++ hybrid code.

manon-traverse:
CI failure is unrelated to this PR.

PR #1073, opened 2026-04-09
Add WaveActiveMin Tests

### LINKED ISSUE #913: Add test for `WaveActiveMin`
- [ ] Test for 16 bit int types (`WaveActiveMin.int16.test`)
- [ ] Test for Half type (`WaveActiveMin.fp16.test`)
- [ ] Test for 32 bit types (`WaveActiveMin.32.test`)
- [ ] Test for 64 bit int types (`WaveActiveMin.int64.test`)
- [ ] Test for Double type (`WaveActiveMin.fp64.test`)

A .test file should contain 
```
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# RUN: split-file %s %t
```

bob80905:
Adds WaveActiveMin tests. Uses WaveActiveMax tests as a reference.
Fixes https://github.com/llvm/offload-test-suite/issues/913

Assisted by: Github Copilot


manon-traverse (approved):

Icohedron (approved):

test/WaveOps/WaveActiveMin.fp16.test
```diff
@@ -0,0 +1,317 @@
...
+            half4 v = In[ValueSet * ValueSetOffset + MaskIdx * NUM_THREADS + tid.x];
+            const uint OutIdx = ValueSetOffset + MaskIdx * NUM_THREADS + tid.x;
+            if (Masks[MaskIdx].mask[tid.x]) {
+                Out1[OutIdx] = WaveActiveMin( v.x );
+                Out2[OutIdx].xy = WaveActiveMin( v.xy );
+                Out3[OutIdx].xyz = WaveActiveMin( v.xyz );
+                Out4[OutIdx] = WaveActiveMin( v );
+            }
+        }
+    }
```
joaosaffran, line 34: This seems too complicated. Could something simpler:
```cpp
Out4[tid] = WaveActiveMin( 10 - tid );
```

work as well?

test/WaveOps/WaveActiveMin.test
```diff
@@ -0,0 +1,106 @@
...
+    Data: [ nan, nan, nan, nan ]
+  - Name: Infs
+    Format: Float32
+    Data: [ inf, inf, inf, inf ]
+  - Name: NegInfs
+    Format: Float32
+    Data: [ -inf, -inf, -inf, -inf ]
+  - Name: Mix
+    Format: Float32
+    Data: [ inf, -inf, nan, 0 ]
```
joaosaffran: Testing nan, infs and denorm is not a good idea. IHVs don't deal with them consistently in the drivers, QC is particularly sensitive to them. And might cause false failing tests.

joaosaffran (changes requested):

test/WaveOps/WaveActiveMin.fp16.test
bob80905: For the number of values that I am currently testing, I would need many more output buffer resources. The reason OutIdx is used is to account for which value set I'm using, and which thread mask. The complexity is just to consolidate the size of the test into one index, and the basis of this is already in the WaveActiveMax tests.

joaosaffran:
Also, can we add the `test-all` label and add push another commit, to trigger it?

joaosaffran (approved):

PR #1074, opened 2026-04-09
Add WavePrefixProduct Tests

### LINKED ISSUE #914: Add test for `WavePrefixProduct`
- [ ] Test for 16 bit int types (`WavePrefixProduct.int16.test`)
- [ ] Test for Half type (`WavePrefixProduct.fp16.test`)
- [ ] Test for 32 bit types (`WavePrefixProduct.32.test`)
- [ ] Test for 64 bit int types (`WavePrefixProduct.int64.test`)
- [ ] Test for Double type (`WavePrefixProduct.fp64.test`)

A .test file should contain 
```
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# RUN: split-file %s %t
```

bob80905:
Adds WavePrefixProduct tests.
Based on WavePrefixSum, except the first lane returns 1 instead of 0.
Fixes https://github.com/llvm/offload-test-suite/issues/914

### DIFF
--- a/test/WaveOps/WavePrefixProduct.32.test
+++ b/test/WaveOps/WavePrefixProduct.32.test
@@ -469,7 +469,7 @@ DescriptorSets:
 # Bug: https://github.com/llvm/llvm-project/issues/186151
 # XFAIL: Clang && Vulkan
 
-# Bug: https://github.com/llvm/offload-test-suite/issues/960
+# Bug: https://github.com/llvm/offload-test-suite/issues/351
 # XFAIL: Metal
 
 # Bug: https://github.com/llvm/offload-test-suite/issues/961
--- a/test/WaveOps/WavePrefixProduct.32.test
+++ b/test/WaveOps/WavePrefixProduct.32.test
@@ -471,8 +471,14 @@ DescriptorSets:
 ...
 #--- end
 
+# Bug: https://github.com/llvm/offload-test-suite/issues/1083
+# XFAIL: Clang && Vulkan
+
 # Bug: https://github.com/llvm/offload-test-suite/issues/351
 # XFAIL: Metal
+
+# Bug: https://github.com/llvm/offload-test-suite/issues/1084
+# XFAIL: NV && Clang && DirectX
 
 # RUN: split-file %s %t
 # RUN: %dxc_target -T cs_6_5 -Fo %t.o %t/source.hlsl
--- a/test/WaveOps/WavePrefixProduct.32.test
+++ b/test/WaveOps/WavePrefixProduct.32.test
@@ -315,23 +315,28 @@ Results:
     Actual: UOut5
     Expected: UExpectedOut5
   - Result: FExpectedOut1
-    Rule: BufferExact
+    Rule: BufferFloatULP
+    ULPT: 1
     Actual: FOut1
     Expected: FExpectedOut1
   - Result: FExpectedOut2
-    Rule: BufferExact
+    Rule: BufferFloatULP
+    ULPT: 1
     Actual: FOut2
     Expected: FExpectedOut2
   - Result: FExpectedOut3
-    Rule: BufferExact
+    Rule: BufferFloatULP
+    ULPT: 1
     Actual: FOut3
     Expected: FExpectedOut3
   - Result: FExpectedOut4
-    Rule: BufferExact
+    Rule: BufferFloatULP
+    ULPT: 1
     Actual: FOut4
     Expected: FExpectedOut4
   - Result: FExpectedOut5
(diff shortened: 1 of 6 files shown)


Icohedron (approved):

test/WaveOps/WavePrefixProduct.32.test
```diff
@@ -0,0 +1,480 @@
...
+  - Result: FExpectedOut3
+    Rule: BufferExact
+    Actual: FOut3
+    Expected: FExpectedOut3
+  - Result: FExpectedOut4
+    Rule: BufferExact
+    Actual: FOut4
+    Expected: FExpectedOut4
+  - Result: FExpectedOut5
+    Rule: BufferExact
```
joaosaffran: I think we should have an small ULP here, like 1, for the float outputs

test/WaveOps/WavePrefixProduct.32.test
```diff
@@ -0,0 +1,480 @@
...
+      DirectXBinding:
+        Register: 17
+        Space: 0
+      VulkanBinding:
+        Binding: 17
+
+...
+#--- end
+
+# Bug: https://github.com/llvm/llvm-project/issues/186151
```
joaosaffran: This bug is regarding WavePrefixSum, is the same bug happening in here? If that is the case we should have a different bug anyways

test/WaveOps/WavePrefixProduct.32.test
```diff
@@ -0,0 +1,480 @@
...
+# XFAIL: Clang && Vulkan
+
+# Bug: https://github.com/llvm/offload-test-suite/issues/351
+# XFAIL: Metal
+
+# Bug: https://github.com/llvm/offload-test-suite/issues/961
```
joaosaffran: Another WavePrefixSum specific bug

test/WaveOps/WavePrefixProduct.fp16.test
```diff
@@ -0,0 +1,178 @@
...
+---
+Shaders:
+  - Stage: Compute
+    Entry: main
+    DispatchSize: [ 1, 1, 1 ]
+Buffers:
+  - Name: In
+    Format: Float16
+    Stride: 8
+    Data: [ 0x4000, 0x4200, 0x4500, 0x4700 ]
```
joaosaffran, line 58: Can we get a comment bellow showing what those numbers are please?

test/WaveOps/WavePrefixProduct.fp16.test
```diff
@@ -0,0 +1,178 @@
...
+    Format: Float16
+    Stride: 8
+    Data: [ 0x3c00, 0x3c00, 0x3c00, 0x3c00, 0x4000, 0x4200, 0x4500, 0x4700, 0x4400, 0x4880, 0x4e40, 0x5220, 0x4800, 0x4ec0, 0x57d0, 0x5d5c ]
+  - Name: ExpectedOut5
+    Format: Float16
+    Stride: 8
+    Data: [ 0x3c00, 0x3c00, 0x3c00, 0x3c00, 0x4000, 0x4200, 0x4500, 0x4700, 0x4400, 0x4880, 0x4e40, 0x5220, 0x4800, 0x4ec0, 0x57d0, 0x5d5c ]
+Results:
+  - Result: ExpectedOut1
+    Rule: BufferExact
```
joaosaffran: We definitely need ULP here. It should be higher than the float case, like 2 since I think the offload test expects integers in here.

joaosaffran (commented):
The testing logic looks fine. Left some comments regarding config changes. We need to specify some ULP level when working with floating types.

joaosaffran:
Also, can we add the `test-all` label and push a new commit to trigger another build?

PR #1076, opened 2026-04-10
[d3d12] Rework signal handlers to dump debug info. Prevent rogue pointer access.

manon-traverse:
We can register the signal handler on device creation instead of in the executeProgram function. This brings the three executeProgram implementations closer to each other and causes this debug feature to now always be enabled.

Additionally, there was an issue where a rogue pointer access could occur if the signal handler was triggered after device clean-up. That problem is resolved with this PR as well.

### DIFF
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -56,7 +56,7 @@ static std::mutex SignalHandlerMutex;
 static llvm::SmallVector<ID3D12Device *> SignalHandlerDevices;
 
 static void dumpD3DInfoQueues(void *) {
-  std::lock_guard<std::mutex> Lock(SignalHandlerMutex);
+  const std::lock_guard<std::mutex> Lock(SignalHandlerMutex);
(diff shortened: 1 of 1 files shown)


lib/API/DX/Device.cpp
```diff
@@ -44,13 +44,42 @@
...
 #include <locale>
+#include <mutex>
 
 using namespace offloadtest;
 using Microsoft::WRL::ComPtr;
 
 template <> char CapabilityValueEnum<directx::ShaderModel>::ID = 0;
 template <> char CapabilityValueEnum<directx::RootSignature>::ID = 0;
 
+static std::mutex SignalHandlerMutex;
```
manon-traverse, line 55: The lock may seem overkill as our use cases are single-threaded at the moment. However, the overhead of this mutex is so small it makes sense to have it there in case we ever go into a complex multi-threaded scenario.

MarijnS95 (approved):

damyanp (approved):

PR #1079, opened 2026-04-12
[Vulkan] Support testing multi sampled textures

luciechoi:
Add Vulkan support for testing multisampled texture resources `Texture2DMS` and `Texture2DMSArray`.

The existing CPU-buffer-to-image upload path doesn't initialize multi sampled textures correctly, so the change here was to run another compute pass that writes the test data into multisampled images before the test shader runs.

I also added `SampleCount_N` device feature to run only on those devices that support `N` sample counts.


luciechoi:
cc: @Keenuts 

On whether we can reduce the glsl dependency here: 

I realized we cannot rewrite the data transfer shaders in HLSL in the test files, since `RWTexture2DMS` isn't implemtented in DXC ([FR](https://github.com/microsoft/DirectXShaderCompiler/issues/5244)) (`Texture2DMS` is only read-only)

PR #1085, opened 2026-04-14
Add a VK::toError function to simplify vulkan error handling.

manon-traverse:
Error handling on Vulkan was a bit manual and verbose, with random error codes being assigned. This PR simplifies this using a `VK::toError` function (similar to the `HR::toError` function). This way, we preserve the original VkResult in the error message and don't have arbitrarily chosen error codes.

### DIFF
--- a/include/Support/VkError.h
+++ b/include/Support/VkError.h
@@ -1,0 +1,29 @@
+//===- OffloadTest/Support/VkError.h - Vulkan Error Utils ---------*- C++
+//-*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef OFFLOADTEST_SUPPORT_VKERROR_H
+#define OFFLOADTEST_SUPPORT_VKERROR_H
+
+#include <vulkan/vulkan.h>
+
+#include "llvm/Support/Error.h"
+
+namespace VK {
+inline llvm::Error toError(VkResult Result, llvm::StringRef Msg) {
+  if (Result != VK_SUCCESS)
+    return llvm::createStringError("%s (VkResult = %d)", Msg.data(),
+                                   static_cast<int>(Result));
+  return llvm::Error::success();
+}
+} // namespace VK
+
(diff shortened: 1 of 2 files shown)


include/Support/VkError.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include <vulkan/vulkan.h>
+
+#include "llvm/Support/Error.h"
+
+namespace VK {
+inline llvm::Error toError(VkResult Result, llvm::StringRef Msg) {
+  if (Result != VK_SUCCESS)
+    return llvm::createStringError("%s (VkResult = %d)", Msg.data(),
+                                   static_cast<int>(Result));
```
MarijnS95, line 24: Not sure how much more effort it is to stringify the error codes (or at least a few common ones), and/or to translate a few common ones to `std::error_code`. HR::toError()` seems to do it but not sure if that yields any advantage or if anything "relies" on it...

MarijnS95 (commented):
LGTM

include/Support/VkError.h
manon-traverse: HR::toError just stores it as an integer under `std::system_category`. The only thing in practice is that the error number is logged. This kind of does the same thing.

MarijnS95 (approved):

include/Support/VkError.h
```diff
@@ -0,0 +1,29 @@
+//===- OffloadTest/Support/VkError.h - Vulkan Error Utils ---------*- C++
+//-*-===//
```
bogner: I think clang-format mangled this.

include/Support/VkError.h
```diff
@@ -0,0 +1,29 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+//
+//===----------------------------------------------------------------------===//
```
bogner: No need for the empty comment block

include/Support/VkError.h
```diff
@@ -0,0 +1,29 @@
...
+
+#ifndef OFFLOADTEST_SUPPORT_VKERROR_H
+#define OFFLOADTEST_SUPPORT_VKERROR_H
+
+#include <vulkan/vulkan.h>
+
+#include "llvm/Support/Error.h"
```
bogner: Doesn't really matter, but I think we usually put the LLVM includes before the more system-y ones

bogner (approved):

PR #1086, opened 2026-04-14
Address post-merge review comments on CommandBuffer (#1033)

MarijnS95:
Address post-review comments from @bogner on #1033. These felt significant enough that they weren't worth cramming into an "unrelated" PR but deserve their own clean follow-up in three individual commits.

- Add virtual method anchors for all polymorphic types (`Buffer`, `CommandBuffer`, `Fence`, `ImageComparatorBase`, `ImageComparatorDiffImage`, `Texture`) per LLVM coding standards
- Add file description to header comment
- Switch to LLVM-style RTTI (`classof()`) instead of hand-rolled `as<T>()` + `BackendAPI`, fixing `-Wunused-const-variable`

Adding `-Werror` to the build is left for a separate PR as it requires a warning audit first — the project inherits compile flags from LLVM's CMake infrastructure and backend SDK headers may introduce additional warnings.

### DIFF
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -414,7 +414,9 @@ public:
 
 class DXCommandBuffer : public offloadtest::CommandBuffer {
 public:
-  static constexpr GPUAPI BackendAPI = GPUAPI::DirectX;
+  static bool classof(const CommandBuffer *CB) {
+    return CB->getKind() == GPUAPI::DirectX;
+  }
 
   ComPtr<ID3D12CommandAllocator> Allocator;
   ComPtr<ID3D12GraphicsCommandList> CmdList;
--- a/include/API/Buffer.h
+++ b/include/API/Buffer.h
@@ -22,7 +22,7 @@ struct BufferCreateDesc {
 
 class Buffer {
 public:
-  virtual ~Buffer() = default;
+  virtual ~Buffer();
 
   Buffer(const Buffer &) = delete;
   Buffer &operator=(const Buffer &) = delete;
--- a/include/API/CommandBuffer.h
+++ b/include/API/CommandBuffer.h
@@ -6,6 +6,10 @@
 //
 //===----------------------------------------------------------------------===//
 //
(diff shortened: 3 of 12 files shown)


lib/API/DX/Device.cpp
```diff
@@ -391,7 +391,9 @@ class DXQueue : public offloadtest::Queue {
 
 class DXCommandBuffer : public offloadtest::CommandBuffer {
 public:
-  static constexpr GPUAPI BackendAPI = GPUAPI::DirectX;
+  static bool classof(const CommandBuffer *CB) {
+    return CB->getKind() == GPUAPI::DirectX;
+  }
```
manon-traverse, line 419: Nit: don't like the name. But I don't really have an alternative. Also, this doesn't seem to be called?

MarijnS95: https://llvm.org/docs/HowToSetUpLLVMStyleRTTI.html

> A common naming convention is that these enums are “kind”s, to avoid ambiguity with the words “type” or “class” which have overloaded meanings in many contexts within LLVM. Sometimes there will be a natural name for it, like “opcode”. Don’t bikeshed over this; when in doubt use Kind.

---

> Also, this doesn't seem to be called?

Probably for the same reason as https://github.com/llvm/offload-test-suite/pull/1033#discussion_r3074256827; it'll get used (via the builtin `isa<>`/`dyn_cast<>`) when at least `Queue::submit()` lands, etc.

manon-traverse: Let's do it this way then!

bogner (approved):

MarijnS95:
@bogner FYI I asked Claude to find/summarize more classes that lack anchors; these are `Fence`/`Buffer` (our new types) and the preexisting `ImageComparatorBase`/`ImageComparatorDiffImage`. I'll push an updated commit that fixes all of them shortly.

EDIT: And the new `Texture` too.

manon-traverse (approved):

manon-traverse:
CI failure is unrelated to this PR, so I am merging this 👍

PR #1087, opened 2026-04-14
New `VertexBuffer` YAML definition using new `Format` enum

EmilioLaiso:
## Summary

This PR introduces a new `VertexBuffers` YAML schema and vertex buffer support into the offload test suite, replacing the previous approach where vertex data was defined as a raw `CPUBuffer` entry under `Buffers:` and manually described via `VertexAttributes:` in `Bindings:`.

### New YAML schema: `VertexBuffers`

A new `VertexBuffers:` section is added to test pipeline definitions. 
Each vertex buffer declares named streams with a typed `Format` and inline `Data`:

```yaml
VertexBuffers:
  - Name: VB
    Streams:
      - Name: POSITION
        Format: RGB32Float
        Data: [...]
Bindings:
  VertexBuffer: VB
```

This replaces the old pattern where vertex data lived in `Buffers:` and layout was specified separately in `Bindings: > VertexAttributes:`. `Channels`, `Offsets` and `Stride` are now inferred.

```yaml
# Old (removed)
Buffers:
  - Name: VB
    Format: Float32
    Channels: 3
    Data: [...]
Bindings:
  VertexBuffer: VB
  VertexAttributes:
    - Format: Float32
      Channels: 3
      Offset: 0
      Name: POSITION
```

### Parsed representation

- **`ParsedVertexBuffer` / `VertexStreamData`** (`include/Support/Pipeline.h`) -- YAML-side parsed representation. Stream values are stored as `double` and then interleaved + type-converted during parsing into a single `InterleavedData` buffer ready for GPU upload. Provides helpers for computing stride, per-stream byte offsets, and vertex count.

### `BufferUsage` enum and per-backend changes

A `BufferUsage` enum (`Storage` / `VertexBuffer`) is added to `BufferCreateDesc` so backends can set the correct usage flags. Each backend creates its vertex buffer directly via `createBuffer` with `VertexBuffer` usage, and derives the input layout from the parsed pipeline (`ParsedVertexBuffer`) at PSO/pipeline creation time:
- **Vulkan** (`lib/API/VK/Device.cpp`): sets `VK_BUFFER_USAGE_VERTEX_BUFFER_BIT` for vertex buffers; eliminates the old host-staging + device-copy pair in favor of a single `CpuToGpu` mapped buffer (with a TODO to revisit for discrete GPUs).
- **DirectX 12** (`lib/API/DX/Device.cpp`): uses `D3D12_HEAP_TYPE_UPLOAD` for `CpuToGpu` vertex buffers; builds `D3D12_INPUT_ELEMENT_DESC` from `ParsedVertexBuffer` streams instead of `VertexAttribute`.
- **Metal** (`lib/API/MTL/MTLDevice.cpp`): builds `MTL::VertexDescriptor` from `ParsedVertexBuffer` streams; copies interleaved data into a managed Metal buffer.

### `Buffer::getSizeInBytes()`

A new pure virtual `getSizeInBytes()` is added to `Buffer`, implemented by `DXBuffer`, `VulkanBuffer`, and `MTLBuffer`.

### `Resources.h` helpers

- `getComponentCount(Format)` -- returns the number of components per element for a format.

### Pipeline parser changes (`lib/Support/Pipeline.cpp`)

- Parses the new `VertexBuffers:` section with `VertexStreamData` mapping (Name, Format, Data).
- Validates format compatibility (`isVertexCompatible`), data-size-to-component-count alignment, and consistent vertex count across streams.
- Interleaves per-stream double values into a packed byte buffer (`InterleavedData`) with format-specific type conversion via `writeComponent()`.
- Resolves `Bindings::VertexBuffer` against `Pipeline::VertexBuffers` instead of `Pipeline::Buffers`.
- Removes `VertexAttribute` mapping and the `VertexAttributes` binding key.
- Adds `ScalarEnumerationTraits<Format>` so that the new `Format` enum can be used directly in YAML fields.

### Test migrations

All 22 graphics test files are migrated from the old schema to the new `VertexBuffers:` + `Streams:` format.


include/API/VertexBuffer.h
```diff
@@ -0,0 +1,66 @@
...
+  VertexBufferDesc Desc;
+  std::shared_ptr<Buffer> Data;
+
+  uint32_t getVertexCount() const {
+    uint32_t Stride = Desc.getStride();
+    if (Stride == 0)
+      return 0;
+    return static_cast<uint32_t>(Data->getSizeInBytes()) / Stride;
+  }
+};
```
manon-traverse: I don't think this is the right abstraction.
The layout is part of the pipelinelayout/inputlayout which is part of the PipelineState.
A vertex buffer should just be `Buffer` type with the `VertexBuffer` usage.

EmilioLaiso: In the future we want to build RT acceleration structures, right?
It might be useful to have this information at hand when filling BLAS descriptions as you'll need the vertex format and stride.

manon-traverse: For acceleration structures we don't need the semantics name, just offset and stride for the positions. I would first like to see how this fits into my PR https://github.com/llvm/offload-test-suite/pull/1070 as this creates an interface for specifying the input layout. After the semantics have been specified we don't need them anymore to bind the buffer.

EmilioLaiso:
Use of `shared_ptr` will be replaced with `unique_ptr` after https://github.com/llvm/offload-test-suite/pull/1094 has gone in

include/API/VertexBuffer.h
EmilioLaiso: Removed the `VertexBuffer` abstraction :)

test/Feature/Semantics/ArraySemantics.test
```diff
@@ -52,11 +54,6 @@ Buffers:
       Depth: 1
 Bindings:
   VertexBuffer: VertexData
-  VertexAttributes:
-    - Format: Float32
-      Channels: 4
-      Offset: 0
-      Name: POSITION
```
manon-traverse, line 59: Without this data, we cannot initialize the input layout on pipelines. The current format allows the test creator to provide data closer to how the API works.
Also, I think we should avoid changing the test format unless it's _really_ necessary.

test/Feature/Semantics/ArraySemantics.test
```diff
@@ -35,13 +35,15 @@ Shaders:
...
+VertexBuffers:
   - Name: VertexData
-    Format: Float32
-    Stride: 16
-    Data: [  0.0,  3.0, 0.0, 1.0,
-             3.0, -3.0, 0.0, 1.0,
-            -3.0, -3.0, 0.0, 1.0 ]
+    Streams:
+      - Name: POSITION
+        Format: RGBA32Float
```
manon-traverse, line 42: I do think this format specifier is better. I don't know the exact allowed formats for vertex buffers on top of my head, but I think they are pretty similar to texture formats.

PR #1094, opened 2026-04-16
Address post-merge review comments on #1020 and use LLVM-style RTTI

EmilioLaiso:
Follow-up on the discussions at: https://github.com/llvm/offload-test-suite/pull/1020#discussion_r3073568722 & https://github.com/llvm/offload-test-suite/pull/1020#discussion_r3063962742

This PR does the following:
- Use `offloadtest::Texture` and `offloadtest::Buffer` types in backends `InvocationState` 
- Switch the usage of `shared_ptr` for those types to `unique_ptr`
- Add LLVM-style RTTI on `Buffer` and `Texture`

### DIFF
--- a/include/API/Buffer.h
+++ b/include/API/Buffer.h
@@ -24,17 +24,17 @@ struct BufferCreateDesc {
 };
 
 class Buffer {
-  GPUAPI Kind;
+  GPUAPI API;
 
 public:
   virtual ~Buffer();
   Buffer(const Buffer &) = delete;
   Buffer &operator=(const Buffer &) = delete;
 
-  GPUAPI getKind() const { return Kind; }
+  GPUAPI getAPI() const { return API; }
 
 protected:
-  explicit Buffer(GPUAPI Kind) : Kind(Kind) {}
+  explicit Buffer(GPUAPI API) : API(API) {}
 };
 
 } // namespace offloadtest
--- a/include/API/Buffer.h
+++ b/include/API/Buffer.h
@@ -15,6 +15,8 @@
 #include "API/API.h"
 #include "API/Resources.h"
 
+#include "llvm/Support/Casting.h"
+
 namespace offloadtest {
 
 struct BufferCreateDesc {
@@ -25,12 +27,14 @@ class Buffer {
   GPUAPI Kind;
 
 public:
-  explicit Buffer(GPUAPI Kind) : Kind(Kind) {}
   virtual ~Buffer();
   Buffer(const Buffer &) = delete;
   Buffer &operator=(const Buffer &) = delete;
 
   GPUAPI getKind() const { return Kind; }
+
+protected:
+  explicit Buffer(GPUAPI Kind) : Kind(Kind) {}
 };
 
 } // namespace offloadtest
--- a/include/API/Buffer.h
+++ b/include/API/Buffer.h
@@ -34,7 +34,7 @@ public:
   GPUAPI getAPI() const { return API; }
 
 protected:
-  explicit Buffer(GPUAPI API);
+  explicit Buffer(GPUAPI API) : API(API) {}
 };
 
 } // namespace offloadtest
--- a/include/API/Buffer.h
+++ b/include/API/Buffer.h
@@ -34,7 +34,7 @@ public:
   GPUAPI getAPI() const { return API; }
 
 protected:
-  explicit Buffer(GPUAPI API) : API(API) {}
+  explicit Buffer(GPUAPI API);
 };
 
 } // namespace offloadtest
--- a/include/API/Buffer.h
+++ b/include/API/Buffer.h
@@ -12,6 +12,7 @@
 #ifndef OFFLOADTEST_API_BUFFER_H
 #define OFFLOADTEST_API_BUFFER_H
 
+#include "API/API.h"
 #include "API/Resources.h"
 
 namespace offloadtest {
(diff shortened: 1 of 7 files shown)


EmilioLaiso:
CC: @bogner @damyanp @s-perron @manon-traverse

lib/API/DX/Device.cpp
```diff
@@ -479,9 +490,9 @@ class DXDevice : public offloadtest::Device {
     std::unique_ptr<offloadtest::Fence> Fence;
 
     // Resources for graphics pipelines.
-    std::shared_ptr<DXTexture> RT;
-    std::shared_ptr<DXBuffer> RTReadback;
-    std::shared_ptr<DXTexture> DS;
+    std::unique_ptr<offloadtest::Texture> RT;
+    std::unique_ptr<offloadtest::Buffer> RTReadback;
+    std::unique_ptr<offloadtest::Texture> DS;
```
manon-traverse, line 495: Fantastic, using the abstract types brings InvocationState closer to being identical between render backends. The downcasting is needed for now, as we don't have a complete interface yet, but the places where we do cast act as TODO items for what has highest priority to be added to the interface.

lib/API/DX/Device.cpp
```diff
@@ -41,6 +41,7 @@
 
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/Object/DXContainer.h"
+#include "llvm/Support/Casting.h"
```
MarijnS95, line 44: Nit: https://llvm.org/docs/HowToSetUpLLVMStyleRTTI.html#basic-setup recommends to include this in the relevant header(s) so that every consumer has the casting compatible with `getKind()`/`classof` readily available.

MarijnS95 (commented):
LGTM

include/API/Buffer.h
```diff
@@ -21,14 +22,15 @@ struct BufferCreateDesc {
 };
 
 class Buffer {
+  GPUAPI Kind;
+
 public:
+  explicit Buffer(GPUAPI Kind) : Kind(Kind) {}
```
s-perron, line 28: Is there a reason this constructor is public instead of protected like the older constructor? The same for `Texture`.

s-perron: I would expect it to be protected because I don't think we want to create a `Buffer` when it is not a specialized type.

EmilioLaiso: It is now protected as it should've been! 🙌

include/API/Buffer.h
```diff
@@ -12,23 +12,29 @@
...
 class Buffer {
+  GPUAPI Kind;
+
 public:
   virtual ~Buffer();
-
   Buffer(const Buffer &) = delete;
   Buffer &operator=(const Buffer &) = delete;
 
+  GPUAPI getKind() const { return Kind; }
```
MarijnS95: I'm sorry for only noticing this just now while rebasing some other work, as this is a discrepancy introduced on my end in #1086; and think in hindsight that we should probably rename this (and the field) from `Kind` to `API`. Also to signify that it's not "just" a specific kind of a struct but specifically related to the GPU API (i.e. we won't have a `MetalSpecialThingy` and `MetalOtherThingy` subclass for example).

EmilioLaiso: Renamed 👍

include/API/Device.h
```diff
@@ -78,11 +78,11 @@ class Device {
   virtual llvm::Expected<std::unique_ptr<Fence>>
   createFence(llvm::StringRef Name) = 0;
 
-  virtual llvm::Expected<std::shared_ptr<Buffer>>
+  virtual llvm::Expected<std::unique_ptr<Buffer>>
```
manon-traverse, line 81: Just want to make a note that I still think we should go down the shared_ptr route. However that is currently being discussed in issue #1099.
I am okay with going unique_ptr for now because the keep alive system is not place yet. However, we are close to implementing and PRing that and will probably need to switch back to shared_ptr there.

include/API/Texture.h
```diff
@@ -140,14 +142,17 @@ inline llvm::Error validateTextureCreateDesc(const TextureCreateDesc &Desc) {
...
   virtual ~Texture();
-
   Texture(const Texture &) = delete;
   Texture &operator=(const Texture &) = delete;
 
+  GPUAPI getAPI() const { return API; }
+
 protected:
-  Texture() = default;
+  explicit Texture(GPUAPI API) : API(API) {}
```
manon-traverse, line 155: The LLVM code standard prefers defining this function in a .cpp file to reduce binary size.

manon-traverse (approved):

include/API/Texture.h
MarijnS95: IIRC it only requires _just one_ function in a CPP file -typically the destructor - so that it can "anchor" (as it's called) the vtable to that (and I think also these inline functions?). Otherwise every CU has to carry the inline function, to be deduplicated by the linker.

manon-traverse: Thanks @MarijnS95, I think you are correct! The anchor points are indeed just for `virtual` functions as those cannot be inlined. but this can (and probably should) be inlined.

MarijnS95: One funny thing, when I asked Claude to cross-check this (specifically if anchoring also implicitly works for such implicit inline functions), it didn't see that the review was on line 155 but instead complained about the `getAPI()` functions right _above_ that context which probably deserve the same review comment/context... We should either change both (the constructor and the getter) or neither?

EmilioLaiso: Reverted the last commit 👍

MarijnS95 (approved):

lib/API/DX/Device.cpp
```diff
@@ -285,18 +285,27 @@ namespace {
 
 class DXBuffer : public offloadtest::Buffer {
 public:
+  static bool classof(const offloadtest::Buffer *B) {
+    return B->getAPI() == GPUAPI::DirectX;
+  }
```
bogner: Let's put these at the end of the class decls. Most people reading this will be more interested in the members/constructor/etc

bogner (approved):
Generally looks good, just a style comment that applies throughout.

lib/API/DX/Device.cpp
EmilioLaiso: Done ✅

manon-traverse:
CI Failures are unrelated to this clean-up PR. Going to go ahead and merge it since all feedback is addressed.

Issue #351, opened 2025-08-13
Tests `test\WaveOps\WaveReadLaneAt.32.test` and `test\WaveOps\WaveReadLaneAt.16.test` fail under  Metal


bob80905:
Both of these tests fail in the Metal configuration. They have all 0 results, rather than expected results.
Logs:
```
******************** TEST 'OffloadTest-clang-mtl :: WaveOps/WaveReadLaneAt.32.test' FAILED ********************
Exit Code: 1

Command Output (stdout):
--
Signed DxilContainer written to "/var/folders/zv/v734td6n3_3fl4qmc2hcdkcc0000gn/T/lit-tmp-5zk13nlq/source-112ea5.dxo"
Validation succeeded.Using Metal API

--
Command Output (stderr):
--
/Users/cbieneman/actions-runner/_work/offload-test-suite/offload-test-suite/llvm-project/build/bin/split-file /Users/cbieneman/actions-runner/_work/offload-test-suite/offload-test-suite/OffloadTest/test/WaveOps/WaveReadLaneAt.32.test /Users/cbieneman/actions-runner/_work/offload-test-suite/offload-test-suite/llvm-project/build/tools/OffloadTest/test/clang-mtl/WaveOps/Output/WaveReadLaneAt.32.test.tmp # RUN: at line 176
+ /Users/cbieneman/actions-runner/_work/offload-test-suite/offload-test-suite/llvm-project/build/bin/split-file /Users/cbieneman/actions-runner/_work/offload-test-suite/offload-test-suite/OffloadTest/test/WaveOps/WaveReadLaneAt.32.test /Users/cbieneman/actions-runner/_work/offload-test-suite/offload-test-suite/llvm-project/build/tools/OffloadTest/test/clang-mtl/WaveOps/Output/WaveReadLaneAt.32.test.tmp
[...]

PR #1096, opened 2026-04-16
Remove/update XFAILs on tests related to resources in structs

hekota:
These should all be passing since https://github.com/llvm/llvm-project/issues/182989 is completed.

### DIFF
--- a/test/Feature/ResourcesInStructs/array-of-structs-with-res.test
+++ b/test/Feature/ResourcesInStructs/array-of-structs-with-res.test
@@ -100,7 +100,7 @@ DescriptorSets:
 #--- end
 
 # Bug: https://github.com/llvm/offload-test-suite/issues/351
-# XFAIL: Metal
+# XFAIL: Clang && Metal
 
 # RUN: split-file %s %t
 # RUN: %dxc_target -T cs_6_0 -Fo %t.o %t/source.hlsl
--- a/test/Feature/ResourcesInStructs/array-of-structs-with-res.test
+++ b/test/Feature/ResourcesInStructs/array-of-structs-with-res.test
@@ -99,6 +99,10 @@ DescriptorSets:
 ...
 #--- end
 
+# Bug: https://github.com/llvm/offload-test-suite/issues/305
+# Bug: https://github.com/llvm/offload-test-suite/issues/351
+# XFAIL: Metal
+
 # RUN: split-file %s %t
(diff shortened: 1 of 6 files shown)


test/Feature/ResourcesInStructs/res-in-struct-mix.test
```diff
@@ -14,7 +14,6 @@ class G : F {
   E e;
   StructuredBuffer<int> SrvBuf2;
   RWBuffer<int> UavBuf2;
-  uint n;
```
hekota, line 17: I've removed this because:
1. Vulkan does not support structs with resource and non-resource fiels.
2. For Clang-DirectX this creates a $Globals constant buffer that is not used. Because it is not mapped to a CBV resource, the offloat test runner crashes during PSO creation. https://github.com/llvm/llvm-project/issues/192524.

bogner (approved):

test/Feature/ResourcesInStructs/array-of-structs-with-res.test
```diff
@@ -99,9 +99,6 @@ DescriptorSets:
 ...
 #--- end
 
-# Unimplemented https://github.com/llvm/wg-hlsl/issues/367
-# XFAIL: Clang
-
```
bogner, line 104: Needs a metal xfail on #305 / #351

test/Feature/ResourcesInStructs/res-in-struct-simple-array.test
```diff
@@ -70,9 +70,6 @@ DescriptorSets:
 ...
 #--- end
 
-# Unimplemented https://github.com/llvm/wg-hlsl/issues/367
-# XFAIL: Clang
-
```
bogner, line 75: Needs a metal xfail on #305 / #351

bob80905 (approved):

test/Feature/ResourcesInStructs/res-in-struct-mix.test
tex3d: I suppose we could put this kind of case in a new test once we have some more support for it. It seems that for Vulkan targets, it should be splitting out the constant buffer content into a separate struct. So, DXC fails this too on Vulkan?

tex3d: Looking down, I answered my own question. So this is expected to fail on DXC and Vulkan even without this.

tex3d (approved):

PR #1098, opened 2026-04-16
Add test for resources in structs with local variables/args and `this` access

hekota:
Add tests with resources in structs that are testing:
- structure with a resource as a function argument and local variable
- resource in a struct accessed from a struct method

### DIFF
--- a/test/Feature/ResourcesInStructs/res-in-struct-this-access.test
+++ b/test/Feature/ResourcesInStructs/res-in-struct-this-access.test
@@ -7,7 +7,7 @@ struct A {
   RWBuffer<int> Buf;
 
   int get(uint Index) {
-    return Buf[Index];
+    return this.Buf[Index];
   }
 };
 
--- a/test/Feature/ResourcesInStructs/res-in-struct-this-access.test
+++ b/test/Feature/ResourcesInStructs/res-in-struct-this-access.test
@@ -1,0 +1,76 @@
+#--- source.hlsl
+
+// This test verifies handling of a simple resource in a struct
+// when used as a local argument.
+
+struct A {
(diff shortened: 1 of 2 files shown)


test/Feature/ResourcesInStructs/res-in-struct-this-access.test
```diff
@@ -0,0 +1,76 @@
+#--- source.hlsl
+
+// This test verifies handling of a simple resource in a struct
+// when used as a local argument.
+
+struct A {
+  RWBuffer<int> Buf;
+
+  int get(uint Index) {
+    return Buf[Index];
```
alsepkow: nit: Would it be better to be explicit and add 'return this.Buf[Index]' ?

test/Feature/ResourcesInStructs/res-in-struct-this-access.test
```diff
@@ -0,0 +1,76 @@
...
+
+//--- pipeline.yaml
+---
+Shaders:
+  - Stage: Compute
+    Entry: main
+    DispatchSize: [1, 1, 1]
+Buffers:
+  - Name: BufA1
+    Format: Int32
```
alsepkow, line 32: Would there be value in testing for different width resources, like int16?

alsepkow (commented):
LGTUS: Minor comments from me and @bob80905

test/Feature/ResourcesInStructs/res-in-struct-this-access.test
hekota: Not really, the handling is the same for different kinds of resources and data types.

bob80905 (approved):

PR #1100, opened 2026-04-17
Add `ComputeEncoder` abstraction with parallel/serial barrier modes

MarijnS95:
Introduces a generic command encoder abstraction for recording GPU commands to a command buffer. `ComputeEncoder` provides `dispatch()` and `barrier()` operations with two modes: `Parallel` (no automatic barriers, caller manages synchronization) and `Serial` (auto-inserts barriers between commands using tracked destination scope as next source).

Each backend implements barrier tracking natively:
- Metal: `MTL::BarrierScope` accumulated on the encoder
- Vulkan: `VkPipelineStageFlags`/`VkAccessFlags` on the command buffer
- DX12: UAV barrier flag on the command buffer

VK/DX store barrier state on the command buffer (encoders hold a back-reference) so it persists across encoder lifetimes. Metal stores it on the encoder since each `MTL::ComputeCommandEncoder` is a separate native object with implicit inter-encoder ordering.

Creating a parallel encoder flushes a full barrier on VK/DX to ensure prior work is visible. endEncoding() flushes any pending barriers.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>


include/API/Encoder.h
```diff
@@ -0,0 +1,117 @@
...
+  /// Begin a named debug group. Visible in GPU debuggers (PIX, RenderDoc,
+  /// Xcode). Must be balanced by a corresponding popDebugGroup() call.
+  virtual void pushDebugGroup(llvm::StringRef Label) {}
+
+  /// End the most recently pushed debug group.
+  virtual void popDebugGroup() {}
+
+  /// Insert a point-in-time debug marker.
+  virtual void insertDebugSignpost(llvm::StringRef Label) {}
+
```
manon-traverse, line 77: I think these should have an anchor in a .cpp file
https://llvm.org/docs/CodingStandards.html#provide-a-virtual-method-anchor-for-classes-in-headers

include/API/Encoder.h
```diff
@@ -0,0 +1,117 @@
...
+public:
+  using CommandEncoder::CommandEncoder;
+
+  /// Dispatch a compute grid. ThreadsPerGroup specifies the workgroup
+  /// dimensions; GroupCount specifies how many groups to launch in each
+  /// dimension. Total threads = ThreadsPerGroup * GroupCount per axis.
+  virtual llvm::Error dispatch(uint32_t GroupCountX, uint32_t GroupCountY,
+                               uint32_t GroupCountZ, uint32_t ThreadsPerGroupX,
+                               uint32_t ThreadsPerGroupY,
+                               uint32_t ThreadsPerGroupZ) = 0;
```
manon-traverse, line 101: I would like to be able to specify the pipeline and descriptor sets as well. That way, we can enforce that everything that needs to be set up to dispatch has been set up.

Also, could we perhaps grab the threads per group from the pipeline and let the user only specify the group count OR thread count to launch?

include/API/Encoder.h
```diff
@@ -0,0 +1,117 @@
...
+class Buffer;
+
+/// Determines whether an encoder automatically inserts barriers between
+/// commands.
+enum class EncoderMode {
+  /// No automatic barriers. The caller is responsible for inserting barriers.
+  Parallel,
+  /// Automatically insert a barrier between every command.
+  Serial,
+};
```
manon-traverse, line 32: I think this design makes sense!
In most cases using Serial is fine, as the barrier is usually needed, and if not the overhead is quite small.
But when you are doing large batches of dispatches you can still opt for the Parallel version!

Ultra tiny nit: Should be make Serial be the first in this list and add a comment that it should be used by default?

manon-traverse (commented):
Github is not loading all the files for me right now :(
But I am able to provide some feedback on the files that did manage to load.

lib/API/DX/Device.cpp
```diff
@@ -438,12 +457,179 @@ class DXCommandBuffer : public offloadtest::CommandBuffer {
     return CB;
   }
 
+  /// Whether a UAV barrier is pending from a prior compute command.
+  bool PendingUAVBarrier = false;
```
manon-traverse: Please group all the data fields together. It becomes hard to read when data fields and member functions are interleaved.

lib/API/VK/Device.cpp
```diff
@@ -518,18 +537,230 @@ class VulkanCommandBuffer : public offloadtest::CommandBuffer {
...
+
+    CB->CmdBeginDebugUtilsLabel =
+        (PFN_vkCmdBeginDebugUtilsLabelEXT)vkGetDeviceProcAddr(
+            Device, "vkCmdBeginDebugUtilsLabelEXT");
+    CB->CmdEndDebugUtilsLabel =
+        (PFN_vkCmdEndDebugUtilsLabelEXT)vkGetDeviceProcAddr(
+            Device, "vkCmdEndDebugUtilsLabelEXT");
+    CB->CmdInsertDebugUtilsLabel =
+        (PFN_vkCmdInsertDebugUtilsLabelEXT)vkGetDeviceProcAddr(
+            Device, "vkCmdInsertDebugUtilsLabelEXT");
```
manon-traverse: Is there a way we can query these only once instead?

include/API/Encoder.h
MarijnS95: I think reading back the threads-per-group from reflection is the most convenient. While Metal has ways to specify the total dispatch in thread-count, with automatic bounds-checking in the shader, that is not applicable to DX and Vulkan and makes this API a lot less convenient to use.

Specifically because the thread-count in Metal is fixed in place through transpiling HLSL with `numthreads()` to Metallib; that value should be persisted in the pipeline and passed through.

lib/API/MTL/MTLDevice.cpp
```diff
@@ -205,6 +209,173 @@ llvm::Error MTLQueue::submit(
...
+
+class MTLComputeEncoder : public offloadtest::ComputeEncoder {
+  MTL::CommandBuffer *CmdBuffer;
+  MTL::ComputeCommandEncoder *ComputeEnc = nullptr;
+  MTL::BlitCommandEncoder *BlitEnc = nullptr;
+
+  /// Threadgroup size from shader reflection (the numthreads() attribute
+  /// persisted in the transpiled Metallib). Must be set via
+  /// setThreadGroupSize() before dispatching.
+  MTL::Size ThreadsPerGroup = {1, 1, 1};
```
manon-traverse, line 273: I would love it if this were just taken from the pipeline passed to dispatch, but for that we need another PR to land. I'm fine merging this as-is and fix it later.

manon-traverse (approved):

lib/API/MTL/MTLDevice.cpp
MarijnS95: Yeah, for now this is just a default-initializer and `executeProgram()` calls into `setThreadsPerGroup()` directly. Having the `Pipeline` PR would allow a new `bindPipeline()` (or receiving it as the first argument to `dispatch()`) to update or inline this variable entirely.

MarijnS95:
I'm rewriting this PR a bit to tear out parallel mode again; its barrier placement is less than ideal for something that isn't even used yet, significantly taking away from the intent of this PR: generalizing the entrypoints to adding commands to command buffers, on top of which I intend to add acceleration structure building.

EDIT: Pushed the original rebase as a restore-point, before pushing the rewrite.

lib/API/VK/Device.cpp
```diff
@@ -2635,6 +2792,13 @@ llvm::Expected<offloadtest::SubmitResult> VulkanQueue::submit(
   const uint64_t SignalValue = ++FenceCounter;
   const VkPipelineStageFlags WaitStage = VK_PIPELINE_STAGE_ALL_COMMANDS_BIT;
 
+  // Each command buffer defaults to src=HOST, which is only correct for
+  // standalone submissions.  Multi-CB batches would need inter-CB barriers.
+  if (CBs.size() > 1)
+    llvm::errs()
+        << "Warning: submitting multiple command buffers in a single batch; "
+           "encoder barriers do not account for inter-command-buffer "
+           "dependencies.\n";
```
MarijnS95, line 3053: This also relies on the caller waiting on the returned `SubmitResult` before submitting the next "batch".

manon-traverse (approved):

EmilioLaiso:
LGTM

PR #1101, opened 2026-04-17
Return `SubmitResult` from `Queue::submit()` instead of blocking

MarijnS95:
Change `Queue::submit()` to return `Expected<SubmitResult>` (a fence + signal value pair) instead of blocking until completion. Each backend (Vulkan, DirectX, Metal) now returns its queue fence and counter value, and callers explicitly wait via `SubmitResult::waitForCompletion()` before reading back results.

This allows command buffers submitted between the first and last submit to remain in-flight on the GPU without host-side synchronization.

- Vulkan: `cleanup()` waits on the queue fence for error-path safety
- DirectX: tile mapping sync reuses the queue's `SubmitFence`
- Metal: `MTLQueue` gains a `SharedEvent` fence; the last command buffer in each batch encodes a signal on it

### DIFF
--- a/lib/API/VK/Device.cpp
+++ b/lib/API/VK/Device.cpp
@@ -497,8 +497,8 @@ public:
       : Queue(Q), QueueFamilyIdx(QueueFamilyIdx), Device(Device),
         SubmitFence(std::move(SubmitFence)) {}
 
-  llvm::Error submit(
-      llvm::SmallVectorImpl<std::unique_ptr<offloadtest::CommandBuffer>> &&CBs)
+  llvm::Expected<offloadtest::SubmitResult>
+  submit(llvm::SmallVector<std::unique_ptr<offloadtest::CommandBuffer>> CBs)
       override;
 };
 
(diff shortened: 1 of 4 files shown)


lib/API/VK/Device.cpp
```diff
@@ -2576,12 +2602,13 @@ class VulkanDevice : public offloadtest::Device {
...
     if (IS.Pool)
       vkDestroyDescriptorPool(Device, IS.Pool, nullptr);
+    return llvm::Error::success();
   }
 
   llvm::Error executeProgram(Pipeline &P) override {
     InvocationState State;
     auto CleanupState = llvm::scope_exit([&]() {
-      cleanup(State);
+      llvm::consumeError(cleanup(State));
```
MarijnS95: Now cleanup didn't fully happen. I think `waitForCompletion()` should instead be wrapped in `llvm::consumeError()` then.

manon-traverse (approved):

EmilioLaiso:
LGTM

PR #1103, opened 2026-04-20
Rework workflow design to allow opting in to SplitBuild feature, and use Windows-compatible tar options

### LINKED ISSUE #1117: Abstractify OS, parameterize splitting build and test feature, and make workflow changes in-place for splitting build and test
This issue: https://github.com/llvm/offload-test-suite/issues/1081
says that we should essentially split the offload test suite workflows into two separate jobs, a build and test job.
However, the PR that implemented this made a separate workflow to do this, and made no room for parameterization on whether to split or not.
@bogner left a few comments addressing these architectural concerns.
We should change the workflow architecture to parameterize the split build / test feature, and not make the workflow so aware of the current OS, instead leaving that to the build jobs.
The comments are here: https://github.com/llvm/offload-test-suite/pull/1067#pullrequestreview-4143916972

bob80905:
--touch is not a supported option on the windows version of tar.
This PR changes tar to focus on extraction, and introduces a later step to update the timestamps after extraction in a cross-platform way, so that ninja doesn't rebuild the binaries.

Separately, there were other architectural / design concerns with the way the workflow was organized.
This PR parameterizes the SplitBuild option, so the current workflows can be manually dispatched to opt in to the split build run type.
But, this change will not affect current workflows, except just adding 2 skipped steps at the bottom, for uploading artifacts.
Fixes https://github.com/llvm/offload-test-suite/issues/1117


.github/workflows/build-and-test-split-callable.yaml
```diff
@@ -277,7 +277,14 @@ jobs:
...
             cd $GITHUB_WORKSPACE
-            tar xf $RUNNER_TEMP/build-artifacts.tar --touch
+            tar xf $RUNNER_TEMP/build-artifacts.tar
+      - name: Refresh artifact timestamps
+        shell: powershell
+        run: |
+          # Touch all extracted build outputs so ninja sees them as newer
+          # than source files and skips rebuilding.
+          Get-ChildItem -Recurse -File DXC\build\bin, llvm-project\build |
+            ForEach-Object { $_.LastWriteTime = Get-Date }
```
bogner: Can this really get away with being windows only?

bob80905: Yes, on line 230 we guard this code against macOS runs.
MacOS will not do any of this artifact download / upload stuff, since it is so fast it doesn't need this optimization, and I think this whole thing will be simpler if we exclude MacOS.

bogner: See my post-commit review here: https://github.com/llvm/offload-test-suite/pull/1067#pullrequestreview-4143916972 - I have some misgivings about these architectural choices.

.github/workflows/build-and-test-callable.yaml
```diff
@@ -169,6 +192,136 @@ jobs:
...
+        if: inputs.SplitBuild == true
+        shell: bash
+        run: |
+            cd $GITHUB_WORKSPACE
+            tar cf $RUNNER_TEMP/build-artifacts.tar \
+              --exclude='*.obj' \
+              --exclude='*.o' \
+              --exclude='*.ilk' \
+              --exclude='*.pdb' \
+              --exclude='CMakeFiles' \
```
bogner, line 247: Do we need all of these, or is `CMakeFiles` sufficient? AFAIK the object files all end up within the `CMakeFiles/SomeTarget.dir` directories.

bogner (approved):

.github/workflows/build-and-test-callable.yaml
bob80905: I can confirm that .pdbs should also be excluded, as they do exist in the bin directory of the runner. We don't want to zip up and upload any more than we need to, all we need is enough to execute the tests, to optimize zip/upload artifact speed and the unzipping step as well.
For example, right now these paths exists on the runner:
`"C:\actions-runner\_work\offload-test-suite\offload-test-suite\llvm-project\build\bin\api-query.pdb"`
`"C:\actions-runner\_work\offload-test-suite\offload-test-suite\llvm-project\build\bin\offloader.exe"`
This is not under the CMakeFiles dir, and we don't want the pdb. I believe the other file types will be useful to avoid in other OS cases as well. But we do want the .exe, and it is not under CMakeFiles.

alsepkow (commented):
Reviewed with @inbelic 
The linked issue is for the original work. This PR is to resolve some post PR completion comments. Can we get another issue opened with context and link against that?

PR #1104, opened 2026-04-21
Add a descriptor allocator to DX12 backend. Store handles to RTV and DSV and Texture.

manon-traverse:
A small step forward in how descriptors are managed.
This PR aims to store the RTV and DSV descriptors in centralized heaps on the device, using DX12 as intended by having all the descriptors in a shared heap.

In a follow-up step, this same change will also be made to `CBV_SRV_UAV` descriptors.

### DIFF
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -320,11 +320,6 @@ public:
   BufferCreateDesc Desc;
   size_t SizeInBytes;
 
-  // Staging descriptors, allocated from the device's central non-shader-visible
-  // CBV_SRV_UAV heap. A zero ptr means no descriptor.
-  D3D12_CPU_DESCRIPTOR_HANDLE SRVHandle = {};
-  D3D12_CPU_DESCRIPTOR_HANDLE UAVHandle = {};
-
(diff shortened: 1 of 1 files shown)


lib/API/DX/Device.cpp
```diff
@@ -290,6 +291,11 @@ class DXBuffer : public offloadtest::Buffer {
   BufferCreateDesc Desc;
   size_t SizeInBytes;
 
+  // Staging descriptors, allocated from the device's central non-shader-visible
+  // CBV_SRV_UAV heap. A zero ptr means no descriptor.
+  D3D12_CPU_DESCRIPTOR_HANDLE SRVHandle = {};
+  D3D12_CPU_DESCRIPTOR_HANDLE UAVHandle = {};
+
```
EmilioLaiso: Let's add these in the PR that makes use of them.

EmilioLaiso (commented):
LGTM, one small comment about adding unused fields.

lib/API/DX/Device.cpp
manon-traverse: Whoops that wasn't mean to make it into the PR. Thanks for spotting it.

MarijnS95 (approved):

PR #1113, opened 2026-04-23
Rewrite `Buffer` CPU-mapping API to lift vertex buffer creation into the common API layer

EmilioLaiso:
Lift vertex buffer creation into the common API layer and unifies it across the DX, Vulkan, and Metal backends.

### `Buffer` API
- Add pure-virtual `getSizeInBytes()`, `map()`, `unmap()` to `Buffer`. 
- Add `BufferUsage { Storage, VertexBuffer }` on `BufferCreateDesc`.
- Add `createVertexBufferFromCPUBuffer(Device&, const CPUBuffer&)`

### DirectX 
- Tighten the UAV flag logic: only `HEAP_TYPE_DEFAULT` gets `ALLOW_UNORDERED_ACCESS`

### DIFF
--- a/include/API/Buffer.h
+++ b/include/API/Buffer.h
@@ -19,8 +19,14 @@
 
 namespace offloadtest {
 
+enum class BufferUsage {
+  Storage,
+  VertexBuffer,
+};
+
 struct BufferCreateDesc {
   MemoryLocation Location;
+  BufferUsage Usage;
 };
 
 class Buffer {
@@ -28,6 +34,8 @@ class Buffer {
 
 public:
   virtual ~Buffer();
+  virtual size_t getSizeInBytes() const = 0;
+
   Buffer(const Buffer &) = delete;
   Buffer &operator=(const Buffer &) = delete;
 
--- a/include/API/Buffer.h
+++ b/include/API/Buffer.h
@@ -39,7 +39,7 @@ public:
 
   // Maps the buffer's memory for host access. Only valid for CpuToGpu and
   // GpuToCpu buffers; returns an error for GpuOnly. Each successful map() must
-  // be paired with a call to unmap().
+  // be paired with a call to unmap() before the buffer is used on the GPU.
   virtual llvm::Expected<void *> map() = 0;
   virtual llvm::Error unmap() = 0;
 
--- a/include/API/Buffer.h
+++ b/include/API/Buffer.h
@@ -16,6 +16,7 @@
 #include "API/Resources.h"
 
 #include "llvm/Support/Casting.h"
+#include "llvm/Support/Error.h"
 
 namespace offloadtest {
 
@@ -35,6 +36,12 @@ class Buffer {
 public:
   virtual ~Buffer();
   virtual size_t getSizeInBytes() const = 0;
+
+  // Maps the buffer's memory for host access. Only valid for CpuToGpu and
+  // GpuToCpu buffers; returns an error for GpuOnly. Each successful map() must
+  // be paired with a call to unmap().
+  virtual llvm::Expected<void *> map() = 0;
+  virtual llvm::Error unmap() = 0;
 
   Buffer(const Buffer &) = delete;
   Buffer &operator=(const Buffer &) = delete;
--- a/lib/API/MTL/MTLDevice.cpp
+++ b/lib/API/MTL/MTLDevice.cpp
@@ -159,6 +159,22 @@ public:
 
   size_t getSizeInBytes() const override { return SizeInBytes; }
 
+  llvm::Expected<void *> map() override {
+    if (Desc.Location == MemoryLocation::GpuOnly)
+      return llvm::createStringError(std::errc::invalid_argument,
+                                     "Cannot map a GpuOnly buffer.");
+    return Buf->contents();
+  }
+
+  llvm::Error unmap() override {
+    // Managed storage (CpuToGpu) requires an explicit didModifyRange to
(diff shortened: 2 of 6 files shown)


include/API/Buffer.h
```diff
@@ -16,18 +16,33 @@
...
 
 public:
   virtual ~Buffer();
+  virtual size_t getSizeInBytes() const = 0;
+
+  // Maps the buffer's memory for host access. Only valid for CpuToGpu and
+  // GpuToCpu buffers; returns an error for GpuOnly. Each successful map() must
+  // be paired with a call to unmap().
+  virtual llvm::Expected<void *> map() = 0;
+  virtual llvm::Error unmap() = 0;
```
manon-traverse, line 44: I think this is a fine change for now, as the framework already maps and unmaps buffers. However, the modern APIs allow for persistently mapped resources and I think we should embrace those as well. Although I think that's better for a follow-up PR once the abstraction layer is in a more complete state.

manon-traverse (approved):

include/API/Buffer.h
```diff
@@ -16,18 +16,33 @@
...
 class Buffer {
   GPUAPI API;
 
 public:
   virtual ~Buffer();
+  virtual size_t getSizeInBytes() const = 0;
+
+  // Maps the buffer's memory for host access. Only valid for CpuToGpu and
+  // GpuToCpu buffers; returns an error for GpuOnly. Each successful map() must
+  // be paired with a call to unmap().
```
MarijnS95: To match the limitation regarding flushing on Metal:

```suggestion
  // be paired with a call to unmap() before the buffer is used on the GPU.
```

lib/API/VK/Device.cpp
```diff
@@ -387,6 +387,34 @@ class VulkanBuffer : public offloadtest::Buffer {
...
+  llvm::Expected<void *> map() override {
+    if (Desc.Location == MemoryLocation::GpuOnly)
+      return llvm::createStringError(std::errc::invalid_argument,
+                                     "Cannot map a GpuOnly buffer.");
+    void *Ptr = nullptr;
+    if (vkMapMemory(Dev, Memory, 0, SizeInBytes, 0, &Ptr) != VK_SUCCESS)
+      return llvm::createStringError(std::errc::io_error,
+                                     "Failed to map buffer.");
+    // HOST_CACHED memory (GpuToCpu) needs explicit invalidation so the CPU
+    // sees the GPU-side writes.
```
MarijnS95: Specifically it's **not `HOST_COHERENT`** which requires manual flushing (and not having `HOST_CACHED` is [inherently host-coherent](https://docs.vulkan.org/refpages/latest/refpages/source/VkMemoryPropertyFlagBits.html)).

```suggestion
    // HOST_CACHED memory that is *not* HOST_COHERENT (GpuToCpu) needs explicit
    // invalidation so the CPU sees the GPU-side writes.
```

This PR doesn't use the `map()` function in any such codepaths yet though? Did you forget to update those, or this premature?

lib/API/MTL/MTLDevice.cpp
```diff
@@ -134,6 +134,24 @@ class MTLBuffer : public offloadtest::Buffer {
...
+                                     "Cannot map a GpuOnly buffer.");
+    return Buf->contents();
+  }
+
+  llvm::Error unmap() override {
+    // Managed storage (CpuToGpu) requires an explicit didModifyRange to
+    // propagate CPU-side writes to the GPU. Shared storage (GpuToCpu) is
+    // coherent and needs no action.
+    if (Desc.Location == MemoryLocation::CpuToGpu)
+      Buf->didModifyRange(NS::Range::Make(0, SizeInBytes));
```
MarijnS95, line 174: This is painful, it means `CpuToGpu` is not host coherent on Metal but it is mapped as such on Vulkan.

MarijnS95: I'm also curious if Metal forces/resets this managed mode request back to shared, because that's not available outside of Intel-based Macs: https://developer.apple.com/documentation/metal/mtlstoragemode/managed?language=objc#discussion

MarijnS95 (commented):
Not a fan -at all- of the `CpuToGpu` and related enums with UMA drastically increasing in popularity, but that's out of context for this PR even if affecting it significantly.

Secondly I don't think the title does this PR justice, you're factoring out the (somewhat unsafe) `(un)map()` APIs to be able to copy in and out of buffers on the CPU, and only as a small consequence of that are able to able to generalize vertex buffer creation which maps the buffer to write into it. This PR doesn't seem to update functions like `readBackData()` to use the new mapping API (and demonstrate its usefulness)?

lib/API/VK/Device.cpp
manon-traverse: Why can't we just force `HOST_COHERENT | HOST_VISIBLE | HOST_CACHED` for readback and just not have to deal with this?

MarijnS95: Because host coherent memory may be slower. However, the Vulkan spec requires at least one memory heap that is `HOST_VISIBLE | HOST_COHERENT`: https://docs.vulkan.org/refpages/latest/refpages/source/VkPhysicalDeviceMemoryProperties.html

If I remember correctly `gpu-allocator` relies on this, searching for a `CACHED` variant but falling back otherwise.

EmilioLaiso: > This PR doesn't use the map() function in any such codepaths yet though? Did you forget to update those, or this premature?

Didn't forget, just not in the scope of this PR

Issue #1114, opened 2026-04-23
[WARP] `WaveOps/WaveActiveAllEqual.fp64.test` fails on CPUs lacking AVX-512 support under WARP 1.0.19


Icohedron:
After PR #1111 updated the WARP version to 1.0.19, the test `WaveOps/WaveActiveAllEqual.fp64.test` has started failing specifically on the CBIENEMAN-INTEL machine when running the WARP+D3D12+DXC PR check workflow
https://github.com/llvm/offload-test-suite/actions/runs/24811643905/job/72617582293?pr=1111#step:13:537

This test failure does not reproduce on the AMD machine (https://github.com/llvm/offload-test-suite/actions/workflows/windows-amd-dxc-warp-d3d12.yaml), nor does it reproduce on the Qualcomm machine (https://github.com/llvm/offload-test-suite/actions/workflows/windows-qc-dxc-warp-d3d12.yaml).

I have also locally reproduced the test failure on my own Intel CPU devices, while my AMD CPU devices can not reproduce the test failure.

Edit: @bogner was able to reproduce this issue on AMD, so it appears this issue is not entirely Intel-specific.

Edit2: CPUs lacking AVX-512 appear to fail the test.

## Passing CPUs
| Feature | Threadripper 7970X | Ryzen 7 9700X |
|---|---|---|
| Architecture | Zen 4 | Zen 5 |
| Process | TSMC 5nm | TSMC 4nm |
| Cores / Threads | 32 / 64 | 8 / 16 |
| L3 Cache | 128 MB | 32 MB |
| ISA Extensions | AVX2, **AVX-512** | AVX2, **AVX-512** |

## Failing CPUs
| Feature | Threadripper 2990WX | Threadripper 3970X | i9-10900 | i5-14400F | Ultra 7 165H |
|---|---|---|---|---|---|
| Architecture | Zen+ | Zen 2 | Comet Lake | Raptor Lake | Meteor Lake |
| Process | GF 12nm | TSMC 7nm | Intel 14nm | Intel 7 | Intel 4 |
| Cores / Threads | 32 / 64 | 32 / 64 | 10 / 20 | 10 / 16 | 16 / 22 |
| L3 Cache | 64 MB | 128 MB | 20 MB | 20 MB | 18 MB |
| ISA Extensions | AVX2 | AVX2 | AVX2 | AVX2 | AVX2 |
| **AVX-512** | **No** | **No** | **No** | **No** | **No** |

[...]

PR #1116, opened 2026-04-23
Add IntelCPU feature flag, XFAIL WARP && IntelCPU in WaveActiveAllEqual.fp64

Icohedron:
This PR adds a new `IntelCPU` feature flag, and uses it to add an XFAIL for WARP && IntelCPU in the WaveActiveAllEqual.fp64 test (https://github.com/llvm/offload-test-suite/issues/1114)

Assisted-by: Claude Opus 4.6

### DIFF
--- a/test/WaveOps/WaveActiveAllEqual.fp64.test
+++ b/test/WaveOps/WaveActiveAllEqual.fp64.test
@@ -150,6 +150,9 @@ DescriptorSets:
 # Bug: https://github.com/llvm/offload-test-suite/issues/944
 # XFAIL: Intel && DirectX
 
+# Bug: https://github.com/llvm/offload-test-suite/issues/1114
+# XFAIL: WARP && IntelCPU
+
 # Bug: https://github.com/llvm/offload-test-suite/issues/981
 # XFAIL: Clang
 
--- a/test/lit.cfg.py
+++ b/test/lit.cfg.py
@@ -77,6 +77,31 @@ def setWaveSizeFeaturesDirectX(config, device):
         MinSizeInt *= 2
 
 
+def hostHasIntelCPU():
+    # Check if the host machine has an Intel CPU.
+    system = platform.system()
+    if system == "Windows":
+        return "intel" in platform.processor().lower()
+    elif system == "Linux":
+        try:
+            with open("/proc/cpuinfo", "r") as f:
+                for line in f:
+                    if line.startswith("vendor_id"):
+                        return "GenuineIntel" in line
+        except (IOError, OSError):
+            return False
+        return False
+    elif system == "Darwin":
+        try:
+            output = subprocess.check_output(
+                ["sysctl", "-n", "machdep.cpu.vendor"]
+            ).decode("UTF-8").strip()
+            return output == "GenuineIntel"
+        except (subprocess.CalledProcessError, OSError):
+            return False
+    return False
+
+
 def hostSupportsAVX512():
     # Check if the host CPU supports AVX-512 instructions.
     system = platform.system()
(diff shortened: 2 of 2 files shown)


test/WaveOps/WaveActiveAllEqual.fp64.test
```diff
@@ -150,6 +150,9 @@ DescriptorSets:
 # Bug: https://github.com/llvm/offload-test-suite/issues/944
 # XFAIL: Intel && DirectX
 
+# Bug: https://github.com/llvm/offload-test-suite/issues/1114
```
alsepkow, line 153: Would it be helpful to expand the comment here to say something like "The root cause is a bug in WARP. This will start passing again with a future WARP update." to help make it dead easy for someone to understand in the future that once this starts unexpectedly passing, they just need to delete these lines.

Icohedron: Isn't that how all the unexpected passes are currently fixed? Not sure why this one has to be written differently.

alsepkow: Maybe, I've never interacted with them. So, I was just thinking through what I thought the steps might be. 
Thought that maybe it would be helpful to prevent someone from needing to click the issue and get further context. Maybe not?

If we blindly turn it back on. Then yes, I don't see any value in adding more context to a comment.

Icohedron: The person removing the XFAIL is still required to check the issue link and update/close it, so it doesn't really make a difference.

bob80905 (approved):
This LGTM, and accomplishes the intent of the PR. Though I gotta say, I wonder if this is a good opportunity to do some abstraction and handle the other 3 CPU types. Maybe the function you wrote could handle getting the string, and then another function can return an enum code that indicates which CPU it is based on the returned string, or something along those lines. Though that might be a task for later.

Icohedron:
> This LGTM, and accomplishes the intent of the PR. Though I gotta say, I wonder if this is a good opportunity to do some abstraction and handle the other 3 CPU types. Maybe the function you wrote could handle getting the string, and then another function can return an enum code that indicates which CPU it is based on the returned string, or something along those lines. Though that might be a task for later.

Yea, a follow-up PR can add feature flags for other CPUs if needed.

test/lit.cfg.py
```diff
@@ -77,6 +77,33 @@ def setWaveSizeFeaturesDirectX(config, device):
...
+        try:
+            output = (
+                subprocess.check_output(["sysctl", "-n", "machdep.cpu.vendor"])
+                .decode("UTF-8")
+                .strip()
+            )
+            return output == "GenuineIntel"
+        except (subprocess.CalledProcessError, OSError):
+            return False
+    return False
```
bogner, line 104: Why don't we just use `platform.processor()` for everything here?

As is, it isn't very robust - do we really just want to return False if we're on a platform we haven't explicitly handled? For example, this would just be completely wrong on a FreeBSD machine on intel.

Additionally, the darwin handling prints `sysctl: unknown oid 'machdep.cpu.vendor'` every time this runs on an M-series mac. We should at least silence that, but is this even the right sysctl? I don't have an intel mac to check on. That said, this probably doesn't matter because I really don't see us ever running the test suite on an intel mac at this point.

bogner (commented):
I don't think we have the right condition for this XFAIL - it's now failing locally on my AMD machine

test/lit.cfg.py
Icohedron: Right. I'll make a follow-up PR to just use platform.processor() for all the platforms.

The code was trying to follow the same form as used in hostSupportsAVX512 but that's not necessary since we're only checking for processor name and not its feature set

Icohedron:
> I don't think we have the right condition for this XFAIL - it's now failing locally on my AMD machine

The CI AMD machine and my Threadripper are not encountering the issue, but the CI Intel machine, my Intel desktop, and my Intel laptop all encounter it. So I thought it was Intel-specific

bogner:
> > I don't think we have the right condition for this XFAIL - it's now failing locally on my AMD machine
> 
> The CI AMD machine and my Threadripper are not encountering the issue, but the CI Intel machine, my Intel desktop, and my Intel laptop all encounter it. So I thought it was Intel-specific

Seems like it must be something more subtle unfortunately

PR #1124, opened 2026-04-27
Make WaveActiveAllEqual.fp64 unsupported for WARP. Revert IntelCPU feature flag

Icohedron:
This PR reverts https://github.com/llvm/offload-test-suite/pull/1116 and instead marks the test `test/WaveOps/WaveActiveAllEqual.fp64.test` as unsupported under WARP due to the exact failure conditions for https://github.com/llvm/offload-test-suite/issues/1114 being unknown under WARP 1.0.19

### DIFF
--- a/test/WaveOps/WaveActiveAllEqual.fp64.test
+++ b/test/WaveOps/WaveActiveAllEqual.fp64.test
@@ -147,6 +147,9 @@ DescriptorSets:
 ...
 #--- end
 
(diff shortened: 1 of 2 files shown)


bogner (approved):
As we discussed offline, needing to use Unsupported here is a bit unfortunate, but until we better understand the problem it's our best option.

inbelic (approved):

PR #1125, opened 2026-04-27
XFAIL WARP && !AVX512 on `WaveActiveAllEqual.fp64`

Icohedron:
It appears that the test simply fails on CPUs lacking AVX-512 support. So the `UNSUPPORTED: WARP` can now be `XFAIL: WARP && !AVX512` with reasonable evidence.

I have updated #1114 and the internal bug report to reflect my findings.

### DIFF
--- a/test/WaveOps/WaveActiveAllEqual.fp64.test
+++ b/test/WaveOps/WaveActiveAllEqual.fp64.test
@@ -148,7 +148,7 @@ DescriptorSets:
 #--- end
 
 # Bug: https://github.com/llvm/offload-test-suite/issues/1114
-# XFAIL: WARP && !AVX512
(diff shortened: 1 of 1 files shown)


bob80905 (approved):
Nice find!

bogner (approved):
Probably best to leave the tables of passing/failing CPUs to the issue itself rather than the commit message - markdown tables aren't a great experience when looking through a git log.

Icohedron:
The XFAIL may or may not need to have `&& (x64 || x86)` appended to it. I will check once the Qualcomm WARP CI finishes to add it.

Edit: Yes, it is needed. The test is now an XPASS on Qualcomm https://github.com/llvm/offload-test-suite/actions/runs/25012901149/job/73253072963?pr=1125#step:13:812

PR #1126, opened 2026-04-28
Add DX Texture row pitch alignment for correct uploads and readbacks

EmilioLaiso:
When writing a test outputting to a 2x2 texture, I noticed that -on DX- the readback results contained wrong values. This happens because D3D12 spec requires `RowPitch` in a placed footprint to be a multiple of `D3D12_TEXTURE_DATA_PITCH_ALIGNMENT` (256 bytes).

### DIFF
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -144,8 +144,7 @@ static DXGI_FORMAT getRawDXFormat(const Resource &R) {
 // row size (Width * elementSize) is already a multiple of 256, this is a
 // no-op; for smaller rows it pads up.
 static uint32_t getAlignedTexturePitch(uint32_t Width, uint32_t ElementSize) {
-  return llvm::alignTo(Width * ElementSize,
-                       D3D12_TEXTURE_DATA_PITCH_ALIGNMENT);
+  return llvm::alignTo(Width * ElementSize, D3D12_TEXTURE_DATA_PITCH_ALIGNMENT);
 }
 
 static uint64_t getAlignedTextureBufferSize(const CPUBuffer &B) {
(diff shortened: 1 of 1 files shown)


manon-traverse (approved):

lib/API/DX/Device.cpp
```diff
@@ -138,6 +138,20 @@ static DXGI_FORMAT getRawDXFormat(const Resource &R) {
...
+// row size (Width * elementSize) is already a multiple of 256, this is a
+// no-op; for smaller rows it pads up.
+static uint32_t getAlignedTexturePitch(uint32_t Width, uint32_t ElementSize) {
+  return llvm::alignTo(Width * ElementSize, D3D12_TEXTURE_DATA_PITCH_ALIGNMENT);
+}
+
+static uint64_t getAlignedTextureBufferSize(const CPUBuffer &B) {
+  return uint64_t(B.OutputProps.Height) *
+         getAlignedTexturePitch(B.OutputProps.Width, B.getElementSize());
+}
```
MarijnS95, line 156: I remember complicated maths in our own backend, where we don't apply this alignment to the last line.

Hopefully that was "just" a minor memory-usage optimization, and not something that's required under the hood?

lib/API/DX/Device.cpp
```diff
@@ -1636,10 +1650,14 @@ class DXDevice : public offloadtest::Device {
...
-    // Create readback buffer sized for the pixel data (raw bytes).
+    // Create readback buffer sized for the pixel data with row pitch padded
+    // up to D3D12_TEXTURE_DATA_PITCH_ALIGNMENT, which is what D3D12 requires
+    // for the placed footprint used by CopyTextureRegion. The compaction
+    // back to a tight layout happens in readBack() via GetCopyableFootprints.
     BufferCreateDesc BufDesc = {};
     BufDesc.Location = MemoryLocation::GpuToCpu;
-    auto BufOrErr = createBuffer("RTReadback", BufDesc, OutBuf.size());
+    auto BufOrErr = createBuffer("RTReadback", BufDesc,
+                                 getAlignedTextureBufferSize(OutBuf));
```
MarijnS95, line 1663: Just for sanity-checking, this is effectively `TotalBytes` which `GetCopyableFootprints()` would have otherwise returned?

MarijnS95 (approved):

lib/API/DX/Device.cpp
EmilioLaiso: Dx12 only requires RowPitch to be 256-byte aligned. The padding bytes that round each row's data up to that pitch only exist so the next row starts at an aligned offset. The last row has no "next row," so its trailing padding is unused. I updated the function to reflect this (not padding last row)

lib/API/DX/Device.cpp
EmilioLaiso: See above comment, iirc `GetCopyableFootprints` returns the unpadded last row.

MarijnS95: Yeah, and we can use the unpadded row size too right? That's what the updated function returns after my request?

bogner (commented):
Can we get a test with this?

PR #1127, opened 2026-04-28
Align rasterizer state, readback paths, cull mode, triangle winding order across backends

EmilioLaiso:
## Summary

Make the three backends use the same image coordinate system, same culling mode and same `SV_IsFrontFace` polarity.

The framework had the following setup:

1. **Front-face polarity differed.** 
DX/MTL Y-flip clip→framebuffer at the viewport; Vulkan with positive-height viewport doesn't. 
Same HLSL triangle ended up with opposite framebuffer winding on VK vs DX/MTL, so `SV_IsFrontFace` gave opposite answers for the same input. This surfaced while writing a graphics-stage system-values test that needed `SV_IsFrontFace` to agree across backends.

2. **Visual PNG output was aligned by two compensating flips.**
DX and MTL each had a row-reversal on readback documented as "render target origin is top-left, while our image writer expects bottom-left".
The PNG writer reversed rows again on the way out.
VK had neither flip, but VK's no-Y-flip viewport meant its framebuffer was already "upside down" relative to DX/MTL, so its single PNG flip happened to produce a visually identical image. Two-flips-cancel on DX/MTL, no-flip-needed on VK, by accident. 
Adding any new behavior on top of this _(e.g., the negative-height-viewport change in the earlier stage of this PR that kicked off the investigation)_ immediately broke the cancellation and produced visually-flipped outputs.

## PR

Pick HLSL/DX semantics as canonical:
- Clip-space Y up
- Framebuffer Y down
- CCW = front 

Untangled the asymmetry by removing the readback flips on DX/MTL and the PNG writer's row reversal together with the VK viewport flip, so all three places change as one atomic step. After that, every backend produces the same Y-down memory layout end-to-end.

Exposed shared `copyFromTexture` function that all backends can call. 
If https://github.com/llvm/offload-test-suite/pull/1113 goes in main, which introduces `Map`/`Unmap` as functions on the `Buffer` type, the readback path can be generalized further.

## Golden Images
All graphics golden images still work as expected. The resulting Y-order is maintained.
The `Mandelbrot.test` image required updating as it's a compute path. Here's the PR: https://github.com/llvm/offload-golden-images/pull/6

### DIFF
--- a/include/Support/Pipeline.h
+++ b/include/Support/Pipeline.h
@@ -153,6 +153,19 @@ struct CPUBuffer {
       return Stride;
     return getSingleElementSize() * Channels;
   }
+
+  // The natural per-row byte size of this buffer when interpreted as a 2D
+  // image (no padding).
+  uint32_t getImageRowBytes() const {
+    return OutputProps.Width * getElementSize();
+  }
+
+  // Copy a 2D image readback from a GPU mapping into Data[0]. The host
+  // buffer is tightly packed with top-left origin. SrcRowPitch is the
+  // source's per-row stride in bytes; pass `getImageRowBytes()` when the
+  // source is tightly packed (Vulkan / Metal), or the GPU's reported pitch
+  // when the source has row padding (e.g., D3D12's 256-byte aligned rows).
+  void copyFromTexture(const void *Src, size_t SrcRowPitch);
 };
 
 struct Result {
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -934,7 +934,6 @@ public:
                                      "shader and a pixel shader.");
 
     PSODesc.RasterizerState = CD3DX12_RASTERIZER_DESC(D3D12_DEFAULT);
-    // Match the Vulkan and Metal backends: no culling, CCW = front-facing.
     PSODesc.RasterizerState.CullMode = D3D12_CULL_MODE_NONE;
     PSODesc.RasterizerState.FrontCounterClockwise = TRUE;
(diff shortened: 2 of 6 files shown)


manon-traverse:
I'm fine with either winding mode, but this feels like something we should add to the documentation.
In the future I can imagine we can specify this in the test, but we can then use what we pick now as a default.

manon-traverse (approved):

manon-traverse:
The CI is failing on a Mandelbrot test. @EmilioLaiso Could you check if it uses a golden image and if that is still correct? Perhaps there is still a bug in the texture readbacks?

include/Support/Pipeline.h
```diff
@@ -144,6 +144,19 @@ struct CPUBuffer {
...
+  uint32_t getImageRowBytes() const {
+    return OutputProps.Width * getElementSize();
+  }
+
+  // Copy a 2D image readback from a GPU mapping into Data[0]. The host
+  // buffer is tightly packed with top-left origin. SrcRowPitch is the
+  // source's per-row stride in bytes; pass `getImageRowBytes()` when the
+  // source is tightly packed (Vulkan / Metal), or the GPU's reported pitch
+  // when the source has row padding (e.g., D3D12's 256-byte aligned rows).
+  void copyFromTexture(const void *Src, size_t SrcRowPitch);
```
MarijnS95, line 168: Bit odd that `copyFromTexture()` takes a raw CPU pointer and not a `Texture` object, but it makes sense from the perspective that it's a mapped readback pointer that was the target of a `Texture`-copy operation. Not sure I have a better name in mind though, so let's leave it if you don't have anything better either.

(`copyFromReadbackTexture()`?)

MarijnS95 (approved):

PR #1142, opened 2026-04-30
Add `GraphicsSystemValues.test` for graphics-stage system-value semantics

EmilioLaiso:
Adds a new test that bundles every per-pixel, observable graphics-stage system-value semantic the framework can express today into one shader pass.
Modeled on the existing `ComputeSystemValues.test`. 
Covers `SV_VertexID`, `SV_PrimitiveID`, `SV_IsFrontFace`, and `SV_POSITION` pass-through.

----------

For this test to succeed on metal, the `drawPrimitives` call had to be replaced with `IRRuntimeDrawPrimitives`.
As per [the docs](https://developer.apple.com/metal/shader-converter/):

> Providing draw parameters
If a pipeline requires additional draw parameters, for example, when its vertex shader uses VertexID or InstanceID semantics, you need to provide these to Metal via a buffer.
> 
> Use Metal shader converter companion header automatically creates and provides these buffers to Metal when using its draw functions.

### DIFF
--- a/lib/API/MTL/MTLDevice.cpp
+++ b/lib/API/MTL/MTLDevice.cpp
@@ -1147,8 +1147,10 @@ class MTLDevice : public offloadtest::Device {
       // references buffer index 0.
       CmdEncoder->setVertexBuffer(llvm::cast<MTLBuffer>(*IS.VB).Buf, 0, 0);
 
-    CmdEncoder->drawPrimitives(MTL::PrimitiveTypeTriangle, NS::UInteger(0),
-                               P.getVertexCount());
+    // IRRuntimeDrawPrimitives also sets the DrawParams / DrawInfo argument
+    // buffers that metal-irconverter consults for SV_VertexID and friends.
+    IRRuntimeDrawPrimitives(CmdEncoder, MTL::PrimitiveTypeTriangle,
+                            NS::UInteger(0), P.getVertexCount());
 
     CmdEncoder->endEncoding();
 
--- a/test/Feature/Semantics/GraphicsSystemValues.test
+++ b/test/Feature/Semantics/GraphicsSystemValues.test
@@ -1,0 +1,164 @@
+#--- vertex.hlsl
+struct VSInput {
+    float4 pos : POSITION;
+    uint   vid : SV_VertexID;
+};
+
+struct VSOutput {
+    float4 position : SV_POSITION;
+    nointerpolation uint vid : VID;
+};
+
+VSOutput main(VSInput input) {
+    VSOutput o;
+    o.position = input.pos;
+    o.vid = input.vid;
+    return o;
+}
+
+#--- pixel.hlsl
+struct PSInput {
+    float4 position : SV_POSITION;
+    nointerpolation uint vid : VID;
+};
+
+struct Record {
+    uint  VertexID;
+    uint  PrimitiveID;
+    uint  IsFrontFace;
+    float PosX;
+    float PosY;
+    float PosZ;
+    float PosW;
+};
+
+RWStructuredBuffer<Record> Output : register(u0);
+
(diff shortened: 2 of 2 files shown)


manon-traverse:
I like the things we are testing, I am just wondering if it is better from a compiler developer's point of view if we have a test for each one of these or have more elements mixed in one test like we did here.

EmilioLaiso:
> I like the things we are testing, I am just wondering if it is better from a compiler developer's point of view if we have a test for each one of these or have more elements mixed in one test like we did here.

Considering `ComputeSystemValues.test` exists, I think it's ok for this pass to also test multiple semantics at the same time. 
`SV_POSITION` is basically tested in every other graphics test as well, so it's really only testing `SV_VertexID`, `SV_PrimitiveID` and `SV_IsFrontFace`. I don't see the point of adding 3 identical tests for some very fundamental semantics.

manon-traverse (approved):

test/Feature/Semantics/GraphicsSystemValues.test
```diff
@@ -0,0 +1,171 @@
...
+#
+# SV_VertexID: https://github.com/llvm/wg-hlsl/issues/151
+# SV_PrimitiveID: https://github.com/llvm/wg-hlsl/issues/160
+# SV_IsFrontFace: https://github.com/llvm/wg-hlsl/issues/162
+# SV_POSITION: https://github.com/llvm/wg-hlsl/issues/141
+# Clang's HLSL -> DXIL lowering does not yet implement these graphics-stage SVs.
+#
+# Also indirectly exercises the `nointerpolation` modifier, which Clang's HLSL
+# frontend does not yet parse.
+# https://github.com/llvm/wg-hlsl/issues/303
```
bogner: Would it be better to put these test explanations at the top of the file so that it's the first thing you see when opening the file?

bogner (approved):

test/Feature/Semantics/GraphicsSystemValues.test
EmilioLaiso: Done ✅

manon-traverse (approved):

lib/API/MTL/MTLDevice.cpp
```diff
@@ -1147,8 +1147,10 @@ class MTLDevice : public offloadtest::Device {
       // references buffer index 0.
       CmdEncoder->setVertexBuffer(llvm::cast<MTLBuffer>(*IS.VB).Buf, 0, 0);
```
MarijnS95, line 1148: Just double-checking that this `0` index doesn't overlap with anything `IRRuntimeDrawPrimitives()` is doing/using, [the docs](https://developer.apple.com/metal/shader-converter/) have this to say about it (even if it seems to work anyway):

> When using Metal vertex fetch, your stage-in attributes needs to start at index `kIRStageInAttributeStartIndex` (set to value `11`) defined in the `metal_irconverter_runtime.h` **and the buffer bind point for your vertex buffers start at `kIRVertexBufferBindPoint` (set to value `6`)**.

I believe that is because `kIRDescriptorHeapBindPoint = 0` reserves that slot already, but it's not currently being used the vertex stage.

But the code-comment is at least accurate in that we call `setBufferIndex(0)` in the `MTL::VertexAttributeDescriptor`.

MarijnS95 (approved):
Approving the Metal fix, but found an existing drive-by issue in the vertex-shader bindings not using the right indices/offsets defined by Metal Shader Converter. I'll address those in a follow-up PR.

lib/API/MTL/MTLDevice.cpp
MarijnS95: Fixed in https://github.com/llvm/offload-test-suite/pull/1207.

PR #1143, opened 2026-04-30
Move `Queue::submit` function definition out of anonymous namespace.

manon-traverse:
On my local machine I was getting this warning being treated as an error:
```
error: place definition of method 'submit' outside of an anonymous namespace [llvm-prefer-static-over-anonymous-namespace,-warnings-as-errors]
  502 | llvm::Error DXQueue::submit(
      |                      ^
```

This is easily resolved by moving the function definition outside of the anonymous namespace.

### DIFF
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -499,43 +499,6 @@ private:
   DXCommandBuffer() : CommandBuffer(GPUAPI::DirectX) {}
 };
 
-llvm::Error DXQueue::submit(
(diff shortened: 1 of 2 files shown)


MarijnS95 (approved):
Heh, this also means that every other abstraction in this anonymous namespace is also unreachable once we start accessing the generic API from a new generic TU 😅

MarijnS95:
Let's also check Metal :sweat_smile:

EmilioLaiso (commented):
LGTM

PR #1148, opened 2026-05-04
Improve validation of shader pipeline type.

manon-traverse:
There are some restrictions on what shaders are allowed to go together in a single shader pipeline.
This change defines two kinds: `Compute`, and `TraditionalRaster`. In the future we can add support for other like `MeshShaderRaster`, `RayTracing`, and `WorkGraph`.

By validating once if the provided shaders are a valid pipeline and assigning a pipeline type when we load from the YAML file, we can keep the logic in `executeProgram` simple and won't have to do any more expensive validation there.

### DIFF
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -2068,7 +2068,7 @@ public:
         return Err;
       llvm::outs() << "Compute command list created.\n";
 
-    } else {
+    } else if (P.isTraditionalRaster()) {
       // Create render target, depth/stencil, readback and vertex buffer and
       // PSO.
       if (auto Err = createRenderTarget(P, State))
@@ -2127,6 +2127,9 @@ public:
       if (auto Err = createGraphicsCommands(P, State))
         return Err;
       llvm::outs() << "Graphics command list created complete.\n";
+    } else {
+      return llvm::createStringError(
+          "Pipeline was neither Compute nor Traditional Graphics");
     }
 
     auto SubmitResult = executeCommandList(State);
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -2129,7 +2129,7 @@ public:
       llvm::outs() << "Graphics command list created complete.\n";
     } else {
       return llvm::createStringError(
-          "Pipeline was neither Compute nor Traditional Graphics");
+          "Pipeline was neither Compute nor Traditional Raster");
     }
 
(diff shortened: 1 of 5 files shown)


llvm-beanz (approved):

bogner (approved):

lib/API/DX/Device.cpp
```diff
@@ -2115,6 +2093,9 @@ class DXDevice : public offloadtest::Device {
       if (auto Err = createGraphicsCommands(P, State))
         return Err;
       llvm::outs() << "Graphics command list created complete.\n";
+    } else {
+      return llvm::createStringError(
+          "Pipeline was neither Compute nor Traditional Graphics");
```
MarijnS95:
```suggestion
          "Pipeline was neither Compute nor Traditional Raster");
```

😅

lib/API/VK/Device.cpp
```diff
@@ -2816,6 +2816,9 @@ class VulkanDevice : public offloadtest::Device {
       if (auto Err = createFrameBuffer(State))
         return Err;
       llvm::outs() << "Frame buffer created.\n";
+    } else {
+      return llvm::createStringError(
+          "Pipeline was neither Compute nor Traditional Graphics");
```
MarijnS95:
```suggestion
          "Pipeline was neither Compute nor Traditional Raster");
```

lib/Support/Pipeline.cpp
```diff
@@ -554,3 +557,35 @@ void MappingTraits<offloadtest::SpecializationConstant>::mapping(
...
+      return llvm::createStringError(
+          "Compute Pipeline is only allowed to have Compute Shader.");
+    Kind = ShaderPipelineKind::Compute;
+    return llvm::Error::success();
+  }
+
+  if (HasShaderType[llvm::to_underlying(Stages::Vertex)]) {
+    Kind = ShaderPipelineKind::TraditionalRaster;
+    return llvm::Error::success();
+  }
```
MarijnS95, line 585: How about having a pixel shader without a vertex shader? Or vice-versa?

MarijnS95 (approved):

lib/Support/Pipeline.cpp
manon-traverse: Pixel shader without a vertex shader is invalid. Vertex shader without pixel shader is valid.

The case without a pixel shader is used for depth-only rendering.

PR #1149, opened 2026-05-04
Unpack LLVM Distribution, and setup lit to run tests for specific test suite

### LINKED ISSUE #1150: Split build test feature cannot find ninja.exe on test runner machine
The test runner machine cannot locate ninja.exe when downloading the cmake files from the build machine.
The build machine's path to ninja.exe is machine specific, so the test runner machine cannot use the path to find ninja.exe.
We need to rewrite / update the test machine's path to ninja to where it actually is, and take all precautions to ensure that cmake can run correctly.

bob80905:
We are still encountering errors when configuring cmake with ninja, as ninja cannot be found:
https://github.com/llvm/offload-test-suite/actions/runs/25338698984/job/74291993540#step:13:155
Though the path is correct on the build machine's environment, it no longer holds in the test machine's (a different machine's) environment.
@Icohedron recommended a completely different approach, which involves installing an llvm distribution to the test machine. This avoids needing to replace exe paths in CMakeCache.txt files, and avoids dealing with CMake entirely. 
This PR sets up the infrastructure for installing the LLVM offload distribution (with all relevant files) without needing ninja or cmake, and just running lit tests on the appropriate suite.

Assisted by Github Copilot.
Fixes https://github.com/llvm/offload-test-suite/issues/1150


Icohedron:
As discussed offline, I do not think this is the correct approach. The CMakeCache file is very specific to the machine and is not portable to other machines. Simply patching the CMakeCache file is not a robust solution.

Instead, I think the better approach is to define an [LLVM distribution](https://llvm.org/docs/BuildingADistribution.html) that builds and installs the offloader, clang, and other dependencies into a self-contained and portable package. This package can then be distributed and ran on other machines without any additional building required.

Similarly for the offload tests, an install target can be defined that copies over all the lit tests into its own install prefix (which may or may not be the same as the install prefix for the LLVM distribution). 

The main issue to solve would be getting lit to find all the dependencies required to run the tests. A solution for this would be to include a Python script that scans for the required dependencies and uses the information to generate a lit.site.cfg.py to satisfy the requirements for running lit tests.

.github/workflows/build-and-test-callable.yaml
```diff
@@ -190,9 +190,9 @@ jobs:
...
+            cmake -G Ninja ${{ inputs.LLVM-ExtraCMakeArgs }} -DCMAKE_BUILD_TYPE=${{ inputs.BuildType }} -DLLVM_ENABLE_ASSERTIONS=On -C ${{ github.workspace }}/llvm-project/clang/cmake/caches/HLSL.cmake -C ${{ github.workspace }}/OffloadTest/cmake/caches/sccache.cmake -DDXC_DIR=${{ github.workspace }}/DXC/build/bin -DLLVM_EXTERNAL_OFFLOADTEST_SOURCE_DIR=${{ github.workspace }}/OffloadTest -DLLVM_EXTERNAL_PROJECTS="OffloadTest" -DLLVM_LIT_ARGS="--xunit-xml-output=testresults.xunit.xml -v" -DOFFLOADTEST_TEST_CLANG=${{steps.Test-Clang.outputs.TEST_CLANG || 'Off' }} -DGOLDENIMAGE_DIR=${{ github.workspace }}/golden-images -DHLSL_ENABLE_OFFLOAD_DISTRIBUTION=${{ inputs.SplitBuild == true && 'On' || 'Off' }} -DCMAKE_INSTALL_PREFIX=${{ github.workspace }}/install ${{ github.workspace }}/llvm-project/llvm/
             ninja hlsl-test-depends
-      - name: Dump GPU Info
+      - name: Dump GPU Info (build runner)
```
bob80905, line 204: Note, though it doesn't seem to make sense long term to have a dump gpu info step on the build runner machine, it does make sense to have it short term, since PRs will continue to run with splitbuild disabled while I test and validate that everything is working correctly.

utils/configure-test-suite.py
```diff
@@ -0,0 +1,276 @@
...
+import platform
+import shutil
+import sys
+
+# Suite name -> (d3d12, vk, mtl, clang, warp, description).
+SUITES = {
+    "d3d12": (True, False, False, False, False, "DXC-compiled shaders on DirectX 12"),
+    "vk": (False, True, False, False, False, "DXC-compiled shaders on Vulkan"),
+    "mtl": (False, False, True, False, False, "DXC-compiled shaders on Metal"),
+    "clang-d3d12": (
```
bob80905, line 37: This looks different because darker formatted it like this.

.github/workflows/build-and-test-callable.yaml
```diff
@@ -232,19 +232,28 @@ jobs:
...
+      - name: Install distribution
+        if: inputs.SplitBuild == true
+        shell: bash
+        run: |
+            set -euxo pipefail
+            cd $GITHUB_WORKSPACE/llvm-project/build
+            ninja install-distribution install-offload-tools install-offload-test-suite
+            # DXC isn't part of the LLVM distribution; install it into the
+            # same prefix so non-clang suites still find dxc on PATH.
+            cmake --install $GITHUB_WORKSPACE/DXC/build --prefix $GITHUB_WORKSPACE/install
```
Icohedron: Installing DXC into the same prefix as the LLVM distribution might cause problems due to mixing of headers, libraries, and potentially-conflicting binary names. DXC should be installed into a separate prefix and packaged into its own artifact.

.github/workflows/build-and-test-callable.yaml
```diff
@@ -190,9 +190,9 @@ jobs:
...
+            cmake -G Ninja ${{ inputs.LLVM-ExtraCMakeArgs }} -DCMAKE_BUILD_TYPE=${{ inputs.BuildType }} -DLLVM_ENABLE_ASSERTIONS=On -C ${{ github.workspace }}/llvm-project/clang/cmake/caches/HLSL.cmake -C ${{ github.workspace }}/OffloadTest/cmake/caches/sccache.cmake -DDXC_DIR=${{ github.workspace }}/DXC/build/bin -DLLVM_EXTERNAL_OFFLOADTEST_SOURCE_DIR=${{ github.workspace }}/OffloadTest -DLLVM_EXTERNAL_PROJECTS="OffloadTest" -DLLVM_LIT_ARGS="--xunit-xml-output=testresults.xunit.xml -v" -DOFFLOADTEST_TEST_CLANG=${{steps.Test-Clang.outputs.TEST_CLANG || 'Off' }} -DGOLDENIMAGE_DIR=${{ github.workspace }}/golden-images -DHLSL_ENABLE_OFFLOAD_DISTRIBUTION=${{ inputs.SplitBuild == true && 'On' || 'Off' }} -DCMAKE_INSTALL_PREFIX=${{ github.workspace }}/install ${{ github.workspace }}/llvm-project/llvm/
```
Icohedron: Note to reviewers: `HLSL_ENABLE_OFFLOAD_DISTRIBUTION` is introduced in PR https://github.com/llvm/llvm-project/pull/196640

test/CMakeLists.txt
```diff
@@ -191,3 +191,55 @@ endif()
...
+#
+# Install the test sources, the canonical lit.site.cfg.py.in template, and
+# the configure-test-suite.py helper into a portable install prefix. After
+# installing, the prefix can be tarred up and consumed on a different
+# machine without a CMake or compiler toolchain. The consumer runs:
+#
+#   cd <prefix>/share/hlsl-test-suite
+#   ./configure-test-suite.py --suite clang-d3d12
+#   lit run/test/clang-d3d12
+#
```
Icohedron: These comments would be better left to the doc in the docs folder.
```suggestion
```

.github/workflows/build-and-test-callable.yaml
Icohedron: Golden images should always be checked out because it gets included in the `install-offload-test-suite` install target.
```suggestion
```

utils/configure-test-suite.py
Icohedron, line 1: I think we should do a similar thing as clang/llvm and put scripts like this into its own `utils` folder.

utils/configure-test-suite.py
```diff
@@ -0,0 +1,276 @@
...
+        if found:
+            dxc_path = pathlib.Path(found).resolve()
+    if needs_dxc and (dxc_path is None or not dxc_path.is_file()):
+        print(
+            "warning: DXC not found; non-clang suites will not be able to compile shaders. "
+            "Pass --dxc-path to silence this warning.",
+            file=sys.stderr,
+        )
+        dxc_path = pathlib.Path("")
+    dxc_dir = str(dxc_path.parent) if dxc_path and str(dxc_path) else ""
```
Icohedron: `str(pathlib.Path("")` resolves to `True`. I think you want this instead:
```suggestion
        dxc_path = None
    dxc_dir = str(dxc_path.parent) if dxc_path is not None else ""
```

utils/configure-test-suite.py
```diff
@@ -0,0 +1,276 @@
...
+        d3d12, vk, mtl, clang, warp, _ = SUITES[suite]
+        suite_obj_root = output_base
+        suite_dir = output_base / "test" / suite
+        suite_dir.mkdir(parents=True, exist_ok=True)
+
+        substitutions = {
+            "OFFLOADTEST_BINARY_DIR": str(suite_obj_root),
+            "OFFLOADTEST_SOURCE_DIR": str(test_src_root.parent),
+            "LLVM_TOOLS_DIR": str(bin_dir),
+            "DXC_EXECUTABLE": str(dxc_path) if dxc_path else "",
```
Icohedron:
```suggestion
            "DXC_EXECUTABLE": str(dxc_path) if dxc_path is not None else "",
```

utils/configure-test-suite.py
```diff
@@ -0,0 +1,276 @@
...
+
+    for suite in args.suites:
+        d3d12, vk, mtl, clang, warp, _ = SUITES[suite]
+        suite_obj_root = output_base
+        suite_dir = output_base / "test" / suite
+        suite_dir.mkdir(parents=True, exist_ok=True)
+
+        substitutions = {
+            "OFFLOADTEST_BINARY_DIR": str(suite_obj_root),
+            "OFFLOADTEST_SOURCE_DIR": str(test_src_root.parent),
```
Icohedron: I think `OFFLOADTEST_SOURCE_DIR` is a bit misleading. The name suggests it points to the offload test suite project root directory, but it's actually pointing to the install prefix.
```suggestion
            "OFFLOADTEST_INSTALL_DIR": str(test_src_root.parent),
```

utils/configure-test-suite.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+def emit_site_cfg(template_text, substitutions):
+    """Replace @VAR@ tokens in template_text. Unsubstituted @VAR@ tokens
+    produce a hard error so we don't silently ship a half-configured file."""
+    text = template_text.replace("@LIT_SITE_CFG_IN_HEADER@", HEADER)
+    for key, value in substitutions.items():
+        text = text.replace("@{}@".format(key), value)
+    # Catch any leftover @TOKEN@ that we didn't account for.
+    import re
```
Icohedron: Perhaps this import should be at the top level instead of inside a function scope?

.github/workflows/build-and-test-callable.yaml
```diff
@@ -320,24 +339,66 @@ jobs:
...
+          pip-install: lit pyyaml
+      - name: Ensure lit is installed
+        shell: bash
         run: |
-            cd llvm-project
-            cd build
-            ./bin/api-query
+            set -euxo pipefail
+            python -m pip install --upgrade pip
+            python -m pip install lit pyyaml
```
Icohedron: It looks like lit and pyyaml are installed twice: once by `pip-install: lit pyyaml` and again by `python -m pip install lit pyyaml`

bob80905: Yep, this was an artifact of testing act locally. I'll remove it.

utils/configure-test-suite.py
bob80905: There are 2 different uses of this variable, and Install Dir isn't the only use, so I renamed this var to be a bit more abstract / generic.

.github/workflows/build-and-test-callable.yaml
```diff
@@ -232,26 +241,59 @@ jobs:
...
+            # instead we just stage the runtime binaries the tests need.
+            mkdir -p $GITHUB_WORKSPACE/install-dxc/bin
+            # Stage the dxc runtime binaries (Windows: dxc.exe + *.dll;
+            # Linux: dxc + libdxcompiler.so + libdxil.so). cd into the
+            # source dir so glob expansion works regardless of how
+            # $GITHUB_WORKSPACE is quoted.
+            cd $GITHUB_WORKSPACE/DXC/build/bin
+            cp dxc* $GITHUB_WORKSPACE/install-dxc/bin/
+            cp *dxcompiler* $GITHUB_WORKSPACE/install-dxc/bin/ 2>/dev/null || true
+            cp *dxil* $GITHUB_WORKSPACE/install-dxc/bin/ 2>/dev/null || true
```
Icohedron: I don't think this is the proper way to produce a DXC package; it's not of the correct folder structure.
Have a bin folder with dxc and dxv, and a lib folder with the required libs.

Icohedron: Run the DXC install target.
The install target will have a bunch of unnecessary LLVM stuff in it.
Just move/copy dxc, dxv, the required libs into a separate output directory with `bin` and `lib` folders, and then package the output directory into an artifact to be distributed to the runners.

.github/workflows/build-and-test-callable.yaml
```diff
@@ -260,53 +316,54 @@ jobs:
...
+                # in the cwd or system dirs (e.g. shipped by other SDKs)
+                # would win. Copy the DLLs next to dxc.exe so they're
+                # found via the app-directory search (#1).
+                cp "$libdir"/*.dll "$bindir"/ ;;
+              Linux)
+                echo "LD_LIBRARY_PATH=$libdir:${LD_LIBRARY_PATH:-}" >> "$GITHUB_ENV" ;;
+              macOS)
+                echo "DYLD_LIBRARY_PATH=$libdir:${DYLD_LIBRARY_PATH:-}" >> "$GITHUB_ENV" ;;
+              *)
+                echo "Unsupported RUNNER_OS=$RUNNER_OS" >&2; exit 1 ;;
```
Icohedron: This shouldn't be necessary. 
On Windows, the DLLs are in the same directory as the binaries.
On Linux and Mac, `../lib` should already be searched when executing the binaries because the binaries set the RUNPATH to do so.

.github/workflows/build-and-test-callable.yaml
```diff
@@ -232,26 +241,73 @@ jobs:
...
+        run: |
+            set -euxo pipefail
+            cd $GITHUB_WORKSPACE/llvm-project/build
+            ninja install-distribution install-offload-tools install-offload-test-suite
+            # Stage DXC into a portable bin/+lib/ tree at dxc-dist.
+            # See docs/offload-distribution.md ("DXC prefix") for why
+            # we use per-component install targets and copy dxv/dxil
+            # straight from the build directory.
+            cd $GITHUB_WORKSPACE/DXC/build
+            ninja install-dxc install-dxcompiler
```
Icohedron: Ok. I didn't expect `ninja install` to actually not work, and `install-dxc install-dxcompiler` targets are also incomplete for what we need.

In this case I'm ok with not using any install targets at all and simply copying the binaries and libraries out of the build folder.

The artifact on Windows should have the contents:
```
bin/dxc.exe
bin/dxcompiler.dll
bin/dxil.dll
bin/dxv.exe
lib/dxcompiler.lib
lib/dxil.lib
```

.github/workflows/build-and-test-callable.yaml
```diff
@@ -232,26 +241,73 @@ jobs:
...
+            distlib=$GITHUB_WORKSPACE/dxc-dist/lib
+            mkdir -p "$distbin" "$distlib"
+            case "$RUNNER_OS" in
+              Windows)
+                cp "$buildbin/dxc.exe"          "$distbin/"
+                cp "$buildbin/dxv.exe"          "$distbin/"
+                cp "$buildbin/dxcompiler.dll"   "$distbin/"
+                cp "$buildbin/dxil.dll"         "$distbin/"
+                cp "$buildlib/dxcompiler.lib"   "$distlib/"
+                cp "$buildlib/dxil.lib"         "$distlib/" ;;
```
Icohedron, line 283: I think we should include .pdb files if they are present

.github/workflows/build-and-test-callable.yaml
```diff
@@ -232,26 +241,77 @@ jobs:
...
+            case "$RUNNER_OS" in
+              Windows)
+                cp "$buildbin/dxc.exe"          "$distbin/"
+                cp "$buildbin/dxv.exe"          "$distbin/"
+                cp "$buildbin/dxcompiler.dll"   "$distbin/"
+                cp "$buildbin/dxil.dll"         "$distbin/"
+                cp "$buildbin/dxc.pdb"          "$distbin/"
+                cp "$buildbin/dxv.pdb"          "$distbin/"
+                cp "$buildbin/dxcompiler.pdb"   "$distbin/"
+                cp "$buildbin/dxil.pdb"         "$distbin/"
```
Icohedron: Does this step fail if you set the build type to Release and the .pdb files are not present as a result?

bob80905: It would fail if the .pdbs were not present, but there are no callers that build with Release at the moment.

Icohedron: One of the inputs to the workflow is `BuildType` which is RelWithDebInfo by default, but it can be set to Release and affects the type of the build for DXC

bob80905: Alright, I'll make this a bit more defensive in case someone wants to run with Release.

Icohedron (approved):

CMakeLists.txt
```diff
@@ -132,3 +139,12 @@ add_subdirectory(tools)
...
 add_subdirectory(unittests)
 add_subdirectory(test)
+
+# Aggregate install target for the offload-tools component. After the
+# `add_subdirectory(tools)` call above has registered every tool under the
+# `offload-tools` component via `add_offloadtest_tool`, this provides a
+# single `install-offload-tools` target that builds + installs all of them.
+if (NOT LLVM_ENABLE_IDE)
+  add_llvm_install_targets(install-offload-tools
+    COMPONENT offload-tools)
```
Icohedron: The `install-offload-tools` target doesn't actually build the tools before installing. Add a DEPENDS to ensure they are built before installing, otherwise the install-offload-tools will fail.
```suggestion
    COMPONENT offload-tools
    DEPENDS api-query imgdiff offloader)
```

Icohedron (changes requested):

Icohedron (approved):

CMakeLists.txt
```diff
@@ -132,3 +139,13 @@ add_subdirectory(tools)
...
 add_subdirectory(test)
+
+# Aggregate install target for the offload-tools component. After the
+# `add_subdirectory(tools)` call above has registered every tool under the
+# `offload-tools` component via `add_offloadtest_tool`, this provides a
+# single `install-offload-tools` target that builds + installs all of them.
+if (NOT LLVM_ENABLE_IDE)
+  add_llvm_install_targets(install-offload-tools
+    COMPONENT offload-tools
+    DEPENDS api-query imgdiff offloader)
```
farzonl, line 150: is this the right way to do this in Cmake? don't you use the env names? While unlikely if you use the cmake  definition for these executable names instead of the executable name it allows us to change the name of the executable in only one place instead of updated across multiple cmake files.

.github/workflows/build-and-test-callable.yaml
```diff
@@ -232,26 +241,79 @@ jobs:
...
+            buildlib=$GITHUB_WORKSPACE/DXC/build/lib
+            distbin=$GITHUB_WORKSPACE/dxc-dist/bin
+            distlib=$GITHUB_WORKSPACE/dxc-dist/lib
+            mkdir -p "$distbin" "$distlib"
+            case "$RUNNER_OS" in
+              Windows)
+                cp "$buildbin/dxc.exe"          "$distbin/"
+                cp "$buildbin/dxv.exe"          "$distbin/"
+                cp "$buildbin/dxcompiler.dll"   "$distbin/"
+                cp "$buildbin/dxil.dll"         "$distbin/"
```
farzonl, line 271: I'm curious why we copy here instead of using `cmake --install` I do think we have a special `install-dxcompiler` flag.

.github/workflows/build-and-test-callable.yaml
```diff
@@ -320,24 +359,60 @@ jobs:
...
+            echo "Will run lit suite: $suite"
+      - name: Configure standalone test suite
+        shell: bash
+        run: |
+            set -euxo pipefail
+            cd $GITHUB_WORKSPACE/install/share/hlsl-test-suite
+            # Locate dxc in the separate DXC install prefix. .exe on Windows,
+            # bare name elsewhere.
+            dxc_path=$GITHUB_WORKSPACE/dxc-dist/bin/dxc
+            if [ -x "${dxc_path}.exe" ]; then dxc_path="${dxc_path}.exe"; fi
```
farzonl, line 402: hmm is dxc_path sanitized. I've had issues with paths with bash on windows. Concatenation is highly vulnerable to escaping errors on Windows and we are concating `GITHUB_WORKSPACE` which makes me worried.

.github/workflows/build-and-test-callable.yaml
```diff
@@ -232,26 +241,79 @@ jobs:
...
+                cp "$buildbin/dxc.exe"          "$distbin/"
+                cp "$buildbin/dxv.exe"          "$distbin/"
+                cp "$buildbin/dxcompiler.dll"   "$distbin/"
+                cp "$buildbin/dxil.dll"         "$distbin/"
+                # Copy PDBs if they exist. If DXC is built with the Release config,
+                # then the PDBs won't exist, so we silently skip copying them.
+                cp "$buildbin/dxc.pdb"          "$distbin/" 2>/dev/null || true
+                cp "$buildbin/dxv.pdb"          "$distbin/" 2>/dev/null || true
+                cp "$buildbin/dxcompiler.pdb"   "$distbin/" 2>/dev/null || true
+                cp "$buildbin/dxil.pdb"         "$distbin/" 2>/dev/null || true
```
farzonl: instead of silently failing for release builds could we just know if we were a debug or release before we do the copy?

farzonl (approved):
Overall LGTM. Minor issues.

CMakeLists.txt
bob80905: I don't think the convention is to define tool targets as variables. E.g., opt, llc, llvm-as, etc are referenced literally, so I think this is the right convention.

.github/workflows/build-and-test-callable.yaml
bob80905: cmake --install / ninja install doesn't work here: it walks every cmake_install.cmake in the build 
tree, including LLVM tools (e.g. llvm-as) that aren't built by our default ninja invocation, and aborts on the first
missing artifact.
Deric pointed this out too, and I documented why this is done instead in the docs file.

PR #1153, opened 2026-05-05
Move dispatch dimensions and vertex count to new `DispatchParameters` with defaults

### LINKED ISSUE #1152: Move `DispatchSize` out of the `Shader` struct, add support for specifying `VertexCount`
This issue stems from the fact that currently the vertex count for drawing a triangle is implied from the vertex buffer. However, drawing with the Traditional Rasterization without a vertex buffer is a very valid use case. (Example: A post-processing filter in a pixel shader, where the vertex shader generates a full-screen quad using the vertex ID). Since this is a case we would like to add tests for, we need to be able to specify the `VertexCount`.

However, at some point we may also want to specify an `IndexCount`, `InstanceCount`, `IndexOffset`, `InstanceOffset`, `VertexOffset`, or even more it doesn't quite make sense to store these fields in the `Shader` type. These parameters are not part of the shader, but are related to dispatching the pipeline.

Instead, the user should be able to specify the `DispatchParameters` for the pipeline instead.
This will also simplify the backend as in cases with more than one shader we don't need to scan for the shader containing the parameters for dispatching work on the GPU.

Additionally, we can add a default dispatch size of [1,1,1] for compute shaders, since most test launch a single group.
Similarly, we can still imply the vertex count from the vertex buffer if the user did not specify one in the test.

Finally, we should validate the passed `DispatchParameters` against the type of pipeline to check if the user did not accidentally specify parameters that are irrelevant and could cause confusion or mistakes.

manon-traverse:
Fixes #1152 

There was no way to specify the vertex count for a test that uses the TraditionalRaster pipeline.
Since vertex buffers are not required for doing rasterization, and is actually quite a common use case, we need to be able to specify that as well.
In the future we will also need to support indexed and instanced draws, for which more parameters need to be specified.
Specifying these parameters are part of the shader doesn't really make sense since they apply to the whole pipeline.
For those reasons I am introducing a `DispatchParameters` field where the user can specify the exact parameters for how a pipeline should be dispatched.

While making this change I found that 99% of the compute shader tests dispatch a single threadgroup. Since there is no use-case for dispatching no threads (as far as I am aware), I made dispatching a single thread group a default value. This simplifies writing tests.

For TraditionalRaster tests we can still rely on calculating the vertex count from the size of the vertex buffer and vertex stride if it is not specified in the .yaml section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,8 @@ This framework provides a YAML representation for describing GPU pipelines and b
 Shaders:
   - Stage: Compute
     Entry: main
-    DispatchSize: [1, 1, 1]
+DispatchParameters:
+  DispatchGroupCount: [2, 1, 1]
 Buffers:
   - Name: Constants
     Format: Int32
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ Shaders:
   - Stage: Compute
     Entry: main
 DispatchParameters:
-  DispatchGroupCount: [2, 1, 1]
+  DispatchGroupCount: [2, 1, 1] # Define how many groups to dispatch, if omitted one group is launched.
 Buffers:
   - Name: Constants
     Format: Int32
--- a/lib/Support/Pipeline.cpp
+++ b/lib/Support/Pipeline.cpp
@@ -52,6 +52,7 @@ void MappingTraits<offloadtest::Pipeline>::mapping(IO &I,
   I.mapRequired("DescriptorSets", P.Sets);
   I.mapOptional("Bindings", P.Bindings);
   I.mapOptional("PushConstants", P.PushConstants);
+  I.mapOptional("DispatchParameters", P.DispatchParameters);
 
   if (!I.outputting()) {
     for (auto &D : P.Sets) {
@@ -464,6 +465,12 @@ void MappingTraits<offloadtest::PushConstantValue>::mapping(
   }
 }
 
+void MappingTraits<offloadtest::DispatchParametersSet>::mapping(
+    IO &I, offloadtest::DispatchParametersSet &Set) {
+  I.mapOptional("DispatchGroupCount", Set.DispatchGroupCount);
+  I.mapOptional("VertexCount", Set.VertexCount);
+}
+
 void MappingTraits<offloadtest::OutputProperties>::mapping(
     IO &I, offloadtest::OutputProperties &P) {
   I.mapRequired("Height", P.Height);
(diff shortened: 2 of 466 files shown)


README.md
```diff
@@ -104,7 +104,8 @@ This framework provides a YAML representation for describing GPU pipelines and b
 Shaders:
   - Stage: Compute
     Entry: main
-    DispatchSize: [1, 1, 1]
+DispatchParameters:
+  DispatchGroupCount: [2, 1, 1]
```
MarijnS95: Does the rest of this YAML sample still make sense (i.e. buffer size) if suddenly launching 2 groups instead of the original 1?

lib/Support/Pipeline.cpp
```diff
@@ -589,3 +592,23 @@ llvm::Error offloadtest::Pipeline::validatePipelineKind() {
...
+          "DispatchParameters.VertexCount set on a Compute pipeline. Only "
+          "allowed on a TraditionalRaster pipeline.");
+    break;
+  case ShaderPipelineKind::TraditionalRaster:
+    if (DispatchParameters.DispatchGroupCount !=
+        std::array<uint32_t, 3>{1, 1, 1})
+      return llvm::createStringError(
+          "DispatchParameters.DispatchGroupCount set on a TraditionalRaster "
+          "pipeline. Only allowed on a Compute pipeline.");
+    break;
```
MarijnS95, line 610: Should we validate that `VertexCount` is _not_ set if a vertex buffer was provided, or are there sensible reasons to override the "default" number of vertices in a buffer?

manon-traverse: There are valid reasons to specify this manually even though you have defined a vertex buffer. Especially when we start passing arguments like `FirstVertex`.

MarijnS95: The only validation would be if `FirstVertex + VertexCount <= NumVerticesInBuffer` perhaps?

README.md
manon-traverse: There is no shader, so there is not really anything to make sense of. I did add a comment explaining that this property can be omitted and what happens if you do. (launch a single group).

MarijnS95: Right yes.

MarijnS95 (approved):

lib/Support/Pipeline.cpp
manon-traverse: Technically that would still be a valid case though, but you would be reading zeroes.

bogner (approved):

PR #1155, opened 2026-05-05
Make vertex buffers optional in the `TraditionalRaster` pipeline and add `VerticesFromVertexID.test`

### LINKED ISSUE #1154: Make specifying a vertex buffer optional
Running a TraditionalRaster pipeline without a vertex buffer is a very common use case. Super common example: a full-screen pass using a full-screen quad generated based on the vertex ID.

Currently specifying a vertex buffer is required while there is no clear reason to do so.
After #1153 is merged, the vertex count that is fed in the draw call is also no longer implied from the vertex buffer either.
Then we can just drop the requirement for a vertex buffer.

manon-traverse:
Fixes #1154 

Add a TraditionalRaster test that doesn't use a vertex buffer, but generates a triangle using the SV_VertexID. This is a pretty common use case in game engines.

To make this work this changes makes the specifying a vertex buffer optional.

Golden image is introduced in: https://github.com/llvm/offload-golden-images/pull/7

### DIFF
--- a/include/Support/Pipeline.h
+++ b/include/Support/Pipeline.h
@@ -383,11 +383,11 @@ struct VertexAttribute {
 
 struct IOBindings {
   std::string VertexBuffer;
-  CPUBuffer *VertexBufferPtr;
+  CPUBuffer *VertexBufferPtr = nullptr;
   llvm::SmallVector<VertexAttribute> VertexAttributes;
 
   std::string RenderTarget;
-  CPUBuffer *RTargetBufferPtr;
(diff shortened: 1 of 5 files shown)


manon-traverse:
NOTE: CI is currently only failing because of the missing golden image in the offload-golden-images repository.

llvm-beanz (approved):
I think this looks good to me. My only feedback is to use the pre-existing golden image instead of adding a new identical one if possible.

bogner (approved):

PR #1156, opened 2026-05-06
Remove error return value from `unmap()`.

manon-traverse:
`unmap` never returns an error value, so there is no point in having all the users of this function checking for errors. Instead it can just have a void return type.


MarijnS95 (approved):

EmilioLaiso (commented):
LGTM

PR #1157, opened 2026-05-06
Mark `CreateDesc` reference parameters as `const`

manon-traverse:
Saw this one while working on something else. This was an oversight when these functions were introduced.


MarijnS95 (approved):

EmilioLaiso (commented):
LGTM

Issue #102, opened 2025-04-30
Add test for InterlockedAdd


spall:
In Test\Feature\Intrinsics create, if applicable: (reference the HLSL headers for the types supported by the intrinsic)
- [ ] Test for 16 bit int types (< intrinsic >.int16.test)
- [ ] Test for Half type (< intrinsic >.fp16.test)
- [ ] Test for 32 bit types (< intrinsic >.32.test)
- [ ] Test for 64 bit int types (< intrinsic >.int64.test)
- [ ] Test for Double type (< intrinsic >.fp64.test)

A .test file should contain 
- [ ] a source portion; If multiple .test files can use the same source; it can go in its own file and be referenced in the run portion of each .test file instead. 
```
@@ -0,0 +1,54 @@
#--- source.hlsl

StructuredBuffer<float4> In : register(t0);
RWStructuredBuffer<float> Out : register(u1);

[numthreads(1,1,1)]
void main() {
// A single test can test for all scalar/vector length as shown here.
  Out[0] = length(In[0]); // Test float4
  Out[1] = length(In[1].x); // Test float
  Out[2] = length(In[1].yzw); // Test float3
  Out[3] = length(In[1].yz); // Test float2
}
```
- [ ] a  yaml portion; If multiple .test files can use the same YAML, it can go in its own file and be referenced in the run portion of each  .test file instead.
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
    Data: [ 4, 4, 4, 4, 3.14159, 0, 5, 12 ]
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
  - Name: ExpectedOut # The result we expect
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
Results:  # A test might have more than 1 result.
  - Result: Test1
    Rule: BufferFuzzy # there is also a BufferExact rule
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
[...]

PR #1163, opened 2026-05-07
Add tests for InterlockedAdd

bob80905:
This PR adds tests for the InterlockedAdd HLSL function, specifically for overloads where the destination parameter is not a resource handle.
Addresses https://github.com/llvm/offload-test-suite/issues/102

### DIFF
--- a/test/Feature/HLSLLib/InterlockedAdd.32.test
+++ b/test/Feature/HLSLLib/InterlockedAdd.32.test
@@ -78,7 +78,7 @@ Buffers:
   - Name: OutSlotsInt
     Format: UInt32
     Stride: 4
-    ZeroInitSize: 256
+    FillSize: 256
   - Name: ExpectedSlotsInt
     Format: UInt32
     Stride: 4
@@ -89,7 +89,7 @@ Buffers:
   - Name: OutSlotsUInt
     Format: UInt32
     Stride: 4
-    ZeroInitSize: 256
+    FillSize: 256
   - Name: ExpectedSlotsUInt
     Format: UInt32
     Stride: 4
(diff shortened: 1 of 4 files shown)


Icohedron:
It appears Intel is failing the 64-bit test on both Vulkan and DirectX
https://github.com/llvm/offload-test-suite/actions/runs/25529373174/job/74932097086?pr=1163#step:13:224
https://github.com/llvm/offload-test-suite/actions/runs/25529373174/job/74932097058?pr=1163#step:13:223

An issue should be filed and the tests XFAILed for Intel.

Icohedron (approved):

test/Feature/HLSLLib/InterlockedAdd.32.test
```diff
@@ -0,0 +1,152 @@
...
+[numthreads(64, 1, 1)]
+void main(uint3 GTID : SV_GroupThreadID) {
+  if (GTID.x == 0) {
+    CounterInt = 0;
+    CounterUInt = 0;
+    CounterIntNoOrig = 0;
+    CounterUIntNoOrig = 0;
+    SignedCounter = 1000;
+  }
+  SlotsInt[GTID.x] = 0;
```
joaosaffran, line 36: I believe we need NURI in here, 
```suggestion
  SlotsInt[NonUniformResourceAccess(GTID.x)] = 0;
```

joaosaffran (approved):
LGTM, Small nit regarding NURI

test/Feature/HLSLLib/InterlockedAdd.32.test
bob80905: I consulted Copilot and it looks like this would only be true if the array we are indexing is an array of resources. SlotsInt is an array of integers, and OutSlotsInt is a resource that contains integers, and neither is an array of resources. So I think we're fine here.

PR #1169, opened 2026-05-11
Implement a RenderEncoder for API agnostic render command recording.

### LINKED ISSUE #1160: Introduce a Render(Pass)CommandEncoder for recording rasterization commands
Recently a `ComputeEncoder` API was introduced which allows us to express us what to record on the command buffer in an API agnostic way. This allows us to have less code duplication between render backends and make it harder for the implementations to diverge.

We should introduce something similar for recording draw calls on the command buffer as well. This will give us the same benefits that the `ComputeEncoder` gives us for compute workloads, but for rasterization workloads.

Additionally, we can extend this render encoder with extra functionality like dispatching mesh shader based pipelines in the future.

manon-traverse:
Previously the `ComputeEncoder` was introduced to record compute dispatch commands in an API-agnostic manner. This uses the same pattern for graphics commands. To keep changes minimal only `setVertexBuffer` and `drawInstanced` functions are added. 

The next step would be to introduce a `meshDispatch` function for mesh shader support. This `RenderEncoder` then allows us to implement Mesh Shader support once properly instead of implementing it for each backend and then having to refactor it later.

Due to how Vulkan is set-up with `RenderPass` objects we needed to add an API-agnostic type along with the creation functions for those. On DX12 and Metal these just save the input parameters which are then used to create the `RenderEncoder`.

A drive-by change is moving the `copyBufferToBuffer` function to the `ComputeEncoder` because Metal does not support copy operations on the `RenderEncoder`.

### DIFF
--- a/include/API/CommandBuffer.h
+++ b/include/API/CommandBuffer.h
@@ -19,11 +19,28 @@
 #include "API/API.h"
 #include "API/Encoder.h"
 
+#include "llvm/ADT/SmallVector.h"
 #include "llvm/Support/Error.h"
 
 #include <memory>
 
 namespace offloadtest {
+
+class RenderPass;
+class Texture;
+
+/// Description used to begin a render pass for one render encoder. The
+/// RenderPass itself carries the format / load / store policy and is
+/// expected to be created up-front via Device::createRenderPass; this
+/// struct binds it to the actual textures for a single encoder's lifetime.
+/// The texture count and order must match the RenderPass's attachment
+/// layout: one texture per color attachment, plus the depth-stencil
+/// texture iff the RenderPass has one.
+struct RenderPassBeginDesc {
+  RenderPass *Pass = nullptr;
+  llvm::SmallVector<Texture *, 8> ColorAttachments;
+  Texture *DepthStencil = nullptr;
+};
 
 class CommandBuffer {
   GPUAPI Kind;
@@ -43,6 +60,18 @@ public:
     return llvm::createStringError(
         std::errc::not_supported,
         "createComputeEncoder not implemented for this backend");
+  }
+
+  /// Create a render command encoder for recording draw commands. The
+  /// returned encoder begins the render pass referenced by \p Desc.Pass
+  /// against the textures in \p Desc; the pass ends when the encoder's
+  /// endEncoding() runs (explicitly or from its destructor).
+  virtual llvm::Expected<std::unique_ptr<RenderEncoder>>
+  createRenderEncoder(const RenderPassBeginDesc &Desc) {
+    (void)Desc;
+    return llvm::createStringError(
+        std::errc::not_supported,
+        "createRenderEncoder not implemented for this backend");
   }
 };
 
--- a/include/API/CommandBuffer.h
+++ b/include/API/CommandBuffer.h
@@ -29,13 +29,6 @@ namespace offloadtest {
 class RenderPass;
 class Texture;
 
-/// Description used to begin a render pass for one render encoder. The
-/// RenderPass itself carries the format / load / store policy and is
-/// expected to be created up-front via Device::createRenderPass; this
-/// struct binds it to the actual textures for a single encoder's lifetime.
-/// The texture count and order must match the RenderPass's attachment
-/// layout: one texture per color attachment, plus the depth-stencil
-/// texture iff the RenderPass has one.
 struct RenderPassBeginDesc {
   RenderPass *Pass = nullptr;
   llvm::SmallVector<Texture *, 8> ColorAttachments;
@@ -62,10 +55,7 @@ public:
         "createComputeEncoder not implemented for this backend");
   }
 
-  /// Create a render command encoder for recording draw commands. The
-  /// returned encoder begins the render pass referenced by \p Desc.Pass
-  /// against the textures in \p Desc; the pass ends when the encoder's
-  /// endEncoding() runs (explicitly or from its destructor).
+  /// Create a render command encoder for recording draw commands.
   virtual llvm::Expected<std::unique_ptr<RenderEncoder>>
   createRenderEncoder(const RenderPassBeginDesc &Desc) {
     (void)Desc;
--- a/include/API/CommandBuffer.h
+++ b/include/API/CommandBuffer.h
@@ -49,20 +49,11 @@ public:
   /// Create a compute command encoder for recording dispatch commands.
   /// Barriers are automatically inserted between commands.
   virtual llvm::Expected<std::unique_ptr<ComputeEncoder>>
-  createComputeEncoder() {
-    return llvm::createStringError(
-        std::errc::not_supported,
-        "createComputeEncoder not implemented for this backend");
-  }
+  createComputeEncoder() = 0;
 
   /// Create a render command encoder for recording draw commands.
   virtual llvm::Expected<std::unique_ptr<RenderEncoder>>
-  createRenderEncoder(const RenderPassBeginDesc &Desc) {
-    (void)Desc;
-    return llvm::createStringError(
-        std::errc::not_supported,
-        "createRenderEncoder not implemented for this backend");
-  }
+  createRenderEncoder(const RenderPassBeginDesc &Desc) = 0;
 };
 
 } // namespace offloadtest
--- a/include/API/Encoder.h
+++ b/include/API/Encoder.h
@@ -20,6 +20,7 @@
 namespace offloadtest {
 
 class Buffer;
+class PipelineState;
 
 /// Base class for all command encoders. An encoder records commands into a
 /// command buffer. Call endEncoding() when done recording. Barriers are
@@ -80,6 +81,57 @@ public:
   /// pipeline state (e.g. the shader's numthreads attribute).
   virtual llvm::Error dispatch(uint32_t GroupCountX, uint32_t GroupCountY,
                                uint32_t GroupCountZ) = 0;
+};
+
+/// Primitive topology used at draw time. The bound pipeline's topology
+/// category (point / line / triangle) must match the value passed to draw().
+enum class PrimitiveTopology {
+  TriangleList,
+};
+
+struct Viewport {
+  float X = 0.0f, Y = 0.0f;
+  float Width = 0.0f, Height = 0.0f;
+  float MinDepth = 0.0f, MaxDepth = 1.0f;
+};
+
+struct ScissorRect {
+  int32_t X = 0, Y = 0;
+  uint32_t Width = 0, Height = 0;
+};
+
+/// Encoder for recording rasterization draw commands within a render pass.
+/// Created via CommandBuffer::createRenderEncoder(const RenderPassDesc &).
+///
+/// State model: every encoder starts with no viewport, scissor, or vertex
+/// buffer bindings. The caller is required to set viewport and scissor (and
+/// any vertex buffers needed by the pipeline) before the first draw — there
+/// is no carry-over from a previous pass. The pipeline state is passed
+/// directly to draw() so a draw cannot be issued with no pipeline bound or
+/// a stale pipeline from another pass.
+class RenderEncoder : public CommandEncoder {
+public:
+  using CommandEncoder::CommandEncoder;
+
+  virtual void setViewport(const Viewport &VP) = 0;
+  virtual void setScissor(const ScissorRect &Rect) = 0;
+
+  /// Bind a vertex buffer at \p Slot with the given per-vertex \p Stride in
+  /// bytes. \p Stride is required at set time (DX12 carries it on the buffer
+  /// view, not the PSO). Pass nullptr to unbind; \p Stride is ignored when
+  /// unbinding.
+  virtual void setVertexBuffer(uint32_t Slot, Buffer *VB, size_t Offset,
+                               uint32_t Stride) = 0;
+
+  /// Non-indexed instanced draw. Mirrors DX12 DrawInstanced / vkCmdDraw /
+  /// MTL drawPrimitives. The pipeline is bound as part of the call.
+  /// Resource bindings (root descriptor tables, descriptor sets, argument
+  /// buffers) must be set up on the underlying command buffer before draw —
+  /// the abstraction for those is still WIP.
+  virtual llvm::Error
+  drawInstanced(const PipelineState &PSO, PrimitiveTopology Topology,
+                uint32_t VertexCount, uint32_t InstanceCount,
+                uint32_t FirstVertex = 0, uint32_t FirstInstance = 0) = 0;
 };
 
 } // namespace offloadtest
--- a/include/API/Encoder.h
+++ b/include/API/Encoder.h
@@ -94,15 +94,6 @@ struct ScissorRect {
   uint32_t Width = 0, Height = 0;
 };
 
-/// Encoder for recording rasterization draw commands within a render pass.
-/// Created via CommandBuffer::createRenderEncoder(const RenderPassDesc &).
-///
-/// State model: every encoder starts with no viewport, scissor, or vertex
-/// buffer bindings. The caller is required to set viewport and scissor (and
-/// any vertex buffers needed by the pipeline) before the first draw — there
-/// is no carry-over from a previous pass. The pipeline state is passed
-/// directly to draw() so a draw cannot be issued with no pipeline bound or
-/// a stale pipeline from another pass.
 class RenderEncoder : public CommandEncoder {
 public:
   using CommandEncoder::CommandEncoder;
@@ -110,18 +101,9 @@ public:
   virtual void setViewport(const Viewport &VP) = 0;
   virtual void setScissor(const ScissorRect &Rect) = 0;
 
-  /// Bind a vertex buffer at \p Slot with the given per-vertex \p Stride in
-  /// bytes. \p Stride is required at set time (DX12 carries it on the buffer
-  /// view, not the PSO). Pass nullptr to unbind; \p Stride is ignored when
-  /// unbinding.
   virtual void setVertexBuffer(uint32_t Slot, Buffer *VB, size_t Offset,
                                uint32_t Stride) = 0;
 
-  /// Non-indexed instanced draw. Mirrors DX12 DrawInstanced / vkCmdDraw /
-  /// MTL drawPrimitives. The pipeline is bound as part of the call.
-  /// Resource bindings (root descriptor tables, descriptor sets, argument
-  /// buffers) must be set up on the underlying command buffer before draw —
-  /// the abstraction for those is still WIP.
   virtual llvm::Error drawInstanced(const PipelineState &PSO,
                                     uint32_t VertexCount,
                                     uint32_t InstanceCount,
--- a/include/API/Encoder.h
+++ b/include/API/Encoder.h
@@ -43,12 +43,6 @@ public:
   GPUAPI getAPI() const { return API; }
   bool isEnded() const { return Ended; }
 
-  /// Copy \p Size bytes from \p Src at \p SrcOffset to \p Dst at
-  /// \p DstOffset.
-  virtual llvm::Error copyBufferToBuffer(Buffer &Src, size_t SrcOffset,
-                                         Buffer &Dst, size_t DstOffset,
-                                         size_t Size) = 0;
-
   /// Begin a named debug group. Visible in GPU debuggers (PIX, RenderDoc,
   /// Xcode). Must be balanced by a corresponding popDebugGroup() call.
   virtual void pushDebugGroup(llvm::StringRef Label) {}
@@ -81,6 +75,12 @@ public:
   /// pipeline state (e.g. the shader's numthreads attribute).
   virtual llvm::Error dispatch(uint32_t GroupCountX, uint32_t GroupCountY,
                                uint32_t GroupCountZ) = 0;
+
+  /// Copy \p Size bytes from \p Src at \p SrcOffset to \p Dst at
+  /// \p DstOffset.
+  virtual llvm::Error copyBufferToBuffer(Buffer &Src, size_t SrcOffset,
+                                         Buffer &Dst, size_t DstOffset,
+                                         size_t Size) = 0;
 };
 
 /// Primitive topology used at draw time. The bound pipeline's topology
--- a/include/API/Encoder.h
+++ b/include/API/Encoder.h
@@ -83,12 +83,6 @@ public:
                                          size_t Size) = 0;
 };
 
-/// Primitive topology used at draw time. The bound pipeline's topology
-/// category (point / line / triangle) must match the value passed to draw().
-enum class PrimitiveTopology {
-  TriangleList,
-};
-
 struct Viewport {
   float X = 0.0f, Y = 0.0f;
   float Width = 0.0f, Height = 0.0f;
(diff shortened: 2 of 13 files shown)


include/API/CommandBuffer.h
```diff
@@ -44,6 +54,15 @@ class CommandBuffer {
         std::errc::not_supported,
         "createComputeEncoder not implemented for this backend");
   }
+
+  /// Create a render command encoder for recording draw commands.
+  virtual llvm::Expected<std::unique_ptr<RenderEncoder>>
+  createRenderEncoder(const RenderPassBeginDesc &Desc) {
+    (void)Desc;
```
llvm-beanz:
```suggestion
  createRenderEncoder(const RenderPassBeginDesc &) {
```
You can just not name the parameter, then it won't produce an unused variable warning.

include/API/CommandBuffer.h
```diff
@@ -44,6 +54,15 @@ class CommandBuffer {
...
+    return llvm::createStringError(
+        std::errc::not_supported,
+        "createRenderEncoder not implemented for this backend");
```
bogner: This matches the existing pattern, so I don't necessarily think we should do anything different in this PR, but should we be making these functions pure virtual since they kind of need to be implemented for all backends anyway?

lib/API/DX/Device.cpp
```diff
@@ -652,6 +657,204 @@ DXCommandBuffer::createComputeEncoder() {
...
+    const auto &DXPSO = llvm::cast<DXPipelineState>(PSO);
+    CB.CmdList->SetGraphicsRootSignature(DXPSO.RootSig.Get());
+    CB.CmdList->SetPipelineState(DXPSO.PSO.Get());
+    CB.CmdList->IASetPrimitiveTopology(D3D_PRIMITIVE_TOPOLOGY_TRIANGLELIST);
+    return llvm::Error::success();
+  }
+
+public:
+  DXRenderEncoder(DXCommandBuffer &CB)
+      : RenderEncoder(GPUAPI::DirectX), CB(CB) {}
```
bogner, line 698: We should probably follow the [rule of three](https://en.cppreference.com/cpp/language/rule_of_three) and provide a definition for the assignment operator here too, or maybe even rule of five and define or explicitly delete move constructor and assignment.

bogner (approved):
LGTM. The comment about rule of 3/5 obviously applies to the classes added in the other backends as well.

include/API/CommandBuffer.h
manon-traverse: Yeah just following the pattern, but I agree that they should just be pure virtual. It makes no sense having a backend that doesn't support render or compute encoders.

I have marked them both as pure virtual.

lib/API/DX/Device.cpp
manon-traverse: I have applied the rule of 5.

include/API/Encoder.h
```diff
@@ -80,6 +75,40 @@ class ComputeEncoder : public CommandEncoder {
...
+public:
+  using CommandEncoder::CommandEncoder;
+
+  virtual void setViewport(const Viewport &VP) = 0;
+  virtual void setScissor(const ScissorRect &Rect) = 0;
+
+  virtual void setVertexBuffer(uint32_t Slot, Buffer *VB, size_t Offset,
+                               uint32_t Stride) = 0;
+
+  virtual llvm::Error drawInstanced(const PipelineState &PSO,
```
MarijnS95, line 107: Would be nice if we could use `PipelineState` in `dispatch()` as well, that's probably intended for a followup PR 👍

include/API/Texture.h
```diff
@@ -150,6 +150,7 @@ class Texture {
   Texture &operator=(const Texture &) = delete;
 
   GPUAPI getAPI() const { return API; }
+  virtual const TextureCreateDesc &getDesc() const = 0;
```
MarijnS95, line 153: Would it make sense if this `Texture` base class stored `Desc`?

lib/API/VK/Device.cpp
```diff
@@ -758,6 +764,285 @@ class VulkanPipelineState : public offloadtest::PipelineState {
...
+  }
+
+  void insertDebugSignpost(llvm::StringRef Label) override {
+    if (!CB.CmdInsertDebugUtilsLabel)
+      return;
+    VkDebugUtilsLabelEXT LabelInfo = {};
+    LabelInfo.sType = VK_STRUCTURE_TYPE_DEBUG_UTILS_LABEL_EXT;
+    LabelInfo.pLabelName = Label.data();
+    CB.CmdInsertDebugUtilsLabel(CB.CmdBuffer, &LabelInfo);
+  }
```
MarijnS95, line 850: It's a tiny bit unfortunate that we have to duplicate this across `*Encoder`s 😅

lib/API/VK/Device.cpp
```diff
@@ -758,6 +764,285 @@ class VulkanPipelineState : public offloadtest::PipelineState {
...
+    // Stride is needed in DX12 at binding time, ignore parameter here.
+    if (!VB) {
+      VkBuffer NullBuf = VK_NULL_HANDLE;
+      const VkDeviceSize Zero = 0;
+      vkCmdBindVertexBuffers(CB.CmdBuffer, Slot, 1, &NullBuf, &Zero);
+      return;
+    }
+    VkBuffer Handle = llvm::cast<VulkanBuffer>(*VB).Buffer;
+    const VkDeviceSize VKOffset = Offset;
+    vkCmdBindVertexBuffers(CB.CmdBuffer, Slot, 1, &Handle, &VKOffset);
```
MarijnS95: This early-out with `return` reads a bit weird, since it's the same (lines of) code as the non-early-out 😅

EDIT: Either an `if-else`, or perhaps?:

```suggestion
    VkBuffer Handle = 0;
    const VkDeviceSize VKOffset = 0;
    if (VB) {
      Handle = llvm::cast<VulkanBuffer>(*VB).Buffer;
      VKOffset = Offset;
    }
    vkCmdBindVertexBuffers(CB.CmdBuffer, Slot, 1, &Handle, &VKOffset);
```

lib/API/VK/Device.cpp
```diff
@@ -758,6 +764,285 @@ class VulkanPipelineState : public offloadtest::PipelineState {
...
+                      VKPSO.Pipeline);
+    vkCmdDraw(CB.CmdBuffer, VertexCount, InstanceCount, FirstVertex,
+              FirstInstance);
+    return llvm::Error::success();
+  }
+
+  void endEncodingImpl() override {
+    vkCmdEndRenderPass(CB.CmdBuffer);
+    popDebugGroup();
+  }
```
MarijnS95, line 913: Tiny nit: it seems "unbalanced" to perform the `pop` in the `RenderEncoder` implementation, while the "caller" is responsible for the `push`, which feels like something the constructor should do?

lib/API/VK/Device.cpp
```diff
@@ -758,6 +764,285 @@ class VulkanPipelineState : public offloadtest::PipelineState {
...
+      assert(ColorCV &&
+             "RenderTarget OptimizedClearValue must be a ClearColor");
+      CV.color = {{ColorCV->R, ColorCV->G, ColorCV->B, ColorCV->A}};
+    }
+    ClearValues.push_back(CV);
+
+    if (Tex.Desc.Width > Width)
+      Width = Tex.Desc.Width;
+    if (Tex.Desc.Height > Height)
+      Height = Tex.Desc.Height;
```
MarijnS95: We should likely assert that the sizes of color attachments and depthstencil are identical, because https://docs.vulkan.org/refpages/latest/refpages/source/VkFramebufferCreateInfo.html#VUID-VkFramebufferCreateInfo-flags-04533 specifies that all images must be at least as large as the framebuffer. IOW, the framebuffer must take the size of the *smallest* image.

lib/API/VK/Device.cpp
```diff
@@ -1194,20 +1479,45 @@ class VulkanDevice : public offloadtest::Device {
...
+      DS.DepthStore = StoreAction::DontCare;
+      DS.StencilLoad = LoadAction::DontCare;
+      DS.StencilStore = StoreAction::DontCare;
+      PassDesc.DepthStencil = DS;
+    }
+
+    // NOTE: After pipeline creation this render pass can be dropped. Later
+    // render passes just need to be compatible with this render pass, or in
+    // other words: the format, sample count and number of targets (rt and ds),
+    // need to match.
```
MarijnS95: For the "explicit" `DontCare` initialization above, perhaps mention/comment that these are unused/irrelevant per https://docs.vulkan.org/spec/latest/chapters/renderpass.html#renderpass-compatibility too?

Perhaps:

```suggestion
    // NOTE: After pipeline creation this render pass can be dropped. Later
    // render passes just need to be compatible with this render pass, or in
    // other words: the format, sample count and number of targets (rt and ds),
    // need to match, but Load/Store ops (and initial/final layout) are ignored.
```

lib/API/VK/Device.cpp
```diff
@@ -1660,6 +1965,77 @@ class VulkanDevice : public offloadtest::Device {
...
+      VkAttachmentDescription AD = {};
+      AD.format = getVulkanFormat(Color.Fmt);
+      AD.samples = VK_SAMPLE_COUNT_1_BIT;
+      AD.loadOp = getVkLoadOp(Color.Load);
+      AD.storeOp = getVkStoreOp(Color.Store);
+      AD.stencilLoadOp = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
+      AD.stencilStoreOp = VK_ATTACHMENT_STORE_OP_DONT_CARE;
+      AD.initialLayout = Color.Load == LoadAction::Load
+                             ? VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL
+                             : VK_IMAGE_LAYOUT_UNDEFINED;
```
MarijnS95, line 1974: Perhaps comment that it's allowed to mismatch the "current" image layout by setting `VK_IMAGE_LAYOUT_UNDEFINED`, but that'll "invalidate" the image contents (which is exactly what we want when the `LoadAction` is not `Load`, but implying that it has `StoreAction::Store` otherwise the texture doesn't seem to be used in the render pass at all).

include/API/Texture.h
manon-traverse: No, I don't want regular inheritance. Only pure-virtual base classes.

lib/API/MTL/MTLDevice.cpp
```diff
@@ -526,6 +559,237 @@ MTLCommandBuffer::createComputeEncoder() {
...
+  ~MTLRenderEncoder() override { endEncoding(); }
+
+  static bool classof(const CommandEncoder *E) {
+    return E->getAPI() == GPUAPI::Metal;
+  }
+
+  /// Access the underlying Metal encoder for state that the abstract
+  /// RenderEncoder API does not yet cover (depth-stencil state, cull
+  /// mode, etc.). Returns nullptr after endEncoding().
+  MTL::RenderCommandEncoder *getNative() const { return RenderEnc; }
```
MarijnS95, line 613: Stale comment? You specifically store the DS-state and cull mode in the `PSO` now to access directly from `drawInstanced()`. `getNative` is instead used to bind the descriptor heap and argument buffer?

lib/API/VK/Device.cpp
manon-traverse: I think your alternative is more convoluted tbh

lib/API/MTL/MTLDevice.cpp
```diff
@@ -526,6 +559,237 @@ MTLCommandBuffer::createComputeEncoder() {
...
+  void popDebugGroup() override {
+    if (RenderEnc)
+      RenderEnc->popDebugGroup();
+  }
+
+  void insertDebugSignpost(llvm::StringRef Label) override {
+    if (RenderEnc)
+      RenderEnc->insertDebugSignpost(
+          NS::String::string(Label.data(), NS::UTF8StringEncoding));
+  }
```
MarijnS95, line 630: Nit: feels like calling these should just be an error if `RenderEnc` was `nullptr`.

lib/API/VK/Device.cpp
manon-traverse: Deduplicated them!

lib/API/MTL/MTLDevice.cpp
```diff
@@ -526,6 +559,237 @@ MTLCommandBuffer::createComputeEncoder() {
...
+        std::errc::invalid_argument,
+        "RenderPassBeginDesc color attachment count does not match its "
+        "RenderPass.");
+  if (PassDesc.DepthStencil.has_value() != (Desc.DepthStencil != nullptr))
+    return llvm::createStringError(std::errc::invalid_argument,
+                                   "RenderPassBeginDesc depth-stencil "
+                                   "presence does not match its RenderPass.");
+
+  MTL::RenderPassDescriptor *MTLDesc =
+      MTL::RenderPassDescriptor::alloc()->init();
```
MarijnS95, line 727: It doesn't seem like `renderTargetWidth`/`renderTargetHeight` is initialized, but IIRC it's superfluous or inferred when textures are attached. (i.e. only needed for render passes without attachments)

lib/API/MTL/MTLDevice.cpp
```diff
@@ -1549,6 +1787,27 @@ class MTLDevice : public offloadtest::Device {
...
+      ColorAttachment.Fmt = *FormatOrErr;
+      ColorAttachment.Load = LoadAction::Clear;
+      ColorAttachment.Store = StoreAction::Store;
+
+      DepthStencilAttachmentFormatDesc DSAttachment = {};
+      DSAttachment.Fmt = Format::D32FloatS8Uint;
+      DSAttachment.DepthLoad = LoadAction::Clear;
+      DSAttachment.DepthStore = StoreAction::DontCare;
+      DSAttachment.StencilLoad = LoadAction::Clear;
+      DSAttachment.StencilStore = StoreAction::DontCare;
```
MarijnS95, line 1809: This seems different from Vulkan and Dx12 which both have:


```suggestion
      DSAttachment.DepthLoad = LoadAction::Clear;
      DSAttachment.DepthStore = StoreAction::Store;
      DSAttachment.StencilLoad = LoadAction::DontCare;
      DSAttachment.StencilStore = StoreAction::DontCare;
```

lib/API/DX/Device.cpp
```diff
@@ -652,6 +657,208 @@ DXCommandBuffer::createComputeEncoder() {
...
+  const D3D12_CPU_DESCRIPTOR_HANDLE *DSVPtr = nullptr;
+  if (Desc.DepthStencil) {
+    auto &DXDS = llvm::cast<DXTexture>(*Desc.DepthStencil);
+    if (DXDS.DSVHandle.ptr == 0)
+      return llvm::createStringError(
+          std::errc::invalid_argument,
+          "Depth-stencil texture was not created with DepthStencil usage.");
+    DSTexture = &DXDS;
+    DSVHandle = DXDS.DSVHandle;
+    DSVPtr = &DSVHandle;
```
MarijnS95, line 820: Nit: the pattern for this in Vulkan is `HasDS ? &DepthReference : nullptr`. Maybe you could update that path to just store the default-`nullptr` pointer too, or use a ternary if below in place of `DSVPtr`?

lib/API/VK/Device.cpp
MarijnS95: Then `if-else`?

lib/API/VK/Device.cpp
manon-traverse: Okay, but then we take the smallest image right? Why would we assert instead?

lib/API/DX/Device.cpp
manon-traverse: I am not following you, please elaborate

lib/API/MTL/MTLDevice.cpp
manon-traverse: According to the docs it just constraints the render target size. I expect it just uses the smallest attachment size if you don't specify it. I'll just specify it just in case the same way as in vulkan.

lib/API/MTL/MTLDevice.cpp
manon-traverse: Could we just assert instead?

lib/API/MTL/MTLDevice.cpp
manon-traverse: Well spotted! Removing the specifics from this comment since those will just keep going out of date.

lib/API/DX/Device.cpp
manon-traverse: Okay read your comment again, I think I understand now.

The key difference is that In DX12 you pass a pointer to a handle that's stored in the `DXTexture`, but in Vulkan you pass a pointer to a locally constructed object. On top of that you need to pass `nullptr` instead if you don't have a depth stencil attachment.

If we want these two to be the same you would need to create a pointer to an the `DepthReference` object that lives on the stack. I see no real advantage of doing that.

I think these small inconsistencies are just something we will have to live with otherwise we are going to get bogged down into details that in the big picture don't really matter.

lib/API/MTL/MTLDevice.cpp
MarijnS95: Yeah, I think so. Since it's all public API (IIRC) it wouldn't make a lot of sense to drop the condition and let it segfault in this "unexpected" scenario.

lib/API/VK/Device.cpp
MarijnS95: Yeah, take the smallest image.

Assert would be in case we have images of different sizes; I'm not sure what's expected in that case, so I'd rather fail early and decide then how we'd handle such a scenario if it occurs at all.

lib/API/MTL/MTLDevice.cpp
```diff
@@ -526,6 +559,262 @@ MTLCommandBuffer::createComputeEncoder() {
...
+                            static_cast<NS::UInteger>(FirstVertex),
+                            static_cast<NS::UInteger>(VertexCount),
+                            static_cast<NS::UInteger>(InstanceCount),
+                            static_cast<NS::UInteger>(FirstInstance));
+
+    return llvm::Error::success();
+  }
+
+  void endEncodingImpl() override {
+    if (RenderEnc) {
```
MarijnS95, line 694: Probably also worth asserting, so that callers don't call this >1 times?

lib/API/MTL/MTLDevice.cpp
```diff
@@ -526,6 +559,262 @@ MTLCommandBuffer::createComputeEncoder() {
...
+      Height = Tex.Desc.Height;
+  }
+
+  if (Width == ~0u && Height == ~0u) {
+    Width = 0;
+    Height = 0;
+  }
+
+  MTLDesc->setRenderTargetWidth(Width);
+  MTLDesc->setRenderTargetHeight(Height);
```
MarijnS95, line 794: I probably shouldn't have asked for this because IIRC these properties are superfluous and inferred, but at least this is more clear!

EDIT: More clear that we believe it takes the smallest attachment - or does Metal take the largest? Only on Vulkan is it documented that all images must be equal to or larger than the framebuffer. But hey, platform-parity 👍

lib/API/VK/Device.cpp
```diff
@@ -1660,6 +1958,82 @@ class VulkanDevice : public offloadtest::Device {
...
+      AD.samples = VK_SAMPLE_COUNT_1_BIT;
+      AD.loadOp = getVkLoadOp(Color.Load);
+      AD.storeOp = getVkStoreOp(Color.Store);
+      AD.stencilLoadOp = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
+      AD.stencilStoreOp = VK_ATTACHMENT_STORE_OP_DONT_CARE;
+      // If we are loading, the layout MUST be defined and in color attachment
+      // optimal state.
+      // If we are NOT loading (clearing or don't care), we are discarding the
+      // original contents of the texture, and use an undefined layout. This
+      // allows us to use _any_ layout including unitialized textures.
```
MarijnS95, line 1971:
```suggestion
      // allows us to receive a texture in _any_ layout including unitialized textures.
```

My nit yesterday was about this specifically allowing input textures in *any* layout (one that was just created and is still `UNDEFINED`, one that's reused after being `STORAGE` for some compute, or one that was already/also written by an earlier render pass and is still in `COLOR_ATTACHMENT_OPTIMAL` - all source layouts are valid when we "pretend" it's `UNDEFINED`, just that the contents are/become inaccessible).

lib/API/DX/Device.cpp
MarijnS95: Well on Dx12 the `D3D12_CPU_DESCRIPTOR_HANDLE DSVHandle = {};` here is also a stack-local; `DSVPtr = &DSVHandle;` creates a pointer to the stack doesn't it?

On both Vulkan and Dx12 a `nullptr` is passed if there's no DS, just in slightly different syntactical (not semantical) ways.

MarijnS95 (approved):
Nice, thanks for seeing though on all the changes! The only remaining discussion revolves around render passes with render targets of different sizes; this seems to be disallowed on all backends except Vulkan (where the framebuffer for a render pass may not be larger than the smallest `VkImage` in the list).

lib/API/MTL/MTLDevice.cpp
manon-traverse: But the only time this goes wrong is if our metal implementation messes up right? The RenderEnc should just always be there. And if it's not there we should make sure it's there.

At least that's how I interpreted the design?

I don't want to get into a situation where every single function can return an error. A lot of functions that return errors only do so if something went horribly wrong, in which case asserting is also fine imo.

lib/API/VK/Device.cpp
manon-traverse: As discussed IRL, DX12 requires them all to be the same size. Finding the sizes and validating that they are all the same has been moved to a shared helper function.

lib/API/MTL/MTLDevice.cpp
manon-traverse: We are now enforcing the same sizes since DX12 requires that. So setting these doesn't really matter that much, but in case we are interpreting the API incorreclty it at least won't be wrong.

lib/API/MTL/MTLDevice.cpp
manon-traverse: The `endEncoding` function filters out multiple calls. If you want, we can use an assert there instead in a follow-up PR?

lib/API/DX/Device.cpp
manon-traverse: I'm sorry, but this is too minor of a nit to slow down this PR further. If you would like, you can open an issue or do a follow-up PR.

lib/API/VK/Device.cpp
manon-traverse: use/receive same thing to me?

MarijnS95: To me `use` sounds like we're *accessing* the texture in that layout. `receive` on the other hand is just the initial incoming layout/status of the texture, that the implementation will **first transition into something appropriate** before we can use it.

lib/API/MTL/MTLDevice.cpp
MarijnS95: I meant assert (what you have now) or segfault (just perform the ptr access, it should never become NULL). Not return an `llvm::Error`, as it's unexpected to reach that case.

PR #1175, opened 2026-05-12
Add a Minimal Mesh Shader Test that runs on the DX12 backend.

### LINKED ISSUE #1161: Add support to .test files for specifying Mesh Shader tests
The .test file format currently does not support mesh shaders.
For this extra enums and yaml parsing needs to be added.

This includes:
- [x] Add Mesh and Amplification shader stages
- [x] Add MeshShaderRaster pipeline type
- [x] Add a test with a mesh shader with a fail marking on all backends types.
- [x] Add an assert/error for a still unimplemented mesh shader path in each backend.

manon-traverse:
Fixes #1161

This PR introduces a new Mesh Shader SimpleTriangle test as well as render backend support for mesh shaders in DX12.

Perhaps we should rename the test to `VerticesFromThreadID` since this test also creates vertices based on the threadID.

To keep the PRs small only DX12 support is added here. In follow-up PRs we can add support for Vulkan and Metal as well.

### DIFF
--- a/include/Support/Pipeline.h
+++ b/include/Support/Pipeline.h
@@ -26,15 +26,25 @@
 
 namespace offloadtest {
 
-enum class Stages { Compute, Vertex, Pixel };
+enum class Stages {
+  // Compute
+  Compute,
+
+  // Traditional Raster
+  Vertex,
+  Pixel,
+
+  // Mesh Shader
+  Amplification,
+  Mesh
+};
 inline constexpr std::array AllStages = {
-    Stages::Compute,
-    Stages::Vertex,
-    Stages::Pixel,
+    Stages::Compute,       Stages::Vertex, Stages::Pixel,
+    Stages::Amplification, Stages::Mesh,
 };
 inline constexpr size_t NumStages = AllStages.size();
 
-enum class ShaderPipelineKind { Compute, TraditionalRaster };
+enum class ShaderPipelineKind { Compute, TraditionalRaster, MeshShaderRaster };
 
 enum class Rule { BufferExact, BufferFloatULP, BufferFloatEpsilon };
 
@@ -503,6 +513,12 @@ struct Pipeline {
   bool isTraditionalRaster() const {
     return Kind == ShaderPipelineKind::TraditionalRaster;
   }
+  bool isMeshShaderRaster() const {
+    return Kind == ShaderPipelineKind::MeshShaderRaster;
+  }
+  bool isRaster() const {
+    return isTraditionalRaster() || isMeshShaderRaster();
+  }
 };
 } // namespace offloadtest
 
@@ -713,6 +729,8 @@ template <> struct ScalarEnumerationTraits<offloadtest::Stages> {
     ENUM_CASE(Compute);
     ENUM_CASE(Vertex);
     ENUM_CASE(Pixel);
+    ENUM_CASE(Amplification);
+    ENUM_CASE(Mesh);
 #undef ENUM_CASE
   }
 };
--- a/include/Support/Pipeline.h
+++ b/include/Support/Pipeline.h
@@ -34,7 +34,7 @@ enum class Stages {
   Vertex,
   Pixel,
 
-  // Mesh Shader
+  // Mesh Shader Raster
   Amplification,
   Mesh
 };
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -57,16 +57,18 @@
 using namespace offloadtest;
 using Microsoft::WRL::ComPtr;
 
+using ID3D12DeviceX = ID3D12Device2;
+
 template <> char CapabilityValueEnum<directx::ShaderModel>::ID = 0;
 template <> char CapabilityValueEnum<directx::RootSignature>::ID = 0;
 template <> char CapabilityValueEnum<directx::MeshShaderTier>::ID = 0;
 
 static std::mutex SignalHandlerMutex;
-static llvm::SmallVector<ID3D12Device *> SignalHandlerDevices;
+static llvm::SmallVector<ID3D12DeviceX *> SignalHandlerDevices;
 
 static void dumpD3DInfoQueues(void *) {
   const std::lock_guard<std::mutex> Lock(SignalHandlerMutex);
-  for (ID3D12Device *Device : SignalHandlerDevices) {
+  for (ID3D12DeviceX *Device : SignalHandlerDevices) {
     ComPtr<ID3D12InfoQueue> InfoQueue;
     HRESULT HR = Device->QueryInterface(InfoQueue.GetAddressOf());
     if (FAILED(HR)) {
@@ -405,7 +407,7 @@ public:
   int Event;
 #endif
 
-  static llvm::Expected<std::unique_ptr<DXFence>> create(ID3D12Device *Device,
+  static llvm::Expected<std::unique_ptr<DXFence>> create(ID3D12DeviceX *Device,
                                                          llvm::StringRef Name) {
     ComPtr<ID3D12Fence> Fence;
     if (auto Err = HR::toError(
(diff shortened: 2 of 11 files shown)


include/Support/Pipeline.h
```diff
@@ -26,15 +26,25 @@
...
-enum class Stages { Compute, Vertex, Pixel };
+enum class Stages {
+  // Compute
+  Compute,
+
+  // Traditional Raster
+  Vertex,
+  Pixel,
+
+  // Mesh Shader
```
bob80905: Could you add Raster to the end of this comment?

lib/API/DX/Device.cpp
```diff
@@ -1143,6 +1152,90 @@ class DXDevice : public offloadtest::Device {
     return std::make_unique<DXPipelineState>(Name, RootSig, PSO);
   }
 
+  llvm::Expected<std::unique_ptr<PipelineState>>
+  createPipelineAsMsPs(llvm::StringRef Name, const BindingsDesc &BindingsDesc,
+                       llvm::ArrayRef<Format> RTFormats,
+                       std::optional<Format> DSFormat,
+                       std::optional<ShaderContainer> AS, ShaderContainer MS,
+                       std::optional<ShaderContainer> PS) /*override*/ {
```
alsepkow:
```suggestion
                       std::optional<ShaderContainer> PS) {
```

lib/API/DX/Device.cpp
```diff
@@ -2272,6 +2376,19 @@ class DXDevice : public offloadtest::Device {
       BindingsDesc.DescriptorSetDescs.push_back(Layout);
     }
 
+    if (P.isRaster()) {
+      // Create render target, depth/stencil, readback and vertex buffer and
```
bob80905: Where is the readback and vertex buffer and PSO creation? Should we remove this part of the comment?

lib/API/VK/Device.cpp
```diff
@@ -208,6 +208,10 @@ static VkShaderStageFlagBits getShaderStageFlag(Stages Stage) {
     return VK_SHADER_STAGE_VERTEX_BIT;
   case Stages::Pixel:
     return VK_SHADER_STAGE_FRAGMENT_BIT;
+  case Stages::Amplification:
```
bob80905, line 211: I wonder if we can change this enum definition to Stages::Task instead of Amplification, since IIUC this goes by the name "task" in vulkan contexts.

manon-traverse: In Vulkan, it is indeed called a Task Shader. However, in both DX12 and Metal, it's called an Amplification Shader. Since we seem to be using DX12 naming conventions (just like we do with Pixel Shader), I would recommend we use the name Amplification Shader for consistency.

manon-traverse:
It looks like I need to do some work for mesh shader support querying since not all our CI runners support mesh shaders.

include/Support/Pipeline.h
```diff
@@ -503,6 +513,12 @@ struct Pipeline {
   bool isTraditionalRaster() const {
     return Kind == ShaderPipelineKind::TraditionalRaster;
   }
+  bool isMeshShaderRaster() const {
```
MarijnS95, line 516: Not sure how to word this, but something about this name makes it hard to read, maybe that the other is* functions don't contain the word Shader

lib/API/DX/Device.cpp
```diff
@@ -1143,6 +1153,90 @@ class DXDevice : public offloadtest::Device {
...
+      Stream.DSVFormat = getDXGIFormat(*DSFormat);
+    Stream.SampleDesc = SampleDesc;
+
+    const D3D12_PIPELINE_STATE_STREAM_DESC StreamDesc = {sizeof(Stream),
+                                                         &Stream};
+
+    ComPtr<ID3D12Device2> Device2;
+    if (auto Err =
+            HR::toError(Device.As(&Device2), "Failed to query ID3D12Device2."))
+      return Err;
```
MarijnS95: I wonder if we could just store `ID3D12DeviceX` in the struct wherever X is the minimum bound that this framework can/needs to support.

MarijnS95 (approved):
https://github.com/llvm/offload-test-suite/actions/runs/26168261765/job/76978409258?pr=1175#step:13:524

```
PASS: OffloadTest-warp-d3d12 :: Graphics/MeshShaders/SimpleTriangle.test (372 of 492)
```

🎉

lib/API/DX/Device.cpp
manon-traverse: Yes I have added `using ID3D12DeviceX = ID3D12Device2;` thing so we can easily upgrade the device version in the future.

include/Support/Pipeline.h
manon-traverse: Yes, but `isMeshRaster` also doesn't feel like a good name. Do you have any suggestions for a better name?

EmilioLaiso (approved):

include/Support/Pipeline.h
```diff
@@ -26,15 +26,25 @@
...
+  // Mesh Shader Raster
+  Amplification,
+  Mesh
+};
 inline constexpr std::array AllStages = {
-    Stages::Compute,
-    Stages::Vertex,
-    Stages::Pixel,
+    Stages::Compute,       Stages::Vertex, Stages::Pixel,
+    Stages::Amplification, Stages::Mesh,
```
bogner, line 43: I guess this is just clang-format making strange choices (in which case it's fine to leave it) but this seems almost intentionally harder to read than one per line, especially considering it mirrors the definitions above.

lib/API/DX/Device.cpp
```diff
@@ -509,12 +512,12 @@ class DXQueue : public offloadtest::Queue {
 class DXCommandBuffer : public offloadtest::CommandBuffer {
 public:
   ComPtr<ID3D12CommandAllocator> Allocator;
-  ComPtr<ID3D12GraphicsCommandList> CmdList;
+  ComPtr<ID3D12GraphicsCommandList6> CmdList;
```
bogner: I'm not too familiar with the conventions for working with these versioned interfaces, but it seems weird to alias ID3D12Device2 but hardcode ID3D12GraphicsCommandList6 here.

test/Graphics/MeshShaders/SimpleTriangle.test
```diff
@@ -0,0 +1,82 @@
...
+...
+#--- rules.yaml
+---
+- Type: PixelPercent
+  Val: 0.2 # No more than 0.2% of pixels may be visibly different.
+...
+#--- end
+
+# REQUIRES: MeshShader
+# UNSUPPORTED: Vulkan || Metal || Clang
```
bogner: We should generally use XFAIL for things that aren't supported *yet*, so that the test will XPASS when we implement it and avoid us forgetting to update it. We also want separate XFAILs per reason and links to the issues where we intend to fix them, so something like:
```
# Unimplemented https://github.com/llvm/offload-test-suite/NNNN0
# XFAIL: Vulkan
# Unimplemented https://github.com/llvm/offload-test-suite/NNNN1
# XFAIL: Metal
# Unimplemented https://github.com/llvm/llvm-project/NNNN2
# XFAIL: Clang
```

where the NNNNs are the appropriate issues.

Edit: though, since we have the `REQUIRES: MeshShader`, which is always false for Vulkan and Metal at this point, we probably only really need the Clang XFAIL.

include/Support/Pipeline.h
manon-traverse: Yeah this is clang-format making a mess of things.

lib/API/DX/Device.cpp
manon-traverse: Main reason here is that we don't have as many `ID3D12GraphicsCommandList6` references, but I'll alias that one as well

test/Graphics/MeshShaders/SimpleTriangle.test
```diff
@@ -72,8 +72,8 @@ DescriptorSets: []
 ...
 #--- end
 
-# REQUIRES: MeshShader
-# UNSUPPORTED: Vulkan || Metal || Clang
+# Unimplemented https://github.com/llvm/llvm-project/issues/136966
+# XFAIL Clang
```
bogner: I think you accidentally dropped the `REQUIRES` here

manon-traverse: I did! I have added it back.

test/Graphics/MeshShaders/SimpleTriangle.test
manon-traverse: Strange, CI is failing the MeshShader test failing on clang even though it's marked `# XFAIL: Clang`.
Are we missing something or did I mess up?

manon-traverse: It looks like I missed a colon (`:`) 🤦‍♀️

bogner (approved):

PR #1176, opened 2026-05-12
Improve vulkan pipeline creation, better spec const parsing and error handling.

manon-traverse:
These are some fixes that I made while implementing mesh shader pipeline creation, which also apply to the other pipeline creation functions. Putting this in a separate PR to keep the work in smaller incremental changes.

### DIFF
--- a/lib/API/VK/Device.cpp
+++ b/lib/API/VK/Device.cpp
@@ -1629,17 +1629,11 @@ public:
                                                          1, &PipelineCI,
                                                          nullptr, &Pipeline),
                                "Failed to create graphics pipeline.")) {
-      for (auto *M : ShaderModules)
-        vkDestroyShaderModule(Device, M, nullptr);
       vkDestroyPipelineLayout(Device, PipelineLayout, nullptr);
       for (auto *L : SetLayouts)
         vkDestroyDescriptorSetLayout(Device, L, nullptr);
       return Err;
     }
-
-    // No longer need shader modules after pipeline compilation.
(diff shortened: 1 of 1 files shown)


EmilioLaiso (approved):
Nice cleanup! 🎉

lib/API/VK/Device.cpp
```diff
@@ -1373,6 +1373,21 @@ class VulkanDevice : public offloadtest::Device {
   llvm::Expected<std::unique_ptr<PipelineState>>
   createPipelineCs(llvm::StringRef Name, const BindingsDesc &BindingsDesc,
                    ShaderContainer CS) override {
+    auto CSModOrErr = createShaderModule(CS.Shader, "compute");
+    if (!CSModOrErr)
+      return CSModOrErr.takeError();
+
+    VkShaderModule CSModule = *CSModOrErr;
+    auto ShaderModuleCleanUp = llvm::scope_exit(
+        [&] { vkDestroyShaderModule(Device, CSModule, nullptr); });
```
MarijnS95, line 1383: Retaining your own comment about destroying these even when the PSO is successfully created would be nice.


```suggestion
     // No longer need shader modules after pipeline compilation.
    auto ShaderModuleCleanUp = llvm::scope_exit(
        [&] { vkDestroyShaderModule(Device, CSModule, nullptr); });
```

MarijnS95 (approved):

manon-traverse:
Merging this before CI finished, because only a comment was added since last CI completion.

PR #1177, opened 2026-05-12
Add Mesh Shader Support to Vulkan

manon-traverse:
Add Mesh Shader support to Vulkan.
Graphics/MeshShaders/SimpleTriangle.test now passes on Vulkan.

### DIFF
--- a/lib/API/VK/Device.cpp
+++ b/lib/API/VK/Device.cpp
@@ -1279,11 +1279,9 @@ public:
       MeshFeatures.multiviewMeshShader = 0;
       MeshFeatures.primitiveFragmentShadingRateMeshShader = 0;
       MeshFeatures.meshShaderQueries = 0;
-#ifdef VK_VERSION_1_4
-      Features14.pNext = &MeshFeatures;
-#else
-      Features13.pNext = &MeshFeatures;
-#endif
+
+      MeshFeatures.pNext = Features.pNext;
+      Features.pNext = &MeshFeatures;
     }
 
     DeviceInfo.enabledExtensionCount =
--- a/lib/API/VK/Device.cpp
+++ b/lib/API/VK/Device.cpp
@@ -1916,17 +1916,11 @@ public:
     for (const Format F : RTFormats) {
       ColorAttachmentFormatDesc CA = {};
       CA.Fmt = F;
-      CA.Load = LoadAction::DontCare;
-      CA.Store = StoreAction::DontCare;
       PassDesc.ColorAttachments.push_back(CA);
     }
     if (DSFormat) {
       DepthStencilAttachmentFormatDesc DS = {};
       DS.Fmt = *DSFormat;
-      DS.DepthLoad = LoadAction::DontCare;
-      DS.DepthStore = StoreAction::DontCare;
-      DS.StencilLoad = LoadAction::DontCare;
-      DS.StencilStore = StoreAction::DontCare;
       PassDesc.DepthStencil = DS;
     }
 
--- a/lib/API/VK/Device.cpp
+++ b/lib/API/VK/Device.cpp
@@ -406,6 +406,18 @@ struct VulkanInstance {
 
 namespace {
 
+struct MeshShaderFunctions {
+  PFN_vkCmdDrawMeshTasksEXT VkCmdDrawMeshTasksEXT = nullptr;
+
+  static MeshShaderFunctions create(VkDevice Device) {
+    MeshShaderFunctions Result;
+    Result.VkCmdDrawMeshTasksEXT =
+        (PFN_vkCmdDrawMeshTasksEXT)vkGetDeviceProcAddr(Device,
+                                                       "vkCmdDrawMeshTasksEXT");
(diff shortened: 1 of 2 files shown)


EmilioLaiso (approved):

lib/API/VK/Device.cpp
```diff
@@ -586,7 +599,8 @@ class VulkanCommandBuffer : public offloadtest::CommandBuffer {
   create(VkDevice Device, uint32_t QueueFamilyIdx,
          PFN_vkCmdBeginDebugUtilsLabelEXT CmdBeginDebugUtilsLabel,
          PFN_vkCmdEndDebugUtilsLabelEXT CmdEndDebugUtilsLabel,
-         PFN_vkCmdInsertDebugUtilsLabelEXT CmdInsertDebugUtilsLabel) {
+         PFN_vkCmdInsertDebugUtilsLabelEXT CmdInsertDebugUtilsLabel,
```
MarijnS95, line 608: Should we design a `DebugFunctions` struct for these too (in a separate PR)?

And do you recommend me to follow the same pattern in https://github.com/llvm/offload-test-suite/pull/1232?

lib/API/VK/Device.cpp
```diff
@@ -1236,6 +1258,31 @@ class VulkanDevice : public offloadtest::Device {
...
+      MeshFeatures.taskShader = 1;
+      MeshFeatures.meshShader = 1;
+      MeshFeatures.multiviewMeshShader = 0;
+      MeshFeatures.primitiveFragmentShadingRateMeshShader = 0;
+      MeshFeatures.meshShaderQueries = 0;
+#ifdef VK_VERSION_1_4
+      Features14.pNext = &MeshFeatures;
+#else
+      Features13.pNext = &MeshFeatures;
+#endif
```
MarijnS95: `VK_EXT_mesh_shader` only requires Vulkan 1.2. So technically if `Props.apiVersion` isn't higher than that, neither of these feature structs might have been appended, only `Features12`?

This should instead find a stable field to add or insert this struct into.

Likewise, the `ifdef` isn't enough here, we need to distinguish via `apiVersion` whether to overwrite the `pNext` of `Features13` or `Features14` too.

lib/API/VK/Device.cpp
```diff
@@ -1707,6 +1756,205 @@ class VulkanDevice : public offloadtest::Device {
...
+    }
+
+    // Build a RenderPassDesc from the PSO's RT/DS formats.
+    RenderPassDesc PassDesc;
+    PassDesc.ColorAttachments.reserve(RTFormats.size());
+    for (const Format F : RTFormats) {
+      ColorAttachmentFormatDesc CA = {};
+      CA.Fmt = F;
+      CA.Load = LoadAction::DontCare;
+      CA.Store = StoreAction::DontCare;
```
MarijnS95: I wonder if we should just delete these lines (in the other PSO create functions as well) because the struct is already default-initialized and these fields are ignored per the comment below. Initializing them to a specific value does cause a slight bit of confusion again.

lib/API/VK/Device.cpp
```diff
@@ -1707,6 +1756,205 @@ class VulkanDevice : public offloadtest::Device {
...
+    PipelineCI.pRasterizationState = &RastCI;
+    PipelineCI.pMultisampleState = &MultisampleCI;
+    PipelineCI.pDepthStencilState = &DepthStencilCI;
+    PipelineCI.pColorBlendState = &BlendCI;
+    PipelineCI.pDynamicState = &DynamicCI;
+    PipelineCI.layout = PipelineLayout;
+    PipelineCI.renderPass = RenderPassHandle;
+
+    VkPipeline Pipeline = VK_NULL_HANDLE;
+    if (auto Err = VK::toError(vkCreateGraphicsPipelines(Device, VK_NULL_HANDLE,
```
MarijnS95, line 2003: Curious: did we previously factor out a helper for all the duplicate `VkGraphicsPipelineCreateInfo` state, that is the same for every graphics pipeline?

lib/API/VK/Device.cpp
manon-traverse: Yes and yes

lib/API/VK/Device.cpp
manon-traverse: I don't see how this creates confusion? What are you confused about?

lib/API/VK/Device.cpp
manon-traverse: I don't think we did?

lib/API/VK/Device.cpp
MarijnS95: Initializing fields that are not used, with a specific value.

manon-traverse: talked offline, understand now. Leaving the fields on default because they don't matter in this use case

MarijnS95 (approved):

manon-traverse:
CI only fails on test irrelevant to this PR, so I feel safe to merge this.

PR #1184, opened 2026-05-13
Add InterlockedAnd tests

### LINKED ISSUE #849: Add test for `InterlockedAnd`
- [ ] Test for 16 bit int types (`InterlockedAnd.int16.test`)
- [ ] Test for Half type (`InterlockedAnd.fp16.test`)
- [ ] Test for 32 bit types (`InterlockedAnd.32.test`)
- [ ] Test for 64 bit int types (`InterlockedAnd.int64.test`)
- [ ] Test for Double type (`InterlockedAnd.fp64.test`)

A .test file should contain 
```
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# RUN: split-file %s %t
```

bob80905:
This PR adds tests for `InterlockedAnd`
Adds 32 bit and 64 bit signed/unsigned integer tests
Fixes https://github.com/llvm/offload-test-suite/issues/849

### DIFF
--- a/test/Feature/HLSLLib/InterlockedAnd.int64.test
+++ b/test/Feature/HLSLLib/InterlockedAnd.int64.test
@@ -128,6 +128,9 @@ DescriptorSets:
 ...
 #--- end
 
+# Unimplemented: https://github.com/llvm/llvm-project/issues/99125
+# XFAIL: Clang
+
 # REQUIRES: Int64
 # RUN: split-file %s %t
 # RUN: %dxc_target -HV 202x -T cs_6_6 -Fo %t.o %t/source.hlsl
--- a/test/Feature/HLSLLib/InterlockedAnd.int64.test
+++ b/test/Feature/HLSLLib/InterlockedAnd.int64.test
@@ -84,7 +84,7 @@ Buffers:
   - Name: OutOrigBitSet
     Format: UInt32
     Stride: 4
-    ZeroInitSize: 256
+    FillSize: 256
   - Name: ExpectedOrigBitSet
     Format: UInt32
     Stride: 4
@@ -95,7 +95,7 @@ Buffers:
   - Name: OutFinal
     Format: UInt64
     Stride: 8
-    ZeroInitSize: 32
+    FillSize: 32
   - Name: ExpectedFinal
     Format: UInt64
     Stride: 8
--- a/test/Feature/HLSLLib/InterlockedAnd.int64.test
+++ b/test/Feature/HLSLLib/InterlockedAnd.int64.test
@@ -1,0 +1,135 @@
+#--- source.hlsl
+
+// This test exercises InterlockedAnd against non-resource (groupshared)
+// destinations using 64-bit integer types. A single threadgroup of 64
+// threads concurrently updates shared counters, so the test actually
+// exercises atomic behavior.
+//
+// Both the 2-argument and 3-argument overloads are covered for int64_t
+// and uint64_t.
+//
+// Atomicity is verified by starting a counter at 0xFFFFFFFFFFFFFFFF and
+// having each of 64 threads atomically clear its own unique bit
+// (AND with ~(1ull << tid)). If any read-modify-write were non-atomic,
+// some thread's bit clear would be lost and the final value would be
+// non-zero. With true atomicity, the counter must end at exactly 0.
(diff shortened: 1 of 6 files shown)


test/Feature/HLSLLib/InterlockedAnd.resources.typed.int64.test
```diff
@@ -0,0 +1,130 @@
+#--- source.hlsl
+
+// This test exercises InterlockedAnd on 64-bit typed RWBuffer resources.
+// 64-bit atomics on typed UAVs are an optional Shader Model 6.6 capability
+// (D3D12 cap AtomicInt64OnTypedResourceSupported, Vulkan feature
+// shaderBufferInt64Atomics), surfaced to lit as `Int64TypedResourceAtomics`.
```
Icohedron: https://github.com/llvm/offload-test-suite/pull/1195#discussion_r3276982841

Icohedron (changes requested):

test/Feature/HLSLLib/InterlockedAnd.int64.test
```diff
@@ -0,0 +1,137 @@
...
+        Space: 0
+      VulkanBinding:
+        Binding: 1
+...
+#--- end
+
+# Unimplemented: https://github.com/llvm/llvm-project/issues/99125
+# XFAIL: Clang
+
+# REQUIRES: Int64
```
Icohedron: I noticed in this PR you are not adding a `REQUIRES: SM_6_6` but in your InterlockedOr PR you are.
```suggestion
# REQUIRES: Int64 && SM_6_6
```

Icohedron: There also should be a `REQUIRES: Int64GroupSharedAtomics` since you're using `int64` groupshared variables.

test/Feature/HLSLLib/InterlockedAnd.int64.test
```diff
@@ -0,0 +1,140 @@
...
+      VulkanBinding:
+        Binding: 1
+...
+#--- end
+
+# Unimplemented: https://github.com/llvm/llvm-project/issues/99125
+# XFAIL: Clang
+
+# Bug: https://github.com/llvm/offload-test-suite/issues/1164
+# XFAIL: Metal && DXC
```
inbelic, line 135: I don't understand why this is required here but not in the 32 bit cases. The tests seem the same?

inbelic (commented):
Mostly LGTM

test/Feature/HLSLLib/InterlockedAnd.int64.test
alsepkow: @bob80905 gentle nudge

Icohedron (approved):

test/Feature/HLSLLib/InterlockedAnd.int64.test
bob80905: Ran through copilot, I agree:

 - The REQUIRES ensures the test only runs on platforms that claim to support 64-bit groupshared atomics.
 - The XFAIL: Metal && DXC is still needed because Metal advertises the capability but has a runtime bug (issue 1164) causing it to fail anyway.
 - The 32-bit test doesn't need either because 32-bit atomics work correctly everywhere.

inbelic (approved):

PR #1185, opened 2026-05-13
Add runner status script

### LINKED ISSUE #1227: [Quality of Life] Make an easier way to track runner status
Currently, there is no way to get a compact, comprehensive overview of the runner status for the offload test suite.
To understand how the test infra distributes jobs, and the degree of congestion, it would be helpful to have a tool that communicates the current runner status.
This is also helpful to know whether build jobs are being well-distributed across compatible SKUs.

bob80905:
This script is handy for looking at the workload of the machines for the offload test suite.
It might be especially helpful in the near future, when the split build/test feature is commonplace.
Presently, it displays which jobs are being allocated (Donev,  Active,  Queued) to which machine SKUs.
It should, in the future, display "build job" and "test job" allocations, for greater clarity.
It accepts one optional argument, to scope the output to a specific SKU, but leave it empty to get status on all SKUs.

For example, here's the output of this command:
image

Fixes https://github.com/llvm/offload-test-suite/issues/1227
Assisted by Github Copilot


tools/runner_status.py
```diff
@@ -0,0 +1,243 @@
...
+    return True
+
+
+def main():
+    if len(sys.argv) < 2:
+        print(f"Usage: python {os.path.basename(__file__)} <GITHUB_TOKEN> [vendor]")
+        print(f"  vendor: {' | '.join(VALID_VENDORS)}  (omit to show all)")
+        sys.exit(1)
+
+    token = sys.argv[1]
```
damyanp: Suggest making the github token be stored in an environment variable rather than passed on the command line as this encourages insecure behaviors where you might end up accidentally including your PAT in a screen shot.

damyanp: Alternatively, you could probably take a dependency on the "gh" command line tool and use `gh auth token` to aquire a token.  In fact, you could use `gh api` to issue the github API calls.

damyanp:
The PR description needs to be updated to match the new usage.

damyanp:
While this LGTM, a couple of questions:
* I'm not sure of the right bar for helper scripts like this
* Is "tools" the right location for it? Would somewhere under .github be better?

bob80905:
> While this LGTM, a couple of questions:
> 
> * I'm not sure of the right bar for helper scripts like this
> * Is "tools" the right location for it? Would somewhere under .github be better?

Sure, I think .github would be better.
Not sure what the bar is either, but I've had this script laying around for a while and have used it pretty frequently myself.
I imagine it'd be useful for folks to know how loaded a machine is, to estimate when their PR's workflows will be done, or to know if the machine that is still pending is being bogged down by scheduled workflows.
Might also somewhat inform us of which machines are slower / could use an additional SKU.
But in my opinion, the best use case for me is to see which machine takes the "build job" for a specific PR, it might be useful data to make sure build jobs are being well distributed.

alsepkow (approved):
LGTUS @inbelic @Icohedron 

Image

PR #1187, opened 2026-05-13
[CI] Promote NVIDIA DirectX target out of experimental

alsepkow:
NVIDIA's DirectX configuration has been stable and has a higher pass rate
than the existing Tier 1 Intel target. Promote it to Tier 1 so it runs on
every PR. AMD, Qualcomm, and the Vulkan IHV configurations remain
experimental and continue to require the `test-all` label.

AMD was originally part of this promotion but is being excluded for now
while we debug some intermittent bugs that were noticed.

### Changes

- `docs/CI.md`: Promote `Windows NVIDIA GPU DirectX` from Experimental to
  Tier 1.
- `README.md`: Move the NVIDIA DirectX row from Experimental Targets to
  Tier 1 Targets in the status table.
- `.github/workflows/pr-matrix.yaml`: Add `windows-nvidia` to the existing
  `Exec-Tests-Windows` matrix (excluding the still-experimental Vulkan
  combinations). `Exec-Tests-Extra` (gated on `test-all`) continues to
  cover AMD D3D12, AMD/NVIDIA Vulkan, and all QC targets.

---
Assisted by Claude Opus 4.7


farzonl (approved):

farzonl:
Thank you Alex!

.github/workflows/pr-matrix.yaml
```diff
@@ -22,6 +22,15 @@ jobs:
       matrix:
         SKU: [windows-intel]
         TestTarget: [check-hlsl-d3d12, check-hlsl-vk, check-hlsl-clang-d3d12, check-hlsl-clang-vk]
+        include:
```
bob80905: nit: 
You might consider something like this:
```
matrix:
   SKU: [windows-intel, windows-amd, windows-nvidia]
   TestTarget: [check-hlsl-d3d12, check-hlsl-vk, check-hlsl-clang-d3d12, check-hlsl-clang-vk]
   exclude:
     - { SKU: windows-amd,    TestTarget: check-hlsl-vk }
     - { SKU: windows-amd,    TestTarget: check-hlsl-clang-vk }
     - { SKU: windows-nvidia, TestTarget: check-hlsl-vk }
     - { SKU: windows-nvidia, TestTarget: check-hlsl-clang-vk }
```
so that as future test targets get promoted, we can just remove exclusions, rather than needing to add specific inclusions. Something about this seems cleaner to me.

bob80905 (approved):
LGTM

.github/workflows/pr-matrix.yaml
alsepkow: This does seem cleaner to me assuming the expected state would eventually be that we don't have exclusions.

bob80905 (approved):
Bob approves 😋

bob80905 (approved):

.github/workflows/pr-matrix.yaml
```diff
@@ -57,8 +60,27 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        SKU: [windows-nvidia, windows-amd, windows-qc]
-        TestTarget: [check-hlsl-d3d12, check-hlsl-vk, check-hlsl-clang-d3d12, check-hlsl-clang-vk]
+        include:
```
damyanp: Why has this been expanded out, rather than using the matrix and exclusions approach used above?  Presumably it'd be preferable to use the matrix/exclude pattern?

damyanp (commented):
I don't think that the PR description matches what the PR actually does?

.github/workflows/pr-matrix.yaml
alsepkow: Came up in discussion with @bob80905 that was that the includes option makes for cleaner looking diffs if we're doing partial promotions. And that partial promotions have typically been the case.

If that weren't the case than sticking with excludes makes more sense. I'm indifferent. Do we have a pattern we typically follow?

damyanp: Previously we seemed to use the matrix and have excludes.  The matrix + excludes IMO specifies intent much more clearly.  I need to reconstruct the matrix to understand what's going on with the explicit include list.

bogner (approved):
LGTM. For the commit message/PR description I don't think the recent pass rates, the fact that we were considering AMD as well here, or the list of what the mechanical changes are are useful. IMO we just need the blurb at the beginning to justify this change and provide sufficient historical context:

> NVIDIA's DirectX configuration has been stable and has a higher pass rate
> than the existing Tier 1 Intel target. Promote it to Tier 1 so it runs on
> every PR. AMD, Qualcomm, and the Vulkan IHV configurations remain
> experimental and continue to require the test-all label.

bob80905 (approved):

PR #1195, opened 2026-05-14
Add InterlockedOr tests

### LINKED ISSUE #101: Add test for InterlockedOr
> Implementation issue for `InterlockedOr`: https://github.com/llvm/llvm-project/issues/99126

> Note: we expect that this test will actually exercise multiple threads / waves / threadgroups so we can see actual atomic behavior.
>
> Next step is enumerating what types of memory this operation can touch - group shared? buffers? etc.

A .test file should contain 
```
@@ -0,0 +1,54 @@
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# UNSUPPORTED: Clang-Vulkan
# RUN: split-file %s %t
```

bob80905:
This PR adds tests for the InterlockedOr HLSL function.
It tests all existing overloads of this function, including the declarations of this function as members of resources.
It also includes a fix to texture element type processing, as there was not a case to account for uint32 element types.
Fixes https://github.com/llvm/offload-test-suite/issues/101


test/Feature/HLSLLib/InterlockedOr.resources.typed.int64.test
```diff
@@ -0,0 +1,129 @@
+#--- source.hlsl
+
+// This test exercises InterlockedOr on 64-bit typed RWBuffer resources.
+// 64-bit atomics on typed UAVs are an optional Shader Model 6.6 capability
+// (D3D12 cap AtomicInt64OnTypedResourceSupported, Vulkan feature
+// shaderBufferInt64Atomics), surfaced to lit as `Int64TypedResourceAtomics`.
```
Icohedron: This statement about `shaderBufferInt64Atomics` for Vulkan with RWBuffer resources is incorrect.
RWBuffers are actually images: https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst#rwbuffer
So you would need the `shaderImageInt64Atomics` Vulkan feature instead.
And unlike `shaderBufferInt64Atomics`, `shaderImageInt64Atomics` did not become a core Vulkan feature and is currently only available as an extension: [VK_EXT_shader_image_atomic_int64](https://docs.vulkan.org/guide/latest/atomics.html#VK_EXT_shader_image_atomic_int64)

Icohedron (changes requested):

test/Feature/HLSLLib/InterlockedOr.resources.int64.test
```diff
@@ -0,0 +1,182 @@
...
+# Unimplemented: https://github.com/llvm/llvm-project/issues/99126
+# XFAIL: Clang
+
+# Bug: https://github.com/llvm/offload-test-suite/issues/1164
+# XFAIL: Metal && DXC
+
+# Unimplemented: https://github.com/microsoft/DirectXShaderCompiler/issues/8451
+# XFAIL: Vulkan && DXC
+
+# REQUIRES: Int64 && SM_6_6
```
Icohedron: I think it would be a good idea to add a new Vulkan-specific feature flag to this REQUIRES called `VulkanInt64BufferAtomics` which is true iff the [VK_KHR_shader_atomic_int64](https://docs.vulkan.org/guide/latest/atomics.html#VK_KHR_shader_atomic_int64) Vulkan extension is enabled or the Vulkan version is greater than or equal to 1.2.

This is because the `RWStructuredBuffer<uint64_t>` and `RWStructuredBuffer<int64_t>` create storage buffers (SSBOs) of 64-bit integers, and you are performing atomic operations on them; therefore you require `VK_KHR_shader_atomic_int64` or Vulkan 1.2+

Specifically, you need `shaderBufferInt64Atomics` from Vulkan only. You don't need `AtomicInt64OnTypedResourceSupported` in DirectX.

test/Feature/HLSLLib/InterlockedAdd.int64.test
```diff
@@ -152,7 +152,7 @@ DescriptorSets:
 ...
 #--- end
 
-# REQUIRES: Int64 && Int64GroupSharedAtomics
+# REQUIRES: Int64 && Int64GroupSharedAtomics && SM_6_6
```
Icohedron, line 155: Interesting change. Do we want to add `REQUIRES: SM_6_X` to every test?

Every test has a `# RUN: %dxc_target -HV 202x -T cs_6_6 -Fo %t.o %t/source.hlsl` line specifying the target shader model (`cs_6_6` implies `SM_6_6`).

bob80905: It's definitely up for conversation. The reason I added it is because there was a machine I ran against that did not have shader model 6.6 support. So this test shouldn't run on machines that don't have GPUs that support high enough shader models.
Here is the failure: https://github.com/llvm/offload-test-suite/actions/runs/25931189782/job/76225188838#step:13:338
Its directly complaining about shader model target incompatibility, so I think the proper resolution is to add these availability features.

farzonl: > Interesting change. Do we want to add `REQUIRES: SM_6_X` to every test?
> 
> Every test has a `# RUN: %dxc_target -HV 202x -T cs_6_6 -Fo %t.o %t/source.hlsl` line specifying the target shader model (`cs_6_6` implies `SM_6_6`).

Hmm I hear why you did this. I'm wondering if there is a way we can unify the run line and the requires. It would  suck if these fell out of sync. Maybe a cleanup task for later?

bob80905: Made a follow up issue here: https://github.com/llvm/offload-test-suite/issues/1219

.github/workflows/windows-nvidia-clang-vk.yaml
```diff
@@ -21,5 +21,9 @@ jobs:
       OS: windows
       SKU: windows-nvidia
       TestTarget: check-hlsl-clang-vk
-      OffloadTest-branch: ${{ github.ref }}
+      OffloadTest-branch: fix-test-job-regen
+      OffloadTest-fork: bob80905
+      LLVM-branch: hlsl-offload-distribution
+      LLVM-fork: bob80905
+      SplitBuild: true
```
Icohedron: I think this is the wrong branch

Icohedron (approved):

alsepkow:
LGTM given that we've filed an issue to follow up on @farzonl comment.

PR #1203, opened 2026-05-18
Add geometry shader support with point -> quad test

EmilioLaiso:
Add geometry shader support to the engine and introduce `PrimitiveTopology` type.
`PrimitiveTopology` defaults to `TriangleList` so that we don't have to specify it in all our existing rasterization tests.

Add a new test `Graphics/gs_point_to_quad.test` exercising a VS → GS → PS pipeline: 
- VS is a pass-through
- GS expands a single input point into a fullscreen quad
- PS writes red. 
- A 2x2 render target verifies that all four pixels are rasterized.

Add `GraphicsPipelineCreateDesc` to keep the pipeline creation function signature tidy.
`GS` and `PS` within `GraphicsPipelineCreateDesc` are optional, but for the time being the backends assert that a PS is set. We can lift that assert once we need a VS-only test.

### DIFF
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -27,6 +27,7 @@
 #include "llvm/ADT/StringRef.h"
 #include "llvm/ADT/iterator_range.h"
 #include "llvm/Support/Error.h"
+#include "llvm/Support/ErrorHandling.h"
 #include "llvm/Support/MemoryBuffer.h"
 
 #include <cstdint>
@@ -79,6 +80,33 @@ struct ShaderContainer {
   std::string EntryPoint;
   const llvm::MemoryBuffer *Shader;
   llvm::SmallVector<SpecializationConstant> SpecializationConstants;
+};
+
+struct GraphicsPipelineCreateDesc {
+  llvm::SmallVector<InputLayoutDesc> InputLayout;
+  llvm::SmallVector<Format> RTFormats;
+  std::optional<Format> DSFormat;
+  PrimitiveTopology Topology;
+  ShaderContainer VS;
+  // TODO: Optional Hull & Domain Shaders
+  std::optional<ShaderContainer> GS;
+  ShaderContainer PS;
+
+  void setShader(Stages Stage, ShaderContainer &&SC) {
+    switch (Stage) {
+    case Stages::Vertex:
+      VS = std::move(SC);
+      break;
+    case Stages::Geometry:
+      GS = std::move(SC);
+      break;
+    case Stages::Pixel:
+      PS = std::move(SC);
+      break;
+    case Stages::Compute:
+      llvm_unreachable("Creating graphics pipeline for compute?");
+    }
+  }
 };
 
 class PipelineState {
@@ -160,11 +188,9 @@ public:
                    ShaderContainer CS) = 0;
 
   virtual llvm::Expected<std::unique_ptr<PipelineState>>
-  createPipelineVsPs(llvm::StringRef Name, const BindingsDesc &BindingsDesc,
-                     llvm::ArrayRef<InputLayoutDesc> InputLayout,
-                     llvm::ArrayRef<Format> RTFormats,
-                     std::optional<Format> DSFormat, ShaderContainer VS,
-                     ShaderContainer PS) = 0;
+  createTraditionalRasterPipeline(llvm::StringRef Name,
+                                  const BindingsDesc &BindingsDesc,
+                                  const GraphicsPipelineCreateDesc &Desc) = 0;
 
   virtual llvm::Expected<std::unique_ptr<Fence>>
   createFence(llvm::StringRef Name) = 0;
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -82,7 +82,7 @@ struct ShaderContainer {
   llvm::SmallVector<SpecializationConstant> SpecializationConstants;
 };
 
-struct GraphicsPipelineCreateDesc {
+struct TraditionalRasterPipelineCreateDesc {
   llvm::SmallVector<InputLayoutDesc> InputLayout;
   llvm::SmallVector<Format> RTFormats;
   std::optional<Format> DSFormat;
@@ -190,9 +190,9 @@ public:
                    ShaderContainer CS) = 0;
 
   virtual llvm::Expected<std::unique_ptr<PipelineState>>
-  createTraditionalRasterPipeline(llvm::StringRef Name,
-                                  const BindingsDesc &BindingsDesc,
-                                  const GraphicsPipelineCreateDesc &Desc) = 0;
+  createTraditionalRasterPipeline(
+      llvm::StringRef Name, const BindingsDesc &BindingsDesc,
+      const TraditionalRasterPipelineCreateDesc &Desc) = 0;
 
   virtual llvm::Expected<std::unique_ptr<Fence>>
   createFence(llvm::StringRef Name) = 0;
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -104,7 +104,9 @@ struct GraphicsPipelineCreateDesc {
       PS = std::move(SC);
       break;
     case Stages::Compute:
-      llvm_unreachable("Creating graphics pipeline for compute?");
+    case Stages::Amplification:
+    case Stages::Mesh:
+      llvm_unreachable("Not a traditional raster pipeline stage.");
     }
   }
 };
--- a/include/Support/Pipeline.h
+++ b/include/Support/Pipeline.h
@@ -32,6 +32,7 @@ enum class Stages {
 
   // Traditional Raster
   Vertex,
+  Geometry,
   Pixel,
 
   // Mesh Shader Raster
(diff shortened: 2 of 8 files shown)


lib/Support/Pipeline.cpp
```diff
@@ -602,6 +604,10 @@ llvm::Error offloadtest::Pipeline::validatePipelineKind() {
     return llvm::Error::success();
   }
 
+  if (HasShaderType[llvm::to_underlying(Stages::Geometry)])
+    return llvm::createStringError(
+        "Pipeline has a Geometry Shader but no Vertex Shader.");
+
```
manon-traverse: The code below already handles the case of a missing vertex shader.

include/Support/Pipeline.h
```diff
@@ -26,16 +26,19 @@
...
     Stages::Vertex,
+    Stages::Geometry,
     Stages::Pixel,
 };
 inline constexpr size_t NumStages = AllStages.size();
 
 enum class ShaderPipelineKind { Compute, TraditionalRaster };
 
+enum class PrimitiveTopology { TriangleList, PointList };
+
```
manon-traverse, line 49: Please add this type to the rendering API instead of the test pipeline definition. Ideally, the rendering backend doesn't depend on the test pipelines at all. We are not at that point yet, but we should keep moving in this direction.

lib/API/MTL/MTLDevice.cpp
```diff
@@ -1304,11 +1304,23 @@ class MTLDevice : public offloadtest::Device {
...
+                         const GraphicsPipelineCreateDesc &Desc) override {
+    if (Desc.GS)
+      return llvm::createStringError(
+          std::errc::not_supported,
+          "Metal: Geometry shaders are not supported on this backend.");
+    if (!Desc.PS)
+      return llvm::createStringError(std::errc::invalid_argument,
+                                     "Graphics pipeline requires a pixel "
+                                     "shader on this backend.");
+    (void)Desc.Topology;
```
manon-traverse: No point in having this line as far as I can tell.

manon-traverse (approved):
Really nice changes! Awesome that we now have geometry shader support :D

lib/API/DX/Device.cpp
```diff
@@ -2305,43 +2342,42 @@ class DXDevice : public offloadtest::Device {
...
-        }
+        ShaderContainer SC = {};
+        SC.EntryPoint = Shader.Entry;
+        SC.Shader = Shader.Shader.get();
+        if (Shader.Stage == Stages::Vertex)
+          PipelineDesc.VS = std::move(SC);
+        else if (Shader.Stage == Stages::Geometry)
+          PipelineDesc.GS = std::move(SC);
+        else if (Shader.Stage == Stages::Pixel)
+          PipelineDesc.PS = std::move(SC);
```
bogner: It's a bit more verbose but it's probably better to use a fully covered switch statement here, so that adding values to the `Stages` enumerator would trigger a warning if we forget to update this.

```suggestion
        switch (Shader.Stage) {
        case Stages::Vertex:
          PipelineDesc.VS = std::move(SC);
          break;
        case Stages::Geometry:
          PipelineDesc.GS = std::move(SC);
          break;
        case Stages::Pixel:
          PipelineDesc.PS = std::move(SC);
          break;
        case Stages::Compute:
          llvm_unreachable("Creating graphics pipeline for compute?");
        }
```

Also it might be reasonable to put this as a helper function on the `GraphicsPipelineCreateDesc` struct, like `void setDescForStage(Stages Stage, ShaderContainer &&SC)` or so, so that we don't need to repeat this boilerplate in all three APIs.

lib/API/MTL/MTLDevice.cpp
```diff
@@ -1546,11 +1548,26 @@ class MTLDevice : public offloadtest::Device {
...
+  createGraphicsPipeline(llvm::StringRef Name, const BindingsDesc &BindingsDesc,
+                         const GraphicsPipelineCreateDesc &Desc) override {
+    if (Desc.GS)
+      return llvm::createStringError(
+          std::errc::not_supported,
+          "Metal: Geometry shaders are not supported on this backend.");
+    if (!Desc.PS)
+      return llvm::createStringError(
+          std::errc::invalid_argument,
+          "Graphics pipeline requires a pixel shader on this backend.");
```
bogner: Should this be prefixed with "Metal:" like the other two error messages here? I suppose if we do that we should update the Vulkan version of the same message to be prefixed "Vulkan:". Alternatively I suppose we could drop the prefixes altogether.

I'm fine with any of these options but would prefer to be consistent.

lib/API/MTL/MTLDevice.cpp
```diff
@@ -1756,42 +1773,41 @@ class MTLDevice : public offloadtest::Device {
...
-          PS.EntryPoint = Shader.Entry;
-          PS.Shader = Shader.Shader.get();
-        }
+        ShaderContainer SC = {};
+        SC.EntryPoint = Shader.Entry;
+        SC.Shader = Shader.Shader.get();
+        if (Shader.Stage == Stages::Vertex)
+          PipelineDesc.VS = std::move(SC);
+        else if (Shader.Stage == Stages::Geometry)
+          PipelineDesc.GS = std::move(SC);
```
bogner: Technically this is unreachable, but given that this is the only thing that could possibly make sense here it's probably fine as is.

include/API/Device.h
```diff
@@ -160,11 +171,8 @@ class Device {
                    ShaderContainer CS) = 0;
 
   virtual llvm::Expected<std::unique_ptr<PipelineState>>
-  createPipelineVsPs(llvm::StringRef Name, const BindingsDesc &BindingsDesc,
-                     llvm::ArrayRef<InputLayoutDesc> InputLayout,
-                     llvm::ArrayRef<Format> RTFormats,
-                     std::optional<Format> DSFormat, ShaderContainer VS,
-                     ShaderContainer PS) = 0;
+  createGraphicsPipeline(llvm::StringRef Name, const BindingsDesc &BindingsDesc,
+                         const GraphicsPipelineCreateDesc &Desc) = 0;
```
manon-traverse: Could you rename this to `createTraditionalRasterPipeline`?
Mesh Shader are also graphics (and so is ray tracing?)

test/Graphics/gs_point_to_quad.test
```diff
@@ -0,0 +1,101 @@
...
+
+# Metal has no native geometry shader stage (equivalents would require mesh or
+# tessellation shaders). Skip this test on that backend entirely.
+# UNSUPPORTED: Metal
+
+# Clang's HLSL frontend does not yet support geometry shader stream types
+# (e.g. TriangleStream<T>::Append), so compiling geometry.hlsl with clang-dxc
+# fails on both D3D12 and Vulkan code paths.
+# Tracked upstream: https://github.com/llvm/llvm-project/issues/136963
+# XFAIL: Clang
```
bogner: We've generally standardized to labeling the corresponding issues for XFAIL with "Bug" or "Unimplemented"

```suggestion
# Clang's HLSL frontend does not yet support geometry shader stream types
# (e.g. TriangleStream<T>::Append), so compiling geometry.hlsl with clang-dxc
# fails on both D3D12 and Vulkan code paths.
# Unimplemented https://github.com/llvm/llvm-project/issues/136963
# XFAIL: Clang
```

It's also fine to omit the comment explaining it, the issue itself should be clear enough.

bogner (approved):
A few minor comments but otherwise LGTM!

lib/API/MTL/MTLDevice.cpp
MarijnS95: It's optional on MTL too: https://developer.apple.com/documentation/metal/mtlrenderpipelinedescriptor/fragmentfunction

MarijnS95 (approved):
Once PS is optional on Metal again, this has a thumbs-up from me!

lib/API/MTL/MTLDevice.cpp
EmilioLaiso: I'm aware, currently the error checking is done on all backends. I'll remove it, wanted to keep it around until we validate that everything can actually run without a PS, which I feel is not the responsibility of this PR.
Previously PS was simply not an optional (at least on DX)

EmilioLaiso: Yeah, I'll revert to PS being mandatory for now.
I think making PS actually optional should be another PR. That would mean tackling other error checks as well, such as:

```
    if (!P.Bindings.RTargetBufferPtr)
          std::errc::invalid_argument,
          "No render target bound for graphics pipeline.");
```

EmilioLaiso: Removed the `Metal:` prefix :)

lib/API/DX/Device.cpp
EmilioLaiso: done ✅

test/Graphics/gs_point_to_quad.test
EmilioLaiso: Done! 🙌

lib/API/MTL/MTLDevice.cpp
EmilioLaiso: ✅

include/API/Device.h
```diff
@@ -81,6 +82,35 @@ struct ShaderContainer {
   llvm::SmallVector<SpecializationConstant> SpecializationConstants;
 };
 
+struct GraphicsPipelineCreateDesc {
```
manon-traverse:
```suggestion
struct TraditionalRasterPipelineCreateDesc {
```

MarijnS95 (approved):

PR #1205, opened 2026-05-19
[api-query] Report GPU driver version

alsepkow:
The "Dump GPU Info" step in CI runs `api-query` to log each adapter's API, description, and driver name. Logging the driver version would be useful too when triaging GPU/driver-specific failures, so we have a record of which driver build was in use when a test failed.

Changes:
- **`Device` base class:** Add `DriverVersion` member and `getDriverVersion()` getter.
- **D3D12:** Read `DXCoreAdapterProperty::DriverVersion` and decode the packed `LARGE_INTEGER` into `Major.Minor.Build.Revision`.
- **Vulkan:** Read `VkPhysicalDeviceDriverProperties::driverInfo` (vendor-specific human-readable build string, e.g. `"26.5.2 (AMD proprietary shader compiler)"`).
- **api-query:** Print a new `Driver Version:` line beneath `Driver:`.

Sample output:
```
- API: DirectX
  Description: AMD Radeon RX 6800
  Driver Version: 32.0.21043.10005
- API: Vulkan
  Description: AMD Radeon RX 6800
  Driver Version: 26.5.2 (AMD proprietary shader compiler)
```

> Note: the `Driver:` line is blank on D3D12 in the sample above — that's a pre-existing gap being addressed separately in #1206. This PR is independent.

---
Assisted by Copilot.

### DIFF
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -145,6 +145,7 @@ class Device {
 protected:
   std::string Description;
   std::string DriverName;
+  std::string DriverVersion;
 
 public:
   virtual const Capabilities &getCapabilities() = 0;
@@ -184,6 +185,7 @@ public:
 
   llvm::StringRef getDescription() const { return Description; }
   llvm::StringRef getDriverName() const { return DriverName; }
+  llvm::StringRef getDriverVersion() const { return DriverVersion; }
 };
 
 llvm::Error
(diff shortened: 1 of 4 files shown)


alsepkow:
I realized after that we are saving a dxdiag. That will have the driver version. Still seems convenient to have it logged here IMO though.

bogner (approved):
LGTM. See my comment in the other PR about how I feel like the current "DriverName" field is confusingly named.

farzonl (approved):

manon-traverse (approved):

PR #1206, opened 2026-05-19
[api-query] Populate `DriverName` on D3D12 devices

alsepkow:
The "Dump GPU Info" step in CI runs `api-query` to log each adapter's API, description, and driver name. On D3D12 devices the `Driver:` field was always empty: DXCore has no direct analog to Vulkan's `VkPhysicalDeviceDriverProperties::driverName`, so we never populated it.

Read `DXCoreAdapterProperty::HardwareID` and map `VendorID` to a vendor string (AMD / NVIDIA / Intel / Microsoft / Qualcomm) via a small lookup. Unknown vendors fall back to a hex ID, e.g. `Unknown vendor (0x1234)`.

Sample output after this change:
```
- API: DirectX
  Description: AMD Radeon RX 6800
  Driver: AMD
- API: DirectX
  Description: Microsoft Basic Render Driver
  Driver: Microsoft
```

---
Assisted by Copilot.


lib/API/DX/Device.cpp
```diff
@@ -1047,6 +1048,23 @@ class DXDevice : public offloadtest::Device {
...
+    case 0x10DE:
+      return "NVIDIA";
+    case 0x8086:
+      return "Intel";
+    case 0x1414:
+      return "Microsoft";
+    case 0x5143:
+      return "Qualcomm";
+    default:
+      return "";
```
V-FEXrt: nit: I'm not really a fan of empty string for "unknown or error" value. std::optional is a type that covers this case very well

V-FEXrt (approved):

lib/API/DX/Device.cpp
bogner: agreed, `std::optional` would be nice here.

bogner (commented):
This looks reasonable enough, but should we change the name of the field from "Driver" to "Vendor" if that's what we're getting here and what we get for Vulkan? With the field name being "Driver" I expect something with a version number, and would probably expect (for example) "WARP" rather than "Microsoft"

farzonl:
Why do we need this for DirectX? For spirv it makes sense because there are many different vulkan  implementations even for a vendors gpu like the proprietary and open source amd driver or the many mesa drivers.

alsepkow:
> Why do we need this for DirectX? For spirv it makes sense because there are many different vulkan implementations even for a vendors gpu like the proprietary and open source amd driver or the many mesa drivers.

We don't really **need** it. But I liked it better than having that field blank in the logs.

alsepkow:
> This looks reasonable enough, but should we change the name of the field from "Driver" to "Vendor" if that's what we're getting here and what we get for Vulkan? With the field name being "Driver" I expect something with a version number, and would probably expect (for example) "WARP" rather than "Microsoft"

I noticed this when I was adding logging to report driver version here. Decided to clean it up at the same time but in a separate PR. The problem on naming for DriveName is that in Vk we are literally populating from 'VkPhysicalDeviceDriverProperties::driverName'. D3D's closest equivalent is its VendorId. 

Maybe we'd prefer to just leave it blank on D3D? I still felt like it was better than that.

V-FEXrt (approved):

farzonl:
> This looks reasonable enough, but should we change the name of the field from "Driver" to "Vendor" if that's what we're getting here and what we get for Vulkan? With the field name being "Driver" I expect something with a version number, and would probably expect (for example) "WARP" rather than "Microsoft"

This is the problem there are no "Driver names"  for DirectX we only have them on Vulkan. Thats how we can distinguish `KosmicKrisp` from `MoltenVK`. I don't love this change because we are inventing a concept for DirectX just to fill in a  property that we already know from the Description.

farzonl:
> > Why do we need this for DirectX? For spirv it makes sense because there are many different vulkan implementations even for a vendors gpu like the proprietary and open source amd driver or the many mesa drivers.
> 
> We don't really **need** it. But I liked it better than having that field blank in the logs.

Why would you feel better? We already have the information you need like vendor name from the description.

alsepkow:
> > > Why do we need this for DirectX? For spirv it makes sense because there are many different vulkan implementations even for a vendors gpu like the proprietary and open source amd driver or the many mesa drivers.
> > 
> > 
> > We don't really **need** it. But I liked it better than having that field blank in the logs.
> 
> Why would you feel better? We already have the information you need like vendor name from the description.

An empty field just felt like a bug. It makes more sense now that it's just not applicable for DirectX. Maybe it would be better to fill it with 'N/A'?  Or maybe have the field name be something like "Vk Driver Name" ?

Do we have a precedent to have blank fields when they aren't applicable?

farzonl:
> > > > Why do we need this for DirectX? For spirv it makes sense because there are many different vulkan implementations even for a vendors gpu like the proprietary and open source amd driver or the many mesa drivers.
> > > 
> > > 
> > > We don't really **need** it. But I liked it better than having that field blank in the logs.
> > 
> > 
> > Why would you feel better? We already have the information you need like vendor name from the description.
> 
> An empty field just felt like a bug. It makes more sense now that it's just not applicable for DirectX. Maybe it would be better to fill it with 'N/A'? Or maybe have the field name be something like "Vk Driver Name" ?
> 
> Do we have a precedent to have blank fields when they aren't applicable?

If you really want a name just call it `DirectX`. There are multiple Vulkan implementations with different Driver names, but there is only one DirectX.

farzonl (approved):

PR #1061, opened 2026-04-06
[Metal] Explicit root signature layout for Metal backend

kcloudy0717:
Resolves https://github.com/llvm/offload-test-suite/issues/351, https://github.com/llvm/offload-test-suite/issues/744, https://github.com/llvm/offload-test-suite/issues/305.

This PR changes the automatic linear layout binding model to the more flexible explicit root signature binding model.

Notable changes:
- We no longer invokes `-metal` && `-Fre` to the compiler command line in LIT config. Reason for this is because explicit root signature model requires the RS to be bound to the IR compiler during IR conversion time.
- Introduced `MTLTopLevelArgumentBuffer` (TLAB) class that replaces the legacy `MTL::Buffer *ArgsBuffer` in `InvocationState`. This class encapsulates a buffer for the backing memory of the data of all IR resources for the given root signature. Provides D3D12-like functions for binding root constants, root descriptors, and descriptor tables.
- Introduced `MTLDescriptorHeap` class that replicates D3D12's descriptor heap concept. It serves as a storage for descriptors, binding a descriptor table to the TLAB must come from a descriptor heap just like D3D12. This bridge the gap between D3D12/Metal API differences allowing future tests such as dynamic resource indexing and unbound resource arrays easy to support.
- Resource/Descriptor creation is similar to the D3D12 backend thanks to `MTLTopLevelArgumentBuffer` and `MTLDescriptorHeap`, this allows resource arrays to be supported in the Metal backend now.

PR #1207, opened 2026-05-19
[Metal] Fix vertex buffer binding and tighten input layout checks

MarijnS95:
The Metal backend bound the vertex buffer at buffer index `0`, which collides with `kIRDescriptorHeapBindPoint` from Metal Shader Converter. When a descriptor heap is in use, `MTLDescriptorHeap::bind()` from #1061 first writes to slot 0 and the subsequent `setVertexBuffer(..., 0)` overwrites it, breaking dynamic resource indexing.

Apple's MSC sample binds the stage-in vertex buffer at `kIRVertexBufferBindPoint`; this PR moves the encoder binding and the vertex descriptor's attribute buffer index / layout slot to that constant.

While here, two related rough edges:

* `ShaderAttrIndices[AttrName]` used `StringMap::operator[]`, which silently inserts `0` on miss. A typo or shader/input-layout mismatch therefore mapped the offending element to MSL `[[attribute(0)]]`, clobbering whatever was placed there. Use `find()` and return a clear error.
* The "vertex shader attribute count vs. pipeline vertex input count" check compared `FnAttrs->count()` (which includes inactive attributes) to `InputLayout.size()`, while the rest of the code only tracks active attributes. Build the active-attribute map first, then compare its size.

Tested locally with `check-hlsl-mtl-graphics` (6/6 pass) and the broader `check-hlsl-mtl` suite (320 pass, 0 regressions).

### DIFF
--- a/lib/API/MTL/MTLDevice.cpp
+++ b/lib/API/MTL/MTLDevice.cpp
@@ -658,6 +658,7 @@ public:
     assert(Slot <
                sizeof(IRRuntimeVertexBuffers) / sizeof(IRRuntimeVertexBuffer) &&
            "Vertex buffer slot exceeds Metal Shader Converter limit");
+    assert(Slot == 0 && "Pipeline vertex descriptor only describes slot 0");
     if (VB) {
       auto &MTLVB = llvm::cast<MTLBuffer>(*VB);
       RenderEnc->setVertexBuffer(MTLVB.Buf, Offset, BufIdx);
--- a/lib/API/MTL/MTLDevice.cpp
+++ b/lib/API/MTL/MTLDevice.cpp
@@ -1628,14 +1628,10 @@ public:
             std::errc::invalid_argument,
             "Vertex shader has no vertex attributes.");
 
-      if (FnAttrs->count() != InputLayout.size())
-        return llvm::createStringError(
-            std::errc::invalid_argument,
-            "Mismatch between vertex shader attribute count and pipeline "
-            "vertex input count.");
-
       // Collect the attribute indices the shader expects so that we can map
-      // the specified attributes onto the correct indices.
+      // the specified attributes onto the correct indices. Only active
+      // attributes are exposed via stage_in; inactive entries are dropped by
+      // the optimizer and have no corresponding [[attribute(N)]] slot.
       llvm::StringMap<uint32_t> ShaderAttrIndices;
       for (uint32_t I = 0; I < FnAttrs->count(); ++I) {
         auto *A = static_cast<MTL::VertexAttribute *>(FnAttrs->object(I));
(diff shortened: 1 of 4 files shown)


manon-traverse (approved):
The plan behind this PR is great, it just needs a rebase I think

lib/API/MTL/MTLDevice.cpp
```diff
@@ -1382,7 +1381,9 @@ class MTLDevice : public offloadtest::Device {
     Encoder.setScissor(Scissor);
 
     if (IS.VB)
-      Encoder.setVertexBuffer(0, IS.VB.get(), 0, P.Bindings.getVertexStride());
+      // Match the buffer index used in the vertex descriptor.
+      Encoder.setVertexBuffer(kIRVertexBufferBindPoint, IS.VB.get(), 0,
+                              P.Bindings.getVertexStride());
```
manon-traverse: This causes a divergence on the code that is supposed to be API-agnostic. Can we hide these details in the `setVertexBuffer` function of the `RenderCommandEncoder`?

manon-traverse (changes requested):
Good change, but I don't think the `kIRVertexBufferBindPoint` is applied in the wrong layer. What do you think?

Also thanks for applying the small clean-ups.

lib/API/MTL/MTLDevice.cpp
MarijnS95: Yes correct, `setVertexBuffer()` in a public/global API shouldn't expose a slot index / bind point (maybe an attribute index).

MarijnS95: @manon-traverse it turns out `kIRVertexBufferBindPoint` is the _starting point_, so we'll simply add that on top of `uint32 Slot` inside the implementation.

However, what slots to use is tightly related to PSO creation (vertex attribute layout), where Metal is currently hardcoded to a single slot. I haven't checked the other backends, how is this dealt with there? Does it make sense to already expose a `Slot` in the API, or should we assert that it's `0` until a PSO constructor becomes available that supports multiple slots (if ever needed at all)?

MarijnS95: Summarized:

  - DX12 lib/API/DX/Device.cpp:1105 — ElementDesc.InputSlot = 0; for every element.
  - Vulkan lib/API/VK/Device.cpp:1610, 1618 — Attr.binding = 0; and single BindingDesc.binding = 0;.
  - MTL lib/API/MTL/MTLDevice.cpp:1667, 1680 — single layout at kIRVertexBufferBindPoint (= slot 0).

So I'll add an `assert(Slot==0` to every backend with a minimal explanation for this.

manon-traverse: That sounds like a good solution for now.
We can support more slots later.

manon-traverse (approved):

MarijnS95:
[The only remaining failure](https://github.com/llvm/offload-test-suite/actions/runs/26251175918/job/77345396011?pr=1207) is:

```
Unexpectedly Passed Tests (1):
  OffloadTest-clang-vk :: Feature/HLSLLib/reversebits.64.test
```

On `check-hlsl-clang-vk` which is [also present on `main`](https://github.com/llvm/offload-test-suite/actions/runs/26274077837/job/77334059647), merging. (EDIT: Wish we vould have rebase-merge for this)

PR #1208, opened 2026-05-19
Introduce `createBufferWithData` function for easy buffer creation.

manon-traverse:
As a step towards making resource and descriptor creation the same across graphics APIs we need a function to handle creating buffers and uploading data. This includes optionally creating an upload buffer.

Right now, it's only used to create the vertex buffer and avoids the use of an upload buffer by placing the vertices in host visible memory.

In a future PR, all generic buffers should be created via this path, and the same for textures with similar functions. However, this is quite a bit of refactoring. Ideally, we do this in smaller steps so PRs are easier to review. On top of that, this work is fairly low priority.

Ideally at some point we do introduce lifetime tracking so that `createBufferWithData` becomes even easier to use since the caller won't have to care about keeping the upload buffer alive.

### DIFF
--- a/include/API/Buffer.h
+++ b/include/API/Buffer.h
@@ -30,8 +30,7 @@ struct BufferCreateDesc {
   BufferUsage Usage;
 
   static BufferCreateDesc uploadBuffer() {
-    return BufferCreateDesc{.Location = MemoryLocation::CpuToGpu,
-                            .Usage = BufferUsage::Storage};
+    return BufferCreateDesc{MemoryLocation::CpuToGpu, BufferUsage::Storage};
   }
 };
 
--- a/include/API/Buffer.h
+++ b/include/API/Buffer.h
@@ -28,6 +28,13 @@ enum class BufferUsage {
 struct BufferCreateDesc {
   MemoryLocation Location;
   BufferUsage Usage;
+
+  static BufferCreateDesc uploadBuffer() {
+    BufferCreateDesc Result = {};
+    Result.Location = MemoryLocation::CpuToGpu;
+    Result.Usage = BufferUsage::Storage;
+    return Result;
+  }
 };
 
 class Buffer {
--- a/include/API/Buffer.h
+++ b/include/API/Buffer.h
@@ -30,10 +30,8 @@ struct BufferCreateDesc {
   BufferUsage Usage;
 
   static BufferCreateDesc uploadBuffer() {
-    BufferCreateDesc Result = {};
-    Result.Location = MemoryLocation::CpuToGpu;
-    Result.Usage = BufferUsage::Storage;
-    return Result;
+    return BufferCreateDesc{.Location = MemoryLocation::CpuToGpu,
+                            .Usage = BufferUsage::Storage};
   }
 };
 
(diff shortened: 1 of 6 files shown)


MarijnS95 (approved):
Yes that is a nice approach.

(Albeit a bit hacky, this flow makes it easy to experiment with upload-less UMA by simply overriding all buffers to be `CpuToGpu` in `createBufferWithData()` 🙂)

manon-traverse:
> Yes that is a nice approach.
> 
> (Albeit a bit hacky, this flow makes it easy to experiment with upload-less UMA by simply overriding all buffers to be `CpuToGpu` in `createBufferWithData()` 🙂)

At some point we could also check if the device is UMA and just map the GpuOnly buffer directly.

include/API/Buffer.h
```diff
@@ -28,6 +28,13 @@ enum class BufferUsage {
 struct BufferCreateDesc {
   MemoryLocation Location;
   BufferUsage Usage;
+
+  static BufferCreateDesc uploadBuffer() {
+    BufferCreateDesc Result = {};
+    Result.Location = MemoryLocation::CpuToGpu;
+    Result.Usage = BufferUsage::Storage;
+    return Result;
```
bogner: I guess we're doing this all over the place already, but I don't really understand why. Initializing members individually is more error prone than using an initializer list or calling a constructor. Why not:

```suggestion
    return BufferCreateDesc{MemoryLocation::CpuToGpu, BufferUsage::Storage};
```

bogner (approved):
LGTM, but my comment here kind of applies generally across a lot of preexisting code.

include/API/Buffer.h
manon-traverse: My main gripe with this pattern is that I need to remember the layout of the struct to see what I am initializing.
I really wish C++ would have gone the C route with something like this:
```
BufferCreateDesc Result = {
  .Location = MemoryLocation::CpuToGpu,
  .Usage = BufferUsage::Storage,
};
```
Rust also has the names there.

I'm okay with with the initializer list, but I wonder what happens when we add a field. Do all usages of the initializer list fail to compiler because a member is missing or will it default initialize those?

manon-traverse: From a little test I can see I get a warning, but no error. I suppose that's better than nothing.
I'll switch them to use a designated initializer and see if that compiles on all the platforms we test on. It's part of C++20 so it should be fine.

bogner: Unfortunately LLVM is still on C++17, so we can't use designated initializers yet (Last discussion about upgrading min compilers was [here](https://discourse.llvm.org/t/rfc-raise-the-minimum-compiler-requirements-to-move-toward-c-20/88894/27), I believe).

> I'm okay with with the initializer list, but I wonder what happens when we add a field. Do all usages of the initializer list fail to compiler because a member is missing or will it default initialize those?

Initializer lists will fail to compile if we add a field, they need to initialize the whole struct. If we want to default things we would need to use a constructor instead.

manon-traverse: Oh from my test with an intializer list it compiled even though I didn't specify all the fields 🤔  It just gave me warning

I am not really a fan of writing constructors since it's a lot of extra boiler plate code to set some fields. Perhaps there is no way around this and just an inherent limitation of C++

bogner: > Oh from my test with an intializer list it compiled even though I didn't specify all the fields 🤔 It just gave me warning

Ah, you're right, it's just a warning under `-Wmissing-field-initializers`. That is set in the LLVM build though, so `-DLLVM_ENABLE_WERROR=On` configs do error at least.

> I am not really a fan of writing constructors since it's a lot of extra boiler plate code to set some fields. Perhaps there is no way around this and just an inherent limitation of C++

Fair enough

Issue #1158, opened 2026-05-07
Build Acceleration Structure infrastructure for Ray Tracing tests


MarijnS95:
Smaller tracking issue to document the needed work for Acceleration Structures in the framework.

- [x] Ability to specify acceleration structures (triangles in BLAS, instance(s) of these BLAS(es) in a TLAS) via YAML
- [x] Creating Acceleration Structure objects
- [x] Binding backing buffers to Acceleration Structure objects
  - [x] Includes related buffer usage flags
- [ ] Building geometry/instance descriptions from GPU buffers (based on `DeviceAddress`)
  - [ ] Lots of pointer-soup and "keepalive"
  - [x] GPU buffers must allow requesting addresses
- [ ] Building geometry/instances into Acceleration Structure objects
  - [x] Requires the new `CommandEncoder` abstraction to perform this generically across backends
  - [ ] Requires proper barrier placement between sequential dependent builds (TLAS must wait for all its BLASes to be built)

Binding these to a ray tracing shader (starting with inline) will be tracked in a separate issue.

PR #1214, opened 2026-05-21
Add YAML-driven acceleration structure definitions for .test files

MarijnS95:

## Summary
- Add `AccelerationStructure` to `ResourceKind` and introduce pipeline structs (`TriangleGeometry`, `AABBGeometry`, `BLASDesc`, `InstanceDesc`, `TLASDesc`) in `include/Support/Pipeline.h` with YAML parsing and name-based resolution for buffer/BLAS/TLAS references in `lib/Support/Pipeline.cpp`.
- Add a DX `RaytracingTier` capability enum (`NotSupported`/`1.0`/`1.1`) wired through `DXFeatures.def`/`.h`/`.cpp` so the DX device can report a raytracing tier.
- Surface an `acceleration-structure` lit feature in `test/lit.cfg.py`, derived per-backend from DX `RaytracingTier != NotSupported`, Metal `supportsRaytracing`, and Vulkan `VK_KHR_acceleration_structure`.
- Add `AccelerationStructure` cases to the resource-kind switch statements in the DX, Metal, and Vulkan device backends (no execution yet — just enum coverage).
- Add three placeholder InlineRT test files (`triangle-setup`, `indexed-triangle-setup`, `multi-instance`) exercising the new YAML format, gated behind `REQUIRES: acceleration-structure` and marked `XFAIL: *` because backend execution of acceleration structures is not implemented yet.

This PR only introduces the YAML surface area, parsing, and capability detection — the runtime side that actually builds/binds acceleration structures is intentionally out of scope and will follow.

## Test plan
- [x] `check-hlsl-offload` passes on DX, Metal, and Vulkan with the three new tests reported as `XFAIL` (or skipped where `acceleration-structure` isn't available).
- [x] Existing tests are unaffected on all three backends.

### DIFF
--- a/include/Support/Pipeline.h
+++ b/include/Support/Pipeline.h
@@ -14,6 +14,7 @@
 #define OFFLOADTEST_SUPPORT_PIPELINE_H
 
 #include "API/Enums.h"
+#include "API/Resources.h"
 #include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/Error.h"
@@ -80,6 +81,7 @@ static inline DescriptorKind getDescriptorKind(ResourceKind RK) {
   case ResourceKind::StructuredBuffer:
   case ResourceKind::ByteAddressBuffer:
   case ResourceKind::Texture2D:
+  case ResourceKind::AccelerationStructure:
     return DescriptorKind::SRV;
 
   case ResourceKind::RWStructuredBuffer:
@@ -221,6 +223,8 @@ struct Result {
   double Epsilon;
 };
 
+struct TLASDesc;
+
 struct Resource {
   ResourceKind Kind;
   std::string Name;
@@ -231,6 +235,11 @@ struct Resource {
   bool HasCounter;
   std::optional<uint32_t> TilesMapped;
   bool IsReserved = false;
+  TLASDesc *TLASPtr = nullptr;
+
+  bool isAccelerationStructure() const {
+    return Kind == ResourceKind::AccelerationStructure;
+  }
 
   bool isRaw() const {
     switch (Kind) {
@@ -240,6 +249,7 @@ struct Resource {
     case ResourceKind::RWTexture2D:
     case ResourceKind::Sampler:
     case ResourceKind::SampledTexture2D:
+    case ResourceKind::AccelerationStructure:
       return false;
     case ResourceKind::StructuredBuffer:
     case ResourceKind::RWStructuredBuffer:
@@ -265,6 +275,7 @@ struct Resource {
     case ResourceKind::Texture2D:
     case ResourceKind::RWTexture2D:
     case ResourceKind::SampledTexture2D:
+    case ResourceKind::AccelerationStructure:
       return false;
     }
     llvm_unreachable("All cases handled");
@@ -280,6 +291,7 @@ struct Resource {
     case ResourceKind::RWByteAddressBuffer:
     case ResourceKind::ConstantBuffer:
     case ResourceKind::Sampler:
+    case ResourceKind::AccelerationStructure:
       return false;
     case ResourceKind::Texture2D:
     case ResourceKind::RWTexture2D:
@@ -319,18 +331,22 @@ struct Resource {
   }
 
   uint32_t getElementSize() const {
-    assert(!isSampler() && "Samplers do not have element size");
+    assert(!isSampler() && !isAccelerationStructure() &&
+           "Samplers and AS do not have element size");
     // ByteAddressBuffers are treated as 4-byte elements to match their memory
     // format.
     return isByteAddressBuffer() ? 4 : BufferPtr->getElementSize();
   }
 
   uint32_t getArraySize() const {
-    return isSampler() ? 1 : BufferPtr->ArraySize;
+    if (isSampler() || isAccelerationStructure())
+      return 1;
+    return BufferPtr->ArraySize;
   }
 
   uint32_t size() const {
-    assert(!isSampler() && "Samplers do not have size");
+    assert(!isSampler() && !isAccelerationStructure() &&
+           "Samplers and AS do not have size");
     return BufferPtr->size();
   }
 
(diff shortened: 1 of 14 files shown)


manon-traverse:
It looks like you forgot to add `include/API/AccelerationStructure.h` to this PR.

include/Support/Pipeline.h
```diff
@@ -534,6 +614,11 @@ LLVM_YAML_IS_SEQUENCE_VECTOR(offloadtest::SpecializationConstant)
 LLVM_YAML_IS_SEQUENCE_VECTOR(offloadtest::PushConstantBlock)
 LLVM_YAML_IS_SEQUENCE_VECTOR(offloadtest::PushConstantValue)
 LLVM_YAML_IS_SEQUENCE_VECTOR(offloadtest::DispatchParametersSet)
+LLVM_YAML_IS_SEQUENCE_VECTOR(offloadtest::CPUTriangleGeometry)
+LLVM_YAML_IS_SEQUENCE_VECTOR(offloadtest::CPUAABBGeometry)
+LLVM_YAML_IS_SEQUENCE_VECTOR(offloadtest::CPUBLASDesc)
+LLVM_YAML_IS_SEQUENCE_VECTOR(offloadtest::CPUInstanceDesc)
+LLVM_YAML_IS_SEQUENCE_VECTOR(offloadtest::CPUTLASDesc)
```
manon-traverse: Not sure about these names.
I feel like the `CPU` prefix is not really working out (even though I introduced it to differentiate between two types called `Buffer`).

I think in this case the names are fine if we just drop the prefix. What do you think?

include/Support/Pipeline.h
```diff
@@ -227,6 +232,11 @@ struct Resource {
   bool HasCounter;
   std::optional<uint32_t> TilesMapped;
   bool IsReserved = false;
+  CPUTLASDesc *TLASPtr = nullptr;
+
+  bool isAccelerationStructure() const {
+    return Kind == ResourceKind::AccelerationStructure;
+  }
```
manon-traverse: This design is okay for now, but I am not sure if we should keep treating buffers, textures and acceleration structures all as resources in the yaml format.
It might make more sense to split these types up at some point. What do you think would make sense in the future?

test/Feature/InlineRT/indexed-triangle-setup.test
```diff
@@ -0,0 +1,80 @@
...
+  - Stage: Compute
+    Entry: main
+Buffers:
+  - Name: Vertices
+    Format: Float32
+    Stride: 12
+    Data: [ 0.0, 1.0, 0.0, -1.0, -1.0, 0.0, 1.0, -1.0, 0.0, 0.0, 0.0, 0.5 ]
+  - Name: Indices
+    Format: UInt32
+    Data: [ 0, 1, 2 ]
```
manon-traverse: If the backend were to ignore the indices this test would still pass. Can use an index buffer that would reorder the vertices in such a way that it will only pass if you actually use the index buffer?

include/Support/Pipeline.h
MarijnS95: Claude would like to follow up on this in a separate PR: agreed the right long-term shape is something like `struct DescriptorBinding { ... std::variant<BufferBinding, TextureBinding, SamplerBinding, AccelerationStructureBinding> Payload; }` so each binding kind owns only its relevant pointer, instead of `Resource` having three "maybe-set" pointers gated by `Kind`. Keeping it as-is in this PR to minimize blast radius into the backends.

manon-traverse (approved):

EmilioLaiso (approved):

test/Feature/InlineRT/indexed-triangle-setup.test
```diff
@@ -0,0 +1,83 @@
+# REQUIRES: acceleration-structure
+# UNSUPPORTED: Clang
+# XFAIL: *
```
bogner: We usually put these in the last part of the file, just before the RUN and CHECK lines. Also, please put a comment before the XFAIL explaining that the backend parts aren't implemented yet (if there's an issue to link to, it would be enough to just link to it)

test/Feature/InlineRT/indexed-triangle-setup.test
```diff
@@ -0,0 +1,83 @@
...
+    Expected: Expected
+...
+#--- end
+
+# RUN: split-file %s %t
+# RUN: %dxc_target -T cs_6_5 -Fo %t.o %t/source.hlsl
+# RUN: %offloader %t/pipeline.yaml %t.o | FileCheck %s
+
+# CHECK: Name: Output
+# CHECK: Data: [ 1 ]
```
bogner: Why are we doing both a `Results` block and `FileCheck`? The FileCheck part is redundant.

bogner (approved):
My comments on the one test apply to all three, but otherwise this LGTM.

test/Feature/InlineRT/indexed-triangle-setup.test
MarijnS95: Interesting. Many other tests still seem to use a stdout-readback + FileCheck/CHECK, and even though FileCheck is more featureful (regex, CHECK-NEXT/NOT, ordering assertions on log output) most of those are also "just compare a buffer" and could equally well use a direct buffer comparison through `Results:`. Any clue why they don't — historical inertia from before `Results:` existed, or is there a deliberate reason I'm missing? Happy to do a consistency pass in a separate PR if you'd like them migrated.

bogner: It's just historical inertia from before `Results` existed. If you feel like doing a consistency pass I wouldn't say no to it :)

test/Feature/InlineRT/indexed-triangle-setup.test
```diff
@@ -0,0 +1,85 @@
+# REQUIRES: acceleration-structure
+# UNSUPPORTED: Clang
```
bogner: I think I wasn't clear in my previous comment. All of the REQUIRES, UNSUPPORTED, and XFAIL should be in the `#--- end` section of the test file, just before the `RUN` lines, so

```
#--- source.hlsl
...
//--- pipeline.yaml
...
#--- end

# REQUIRES: acceleration-structure
# UNSUPPORTED: Clang

# Unimplemented https://github.com/llvm/offload-test-suite/issues/1158.
# XFAIL: *

# RUN: split-file %s %t

bogner: Also, shouldn't this be `XFAIL: Clang`? We're planning to implement this, but haven't yet

MarijnS95:
Test failures are unrelated to this PR, merging.

PR #1217, opened 2026-05-21
Bind compute PSO inside ComputeEncoder::dispatch()

MarijnS95:
`ComputeEncoder::dispatch()` now takes a `PipelineState &` as its first argument and binds the PSO before issuing the dispatch, the same shape as `RenderEncoder::drawInstanced()`.

Making the PSO a required parameter enforces at the type level that callers cannot forget to bind a pipeline (an invalid dispatch on every backend), and keeps the bind adjacent to the command it applies to instead of relying on separate "last bound" state on the command list. It also hides per-API requirements behind the encoder: on Metal the threadgroup-size from the HLSL `numthreads()` annotation is read from shader reflection at pipeline creation and cached on `MTLPipelineState`, instead of leaking out as a `setThreadGroupSize()` call every caller had to remember. `dispatch()` just consumes the cached size, and we drop the encoder-state helpers `ThreadsPerGroup` / `setThreadGroupSize()` entirely. With the bind moved, the per-backend compute PSO bind in `createComputeCommands()` (DX, VK, MTL) is removed.

### DIFF
--- a/lib/API/MTL/MTLDevice.cpp
+++ b/lib/API/MTL/MTLDevice.cpp
@@ -233,9 +233,13 @@ public:
   std::string Name;
   IRRootSignaturePtr RootSig;
   std::unique_ptr<MTLTopLevelArgumentBuffer> ArgBuffer;
-  IRShaderReflectionPtr Reflection;
   MTL::ComputePipelineState *ComputePipeline = nullptr;
   MTL::RenderPipelineState *RenderPipeline = nullptr;
+
+  // Compute pipeline only state. Threadgroup size comes from numthreads() in
+  // the HLSL source and is captured from shader reflection at pipeline
+  // creation, so dispatch() doesn't need to re-query reflection each time.
+  MTL::Size ThreadsPerGroup = MTL::Size(1, 1, 1);
 
   // Rasterization pipeline only state.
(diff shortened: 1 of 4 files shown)


lib/API/MTL/MTLDevice.cpp
```diff
@@ -507,14 +495,37 @@ class MTLComputeEncoder : public offloadtest::ComputeEncoder {
...
     flushBarrier();
     insertDebugSignpost(llvm::formatv("Dispatch [{0},{1},{2}]", GroupCountX,
                                       GroupCountY, GroupCountZ)
                             .str());
+    ComputeEnc->setComputePipelineState(MTLPSO.ComputePipeline);
+
+    // Threadgroup size comes from numthreads() in the HLSL source, persisted
+    // in the transpiled Metallib via shader reflection.
+    MTL::Size ThreadsPerGroup(
+        MTLPSO.ComputePipeline->maxTotalThreadsPerThreadgroup(), 1, 1);
```
manon-traverse: In which case do we expect to have no reflection data to get the number of threads from? If there is no case it would make more sense to error rather than use some kind of default value which may be incorrect.

Also should we grab this data during pipeline creation instead over every time we dispatch?

manon-traverse (approved):

EmilioLaiso (approved):

PR #1221, opened 2026-05-26
Pass-through geometry shader test

EmilioLaiso:
Adds a passthrough geometry shader test that exercises the GS pipeline path without any vertex expansion or topology change. 
The GS takes a triangle in and emits the same triangle, vertices untouched. A single oversized triangle covers the 2×2 render target, and the pixel shader writes opaque red.

### DIFF
--- a/test/Graphics/gs_passthrough.test
+++ b/test/Graphics/gs_passthrough.test
@@ -1,0 +1,102 @@
+#--- vertex.hlsl
+struct VSOutput {
+  float4 position : SV_POSITION;
+};
+
+// Pass-through vertex shader. Real coverage decisions are made by the input
+// vertex data, not the shader.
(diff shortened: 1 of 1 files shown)


test/Graphics/gs_passthrough.test
```diff
@@ -0,0 +1,96 @@
...
+  float4 position : SV_POSITION;
+};
+
+struct GSOutput {
+  float4 position : SV_POSITION;
+};
+
+// Identity geometry shader: one triangle in, one triangle out, vertices
+// untouched. Exercises the GS pipeline path on the default TriangleList
+// topology without any expansion or topology change.
```
manon-traverse: This comment doesn't really add anything.

manon-traverse (approved):

MarijnS95 (approved):

PR #1224, opened 2026-05-26
Add tessellation support to DX/VK backends with accompanying `SimpleTriangleTess.test`

EmilioLaiso:
Adds `Hull` and `Domain` shader support, plus a `PatchList` primitive topology with per-call control-point count. 
Wires the new stages on the DX12 and Vulkan backends. 
Metal explicitly rejects them. Metal's tessellation model (HS as a compute kernel, DS as a `[[patch(...)]]`-tagged vertex function) doesn't neatly fit the HS/DS shape, so the path is intentionally left unimplemented. 
Adds one end-to-end test, `SimpleTriangleTess.test`, exercising a VS → HS → DS → PS pipeline that renders the `SimpleTriangle` image test via tessellation.

### DIFF
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -87,8 +87,13 @@ struct TraditionalRasterPipelineCreateDesc {
   llvm::SmallVector<Format> RTFormats;
   std::optional<Format> DSFormat;
   PrimitiveTopology Topology;
+  // Only meaningful when Topology == PatchList. Validated in Pipeline.cpp.
+  uint32_t PatchControlPoints = 0;
   ShaderContainer VS;
-  // TODO: Optional Hull & Domain Shaders
+  // Hull and Domain are independent optionals here; Pipeline.cpp enforces that
+  // they must be set as a pair (and only with PatchList topology).
+  std::optional<ShaderContainer> HS;
+  std::optional<ShaderContainer> DS;
   std::optional<ShaderContainer> GS;
   ShaderContainer PS;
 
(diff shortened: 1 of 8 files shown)


include/API/Device.h
```diff
@@ -87,8 +87,15 @@ struct TraditionalRasterPipelineCreateDesc {
...
   std::optional<Format> DSFormat;
   PrimitiveTopology Topology;
+  // Set if Topology == PatchList. Validated in
+  // Pipeline.cpp::validatePipelineKind.
+  std::optional<uint32_t> PatchControlPoints;
+
   ShaderContainer VS;
-  // TODO: Optional Hull & Domain Shaders
+  // Hull and Domain are independent optionals here; Pipeline.cpp enforces that
+  // they must be set as a pair (and only with PatchList topology).
```
MarijnS95, line 96: If both must be set, there're not independent, but dependent or **inter**dependent?

include/Support/Pipeline.h
```diff
@@ -402,6 +405,12 @@ struct IOBindings {
   CPUBuffer *RTargetBufferPtr = nullptr;
   PrimitiveTopology Topology = PrimitiveTopology::TriangleList;
 
+  // Set if Topology == PatchList. Validated in
+  // Pipeline.cpp::validatePipelineKind. Valid range is 1..32 (matches both
```
MarijnS95, line 409: Only if you're making changes (resolving comments via Claude...) anyway:

```suggestion
  // Pipeline.cpp::validatePipelineKind(). Valid range is 1..32 (matches both
```

MarijnS95 (approved):

include/API/Device.h
MarijnS95: In-person: the **here** was referring to the two `std::optional<>`s below _allowing them to be independent_ while they're _not supposed_ to be independent. I would reword that to be more like the latter sentence to make the intent clear :)

manon-traverse: I think the comment is fine.

manon-traverse (approved):

PR #1225, opened 2026-05-26
Enable split build feature on all workflows

bob80905:
The infrastructure seems to be in place for github to split jobs into build and test jobs and distribute build jobs across compatible runners. This PR does a bit of quality improvement, but also actually enables the split build feature on all workflows.

Separately, this PR adds an sccache step so that we can track the hit rate, and ensures that clang is built in the build LLVM stage and not during the install step.

Assisted by: Github Copilot.
Manifests the implementation of: https://github.com/llvm/offload-test-suite/issues/1081


.github/workflows/build-and-test-callable.yaml
```diff
@@ -362,11 +376,26 @@ jobs:
           arch: arm64
       - name: Setup Python
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
-        with:
-          # The test runner needs only lit + pyyaml; no source checkout is
-          # available so we can't pip install -r requirements.txt. Keep
-          # this list in sync with OffloadTest/test/requirements.txt.
-          pip-install: lit pyyaml
+      - name: Install lit and pyyaml
```
inbelic, line 379: From [here](https://github.com/llvm/offload-test-suite/actions/runs/26776253649/job/78928518255?pr=1225#step:9:1) I can verify that the sccache step is running as expected.

But it seems like this step isn't and it still uses the user path?

.github/workflows/build-and-test-callable.yaml
```diff
@@ -201,6 +207,14 @@ jobs:
             cd build
             cmake -G Ninja ${{ inputs.LLVM-ExtraCMakeArgs }} -DCMAKE_BUILD_TYPE=${{ inputs.BuildType }} -DLLVM_ENABLE_ASSERTIONS=On -DLLVM_EXTERNAL_PROJECTS="OffloadTest" -DHLSL_ENABLE_OFFLOAD_DISTRIBUTION=${{ inputs.SplitBuild == true && 'On' || 'Off' }} -C ${{ github.workspace }}/llvm-project/clang/cmake/caches/HLSL.cmake -C ${{ github.workspace }}/OffloadTest/cmake/caches/sccache.cmake -DDXC_DIR=${{ github.workspace }}/DXC/build/bin -DLLVM_EXTERNAL_OFFLOADTEST_SOURCE_DIR=${{ github.workspace }}/OffloadTest -DLLVM_LIT_ARGS="--xunit-xml-output=testresults.xunit.xml -v" -DOFFLOADTEST_TEST_CLANG=${{steps.Test-Clang.outputs.TEST_CLANG || 'Off' }} -DGOLDENIMAGE_DIR=${{ github.workspace }}/golden-images -DCMAKE_INSTALL_PREFIX=${{ github.workspace }}/install ${{ github.workspace }}/llvm-project/llvm/
             ninja hlsl-test-depends
+      # Surface sccache cache state after the heavy compile work so we can
+      # track hit rate / cache size across runs and diagnose cold-cache
+      # blowups on new builders. Always-runs so we still get stats on
+      # build failure.
+      - name: sccache stats
```
inbelic: nit:

```suggestion
      - name: Dump sccache Stats
```
The names seem to be verbs and capitalized

.github/workflows/build-and-test-callable.yaml
bob80905: Here is an example where this step runs:
https://github.com/llvm/offload-test-suite/actions/runs/26776253602/job/78937729286?pr=1225#step:8:138
I believe the linked pipeline you're looking at doesn't have the split build-feature enabled. 
From what I can gather, the pipelines without my changes, triggered by PR matrix and configured as exec tests are run. And, the workflows I changed are individually run with the changes. 
`Execution Testing` is the prefix for the unchanged runs, it looks like, and `Windows` is the prefix for the runs with my changes.

inbelic: I see thanks. Okay that makes more sense.

inbelic (approved):

.github/workflows/build-and-test-callable.yaml
```diff
@@ -362,11 +376,26 @@ jobs:
...
+        # The runner service account (NetworkService) can't write to
+        # the system site-packages under C:\Program Files\Python313, so
+        # pip falls back to a --user install. That puts lit.exe in
+        # %APPDATA%\Python\Python313\Scripts\ which is NOT on PATH.
+        # We compute that dir from sysconfig and prepend it to PATH for
+        # all subsequent steps so `lit` resolves correctly.
+        shell: bash
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install lit pyyaml
```
tex3d: Shouldn't we pin exact versions/hashes rather than just getting latest versions?

On that note, I noticed that test/requirements.txt only uses `pyyaml>=6.0.0`, which will drift as well.

.github/workflows/build-and-test-callable.yaml
```diff
@@ -362,11 +376,26 @@ jobs:
...
+          user_scripts=$(python -c "import sysconfig; print(sysconfig.get_path('scripts', f'{sysconfig.get_default_scheme()}_user'))")
+          echo "User scripts dir: $user_scripts"
+          echo "$user_scripts" >> "$GITHUB_PATH"
```
tex3d, line 410: I think you could avoid this path hack by running lit like this instead:
```
python -m lit -v "run/test/$suite" \
  --xunit-xml-output="run/test/$suite/testresults.xunit.xml"
```

.github/workflows/build-and-test-callable.yaml
```diff
@@ -362,11 +376,26 @@ jobs:
...
+        # The test runner needs only lit + pyyaml; no source checkout is
+        # available so we can't pip install -r requirements.txt. Keep
+        # this list in sync with OffloadTest/test/requirements.txt.
+        # We invoke pip explicitly here instead of relying on
+        # setup-python's `pip-install:` input, which was observed to
+        # silently no-op on our self-hosted Windows runners.
+        # The runner service account (NetworkService) can't write to
+        # the system site-packages under C:\Program Files\Python313, so
+        # pip falls back to a --user install. That puts lit.exe in
+        # %APPDATA%\Python\Python313\Scripts\ which is NOT on PATH.
```
tex3d: Note that this sounds Windows-specific, so it would be good to make sure what you have works on all target OS platforms.

.github/workflows/build-and-test-callable.yaml
bob80905: I tried python -m lit initially and it fails with:
 ```
  C:\Program Files\Python313\python.exe: No module named lit.__main__; 'lit' is a package and cannot be directly executed
 ```
 The lit package on PyPI doesn't ship a __main__.py. The lit console script that pip installs is the only supported entry point. That's why we have to make sure the scripts dir is on PATH. I've kept the comment block in the step explaining the rationale.

tex3d (approved):

PR #1229, opened 2026-05-27
[DX][VK] Implement SV_Depth pixel-shader output and Depth32 SRV mapping (#1046)

### LINKED ISSUE #1046: [DirectX] Add depth/stencil buffer support
This is required for testing of the `SV_Depth*` and `SV_StencilRef` semantics.

Current status: Depth/stencil buffer backend is already supported: https://github.com/llvm/offload-test-suite/pull/756 and https://github.com/llvm/offload-test-suite/pull/1020

This allows testing of the `SV_Depth*` semantics, however, the `SV_StencilRef` is currently never enabled through the schema.

AC:
- [ ] Add schema to allow for enabling the stencil ref state
- [ ] Add test demonstrating that we can observe the stencil ref works
- [ ] Add test demonstrating we can observe the various depth types

alsepkow:
Add SV_Depth pixel-shader output and Depth32-as-SRV support to the D3D12 and Vulkan backends.

Introduces a DepthBuffer YAML binding with an explicit Format field so tests can declare a CPU-readable depth attachment. The D3D12 backend creates a DSV, transitions to COPY_SOURCE for readback, and maps Depth32 to R32_FLOAT for SRV reads. Vulkan has a symmetric implementation; Metal returns an unsupported error for now.

Tests: SVDepth.test (pixel-shader depth output) and Texture2D.Load.Depth32.test.yaml (compute SRV read).

Resolves #1046

Assisted by Claude Opus 4.7


lib/API/Device.cpp
```diff
@@ -107,6 +107,31 @@ offloadtest::createDefaultDepthStencilTarget(Device &Dev, uint32_t Width,
...
 
+llvm::Expected<std::unique_ptr<Texture>>
+offloadtest::createDepthBufferFromCPUBuffer(Device &Dev, const CPUBuffer &Buf) {
+  auto TexFmtOrErr = toFormat(Buf.Format, Buf.Channels);
+  if (!TexFmtOrErr)
+    return TexFmtOrErr.takeError();
+  if (*TexFmtOrErr != Format::D32Float)
+    return llvm::createStringError(
+        std::errc::invalid_argument,
+        "Depth buffer binding requires DataFormat::Depth32 with 1 channel.");
```
manon-traverse: Why do we parse a format from the .yaml file if we only accept one format?
Also, we should consider using the `Format` enum in the .yaml so we don't have to do this conversion in the first place. We can then just check `isDepth` instead, which will also allow us to support other depth formats like those that include a stencil buffer.

lib/API/DX/Device.cpp
```diff
@@ -963,6 +969,10 @@ class DXDevice : public offloadtest::Device {
     std::unique_ptr<offloadtest::Texture> RenderTarget;
     std::unique_ptr<offloadtest::Buffer> RTReadback;
     std::unique_ptr<offloadtest::Texture> DepthStencil;
+    // Set only when the pipeline has a Bindings.DepthBuffer; the depth target
+    // contents are copied here after the draw so the test can verify
+    // SV_Depth* writes.
```
manon-traverse: What is this comment for?

lib/API/DX/Device.cpp
```diff
@@ -981,6 +982,8 @@ class DXDevice : public offloadtest::Device {
     std::unique_ptr<offloadtest::Texture> RenderTarget;
     std::unique_ptr<offloadtest::Buffer> RTReadback;
     std::unique_ptr<offloadtest::Texture> DepthStencil;
+    // Populated when Bindings.DepthBuffer is set, for SV_Depth verification.
```
alsepkow:
```suggestion
```

lib/API/DX/Device.cpp
alsepkow: AI generated, removed.

lib/API/Device.cpp
alsepkow: Followed your suggestion.

test/Feature/Textures/Texture2D.Load.Depth32.test.yaml
```diff
@@ -0,0 +1,76 @@
...
+    Actual: Out
+    Expected: Expected
+...
+#--- end
+
+# RUN: split-file %s %t
+# RUN: %dxc_target -T cs_6_0 -Fo %t.o %t/source.hlsl
+# RUN: %offloader %t/pipeline.yaml %t.o
+
+# Metal's offload backend cannot bind a depth-format texture as an SRV
```
alsepkow, line 73: nit: comment not needed

alsepkow:
Moving this back to draft. Some NFC changes were factored out into a separate PR. #1327 
When thats merged I'll merge it into this branch and remove the PR from draft again.

lib/API/DX/Device.cpp
```diff
@@ -2788,6 +2786,28 @@ class DXDevice : public offloadtest::Device {
...
+      uint32_t DSNumRows = 0;
+      uint64_t DSRowSizeInBytes = 0;
+      uint64_t DSTotalBytes = 0;
+      Device->GetCopyableFootprints(&DSDesc, 0u, 1u, 0u, &DSPlaced, &DSNumRows,
+                                    &DSRowSizeInBytes, &DSTotalBytes);
+
+      P.Bindings.DepthBuffer.Ptr->copyFromTexture(DSMapped,
+                                                  DSPlaced.Footprint.RowPitch);
+      DSReadback.Buffer->Unmap(0, nullptr);
+    }
```
manon-traverse, line 2809: Let's use the `copyTextureToBuffer` function on the `ComputeEncoder` instead ^.^
This got recently added to main for the DX12 backend and the VK and MTL backends will use that path soon too.

jzakharovnv:
Hi @alsepkow. Any reason that this was closed? I'll be taking a stab at implementing this myself in the following days and am curious. Also, if you have any wisdom/pitfalls you'd be willing to share, I'd really appreciate it!

alsepkow:
> Hi @alsepkow. Any reason that this was closed? I'll be taking a stab at implementing this myself in the following days and am curious. Also, if you have any wisdom/pitfalls you'd be willing to share, I'd really appreciate it!

@jzakharovnv Some of the required changes were implemented in [#1403](https://github.com/llvm/offload-test-suite/pull/1403) And the remaining work was just abandoned here as this work was moved back to @inbelic

PR #1232, opened 2026-05-27
Add RT acceleration structure abstraction with size queries and resource allocation

MarijnS95:

Introduce the foundational types for ray tracing acceleration structures:
- Abstract `AccelerationStructure` base class
- Geometry/instance descriptors and BLAS/TLAS build-request structs with size queries
- AS resource allocation across DX12, Vulkan, and Metal backends

Recording the actual build commands lands in a follow-up commit on top of the `ComputeEncoder` abstraction.

### DX12 device interface bump

`ID3D12DeviceX` typedef goes from `ID3D12Device2` to `ID3D12Device5`, so the existing `Device` member directly exposes `GetRaytracingAccelerationStructurePrebuildInfo` (and the eventual `CreateStateObject` / `SetPipelineState1` for the PSO RT epic) — no separate `Device5` member or post-create `QueryInterface` dance. `D3D12CreateDevice` is already invoked with `IID_PPV_ARGS(&Device)`, so the bump naturally requires the adapter to support the Device5 interface (Win10 1809+); RT-capable hardware is selected by the `acceleration-structure` lit feature regardless.

### Vulkan device-creation refactor

Single `vkGetPhysicalDeviceFeatures2` call: every extension feature struct we care about (`atomic-int64`, `mesh-shader`, `acceleration-structure`, BDA on 1.1) is chained into `pNext` before the query. Post-query we verify each extension's gating feature bool and clear the sub-features we don't need (capture-replay, indirect-build, multiview, etc.).

Drive-by: rather than letting `vkCreateDevice` reject the device with a generic `VK_ERROR_FEATURE_NOT_PRESENT`, the code returns a descriptive `llvm::Error` naming the extension and the bool that came back zero — pinpointing the case where a driver advertises an extension but reports its base feature as `VK_FALSE`. The duplicate `queryDeviceExtensions` call is gone, and `EnabledDeviceExtensions` is now one list (the previous `EnabledExtensions` separate vector silently dropped mesh-shader / atomic-int64 entries when RT was enabled).

### DIFF
--- a/include/API/AccelerationStructure.h
+++ b/include/API/AccelerationStructure.h
@@ -1,0 +1,171 @@
+//===- AccelerationStructure.h - RT Acceleration Structure Types ----------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef OFFLOADTEST_API_ACCELERATIONSTRUCTURE_H
+#define OFFLOADTEST_API_ACCELERATIONSTRUCTURE_H
+
+#include "API/API.h"
+#include "API/Buffer.h"
+#include "API/Resources.h"
+
+#include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/Error.h"
+
+#include <cstdint>
+#include <variant>
+
+namespace offloadtest {
+
+struct AccelerationStructureSizes {
+  uint64_t ResultDataMaxSizeInBytes = 0;
+  uint64_t ScratchDataSizeInBytes = 0;
+  uint64_t UpdateScratchDataSizeInBytes = 0;
+};
+
+struct TriangleGeometryDesc {
+  Buffer *VertexBuffer = nullptr;
+  uint64_t VertexBufferOffset = 0;
+  uint32_t VertexCount = 0;
+  uint32_t VertexStride = 0;
+  Format VertexFormat = Format::RGB32Float;
+  Buffer *IndexBuffer = nullptr;
+  uint64_t IndexBufferOffset = 0;
+  uint32_t IndexCount = 0;
+  IndexFormat IdxFormat = IndexFormat::Uint32;
+  bool Opaque = true;
+};
+
+struct AABBGeometryDesc {
+  Buffer *AABBBuffer = nullptr;
+  uint64_t AABBBufferOffset = 0;
+  uint32_t AABBCount = 0;
+  uint32_t AABBStride = 24;
+  bool Opaque = true;
+};
+
+class AccelerationStructure;
+
+struct AccelerationStructureInstance {
+  float Transform[3][4] = {{1, 0, 0, 0}, {0, 1, 0, 0}, {0, 0, 1, 0}};
+  uint32_t InstanceID = 0;
+  uint8_t InstanceMask = 0xFF;
+  AccelerationStructure *BLAS = nullptr;
+};
+
+struct BLASBuildRequest {
+  // DXR / Vulkan / Metal all forbid mixing triangle and AABB geometry in a
+  // single BLAS, so the geometry list is held as a variant — the invalid
+  // mixed-geometry state is unrepresentable.
+  std::variant<llvm::SmallVector<TriangleGeometryDesc>,
+               llvm::SmallVector<AABBGeometryDesc>>
+      Geometry;
+  AccelerationStructureSizes Sizes;
+};
+
+struct TLASBuildRequest {
+  llvm::SmallVector<AccelerationStructureInstance> Instances;
+  AccelerationStructureSizes Sizes;
+};
+
+inline llvm::Error validateGeometryDesc(const TriangleGeometryDesc &D) {
+  if (!D.VertexBuffer)
+    return llvm::createStringError(
+        std::errc::invalid_argument,
+        "TriangleGeometryDesc: VertexBuffer is null.");
+  if (!isPositionCompatible(D.VertexFormat))
+    return llvm::createStringError(
+        std::errc::invalid_argument,
+        "TriangleGeometryDesc: VertexFormat '%s' is not position-compatible.",
+        getFormatName(D.VertexFormat).data());
+  if (D.VertexStride < getFormatSizeInBytes(D.VertexFormat))
+    return llvm::createStringError(
+        std::errc::invalid_argument,
+        "TriangleGeometryDesc: VertexStride (%u) must be >= format size (%u).",
+        D.VertexStride, getFormatSizeInBytes(D.VertexFormat));
+  if (D.VertexCount == 0)
+    return llvm::createStringError(std::errc::invalid_argument,
+                                   "TriangleGeometryDesc: VertexCount is 0.");
+  if (D.IndexBuffer && D.IndexCount == 0)
+    return llvm::createStringError(
+        std::errc::invalid_argument,
+        "TriangleGeometryDesc: IndexBuffer is set but IndexCount is 0.");
+  if (!D.IndexBuffer && D.IndexCount != 0)
+    return llvm::createStringError(
+        std::errc::invalid_argument,
+        "TriangleGeometryDesc: IndexCount is set but IndexBuffer is null.");
+  if (D.IndexBuffer && D.IndexCount % 3 != 0)
+    return llvm::createStringError(
+        std::errc::invalid_argument,
+        "TriangleGeometryDesc: IndexCount (%u) must be a multiple of 3.",
+        D.IndexCount);
+  if (!D.IndexBuffer && D.VertexCount % 3 != 0)
+    return llvm::createStringError(
+        std::errc::invalid_argument,
+        "TriangleGeometryDesc: VertexCount (%u) must be a multiple of 3 when "
+        "no index buffer is provided.",
+        D.VertexCount);
+  return llvm::Error::success();
+}
+
+inline llvm::Error validateGeometryDesc(const AABBGeometryDesc &D) {
+  if (!D.AABBBuffer)
+    return llvm::createStringError(std::errc::invalid_argument,
+                                   "AABBGeometryDesc: AABBBuffer is null.");
+  if (D.AABBCount == 0)
+    return llvm::createStringError(std::errc::invalid_argument,
+                                   "AABBGeometryDesc: AABBCount is 0.");
+  if (D.AABBStride < 24)
+    return llvm::createStringError(
+        std::errc::invalid_argument,
+        "AABBGeometryDesc: AABBStride (%u) must be >= 24.", D.AABBStride);
+  return llvm::Error::success();
+}
+
+template <typename T>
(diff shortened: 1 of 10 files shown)


include/API/AccelerationStructure.h
```diff
@@ -0,0 +1,198 @@
...
+  return static_cast<AccelerationStructureBuildFlags>(static_cast<uint32_t>(A) |
+                                                      static_cast<uint32_t>(B));
+}
+
+inline AccelerationStructureBuildFlags
+operator&(AccelerationStructureBuildFlags A,
+          AccelerationStructureBuildFlags B) {
+  return static_cast<AccelerationStructureBuildFlags>(static_cast<uint32_t>(A) &
+                                                      static_cast<uint32_t>(B));
+}
```
EmilioLaiso: This could use `LLVM_MARK_AS_BITMASK_ENUM` I think. Saves you from implementing the operators.

It would require:

- `#include "llvm/ADT/BitmaskEnum.h"`
- `LLVM_MARK_AS_BITMASK_ENUM(/* LargestValue = */ PreferFastBuild)`
- `namespace offloadtest {
LLVM_ENABLE_BITMASK_ENUMS_IN_NAMESPACE(); ...`

In writing this reply I noticed `LLVM_ENABLE_BITMASK_ENUMS_IN_NAMESPACE();` is missing from `enum TextureUsage`, so maybe that could also be a drive-by fix in this PR 😅

MarijnS95: How'd you find this? Your `TextureUsage` is also the first enum in the entire workspace that's using this. That's nice, just curious why it wasn't utilized before.

MarijnS95: Adopted (and the `TextureUsage` drive-by is in) — with one twist. Using `LLVM_MARK_AS_BITMASK_ENUM` on a second enum in the same namespace as `TextureUsage` triggers a `redefinition of enumerator 'LLVM_BITMASK_LARGEST_ENUMERATOR'` error: the macro literally inserts that enumerator inside the enum, and an unscoped enum's enumerators are visible at the enclosing namespace scope, so two such enums in one namespace collide.

The LLVM header documents `LLVM_DECLARE_ENUM_AS_BITMASK` as the non-intrusive alternative for exactly this case ("allows declaring more than one non-scoped enumerations as bitmask types in the same scope"), so I used that for `AccelerationStructureBuildFlags` (placed in `namespace llvm` at the end of the header).

Let me know if you'd prefer to also convert `TextureUsage` over to the non-intrusive form so both look the same — happy to do that as another drive-by here.

EmilioLaiso (approved):
lgtm

include/API/AccelerationStructure.h
```diff
@@ -0,0 +1,193 @@
...
+  return llvm::Error::success();
+}
+
+inline llvm::Error validateBLASBuildRequest(const BLASBuildRequest &Req) {
+  if (Req.Triangles.empty() && Req.AABBs.empty())
+    return llvm::createStringError(
+        std::errc::invalid_argument,
+        "BLASBuildRequest: Must have at least one geometry descriptor.");
+  // Vulkan and Metal forbid mixing triangle and AABB geometry in a single BLAS;
+  // D3D12 technically permits it but is uncommon. Reject for portability.
```
manon-traverse: Incorrect, D3D12 also doesn't permit it according to the spec.

https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#geometry-and-acceleration-structures

> Within a given bottom-level acceleration structure there can be any number: (1) triangle meshes, or (2) procedural primitives initially described only by an axis aligned bounding box (AABB). **A bottom-level acceleration structure can only contain a single geometry type.**

lib/API/DX/Device.cpp
```diff
@@ -933,6 +950,7 @@ class DXDevice : public offloadtest::Device {
 private:
   ComPtr<IDXCoreAdapter> Adapter;
   ComPtr<ID3D12DeviceX> Device;
+  ComPtr<ID3D12Device5> Device5;
```
manon-traverse: This looks like it's going to be a mess. Perhaps we rename `ID3D12DeviceX` to `ID3D12DeviceBaseLevel` or something like that?
I don't wanna have 8 different device pointers in here at some point.

lib/API/DX/Device.cpp
```diff
@@ -1559,6 +1579,179 @@ class DXDevice : public offloadtest::Device {
     return std::make_unique<DXRenderPass>(Desc);
   }
 
+  llvm::Expected<BLASBuildRequest>
+  createBLASBuildRequest(llvm::ArrayRef<TriangleGeometryDesc> Triangles,
+                         llvm::ArrayRef<AABBGeometryDesc> AABBs,
+                         AccelerationStructureBuildFlags Flags) override {
```
manon-traverse: We should use some kind of `std::variant` here or have two function calls, one for triangle geo and one for aabb geo. You are not allowed to mix these two so our API shouldn't allow it.

lib/API/DX/Device.cpp
```diff
@@ -1559,6 +1579,179 @@ class DXDevice : public offloadtest::Device {
...
+
+    if (Flags & AllowUpdate)
+      Inputs.Flags |=
+          D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_UPDATE;
+    if (Flags & PreferFastTrace)
+      Inputs.Flags |=
+          D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_TRACE;
+    if (Flags & PreferFastBuild)
+      Inputs.Flags |=
+          D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_BUILD;
```
manon-traverse: Specifying both of these flags are not allowed. Can we split them in a separate `BuildMode` enum?

`Balanced`, `FastTrace`, `FastBuild`.

lib/API/VK/Device.cpp
```diff
@@ -505,6 +505,40 @@ class VulkanTexture : public offloadtest::Texture {
...
 };
 
+class VulkanAccelerationStructure : public offloadtest::AccelerationStructure {
+public:
+  VkDevice Dev;
+  VkAccelerationStructureKHR AccelStruct;
+  VkBuffer Buffer;
+  VkDeviceMemory Memory;
+  VkDeviceAddress DeviceAddress;
+  PFN_vkDestroyAccelerationStructureKHR FnDestroyAS;
```
manon-traverse, line 515: Should we group all the acceleration structure extensions functions together in a struct instead?

include/API/AccelerationStructure.h
MarijnS95: Reworded to make the runtime-vs-spec distinction explicit: DXR forbids the mix; D3D12 won't catch the violation at build time; only Vulkan and Metal do. We reject up-front so the spec-invalid combo behaves the same everywhere.

lib/API/DX/Device.cpp
MarijnS95: [Claude] Misread — the type added in this PR is `ID3D12Device5` (a versioned interface), not `ID3D12DeviceX`. The `ID3D12DeviceX` alias predates this PR and isn't being introduced here. No rename involved.

[Me] I never intended for a `DeviceX` type to exist, in a previous review I just wanted to resolve whatever X was and name that throughout the API unless we really expect that to change often.
The reason a new device is added here to ensure RT is properly "optional" and not pulling the whole framework down when some setup doesn't host `Device5` yet. We could go either way but this was a conscious choice.

lib/API/DX/Device.cpp
MarijnS95: ~The mixed-geometry case is rejected at runtime in `validateBLASBuildRequest`. Not unwinding the `BLASBuildRequest` layout in this PR.~ Addressing this turns out to be a massive diff across backends; trying to boil it down to keep changes small.

lib/API/DX/Device.cpp
MarijnS95: The bitflag faithfully mirrors the DXR `D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAGS` enum; the mutually-exclusive pair is validated at runtime. Not refactoring to a separate `BuildMode` enum in this PR.

lib/API/VK/Device.cpp
MarijnS95: The AS extension function pointers are already grouped:

https://github.com/llvm/offload-test-suite/blob/19ff2e5b04f40da3e76510355c0a3c9ce902c852/lib/API/VK/Device.cpp#L1154-L1159

 The lone `PFN_vkDestroyAccelerationStructureKHR` held on `VulkanAccelerationStructure` here is the only one needed for cleanup of an AS handle; not worth dragging the whole struct in for a single function.

lib/API/DX/Device.cpp
MarijnS95: Also this pattern was clearly copied from our own codebase. Worth rewriting in both cases.

lib/API/DX/Device.cpp
MarijnS95: Reversing course on this — we did do the variant split after all. `BLASBuildRequest::Geometry` is now `std::variant<SmallVector<TriangleGeometryDesc>, SmallVector<AABBGeometryDesc>>` and the device API splits into `createTriangleBLASBuildRequest` / `createAABBBLASBuildRequest`. The spec-invalid mixed-geometry state is unrepresentable at the type level. (Force-pushed.)

lib/API/DX/Device.cpp
manon-traverse: I know it mirrors DXR, but that's the problem. DXR allows for invalid combinations. Not a big deal but would be nice if we can disallow these combinations on our generic interface.

MarijnS95: Actually deleting it; I remember deleting this before as it's completely irrelevant for what triangles the shaders are going to hit.

If anything we could hardcode FAST_BUILD in each backend to speed up CPU-side AS building and test execution if that ever becomes a problem.

lib/API/VK/Device.cpp
```diff
@@ -1257,76 +1305,143 @@ class VulkanDevice : public offloadtest::Device {
...
     Features.pNext = &Features11;
-    if (Props.apiVersion >= VK_MAKE_API_VERSION(0, 1, 2, 0))
+    if (HasVulkan12)
       Features11.pNext = &Features12;
     if (Props.apiVersion >= VK_MAKE_API_VERSION(0, 1, 3, 0))
       Features12.pNext = &Features13;
 #ifdef VK_VERSION_1_4
     if (Props.apiVersion >= VK_MAKE_API_VERSION(0, 1, 4, 0))
       Features13.pNext = &Features14;
 #endif
```
bogner: It feels a little funny to have `HasVulkan12` without also doing the same for 13/14. Should we add a couple of bools for consistency?

bogner (approved):

lib/API/VK/Device.cpp
MarijnS95: Good catch — added `HasVulkan13`/`HasVulkan14` (the latter under the `VK_VERSION_1_4` guard, matching `HasVulkan12`'s shape). Also threaded them through `queryCapabilities`, which had the same six inline `VK_MAKE_API_VERSION` checks.

PR #1233, opened 2026-05-27
Fix warnings that cause main to fail CI.

manon-traverse:
Some warnings managed to creep in main.
One of the warnings was an actual bug where a field (`DriverVersion`) was not initialized.

### DIFF
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -978,6 +978,7 @@ public:
         DSVAllocator(std::move(DSVAllocator)) {
     Description = std::move(Desc);
     DriverName = "DirectX";
+    DriverVersion = DriverVer;
   }
   DXDevice(const DXDevice &) = delete;
   DXDevice &operator=(const DXDevice &) = delete;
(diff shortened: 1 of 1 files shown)


bogner (approved):
The DriverVersion initialization was fixed in https://github.com/llvm/offload-test-suite/pull/1228 and can be removed. `const` changes LGTM.

bob80905 (approved):

lib/API/DX/Device.cpp
```diff
@@ -978,6 +978,7 @@ class DXDevice : public offloadtest::Device {
         DSVAllocator(std::move(DSVAllocator)) {
     Description = std::move(Desc);
     DriverName = "DirectX";
+    DriverVersion = DriverVer;
```
bogner, line 981:
```suggestion
```

bogner: Hm, I thought I'd be able to apply this suggestion and then merge this to get the CI going, but apparently I don't have that power. Putting up my own PR https://github.com/llvm/offload-test-suite/pull/1242 so I can merge this now.

bogner:
Merged #1242 instead.

PR #1240, opened 2026-05-27
Remove XFAIL for reversebit.64 when targeting vulkan

joaosaffran:
The original issue was fixed, so now we can remove the xfail

### DIFF
--- a/test/Feature/HLSLLib/reversebits.64.test
+++ b/test/Feature/HLSLLib/reversebits.64.test
@@ -79,9 +79,6 @@ DescriptorSets:
 ...
 #--- end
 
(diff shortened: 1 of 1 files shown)


bogner (approved):

bob80905 (approved):

test/Feature/HLSLLib/reversebits.64.test
```diff
@@ -80,7 +80,7 @@ DescriptorSets:
 #--- end
 
 # Bug https://github.com/llvm/llvm-project/issues/197810
```
bob80905: You may want to either re-open the issue, or make a new issue specifying the problem with QC.

bob80905 (approved):

PR #1242, opened 2026-05-28
Fix warnings that cause main to fail CI

bogner:
Some warnings managed to creep into main.


bogner:
Note: this is #1233 with the now duplicate fix from #1228 removed. Creating this version so that I can get it merged and get CI fixed.

PR #1243, opened 2026-05-28
Add a simple passthrough amplification shader test.

manon-traverse:
Add a test that does a simple passthrough ampflication shader that only passes a color to modfiy the green component of the triangle vertices. The test is there just to test that the amplification test runs and that the payload is passed correctly to the mesh shader.

### DIFF
--- a/test/Graphics/MeshShaders/SimpleAmplification.test
+++ b/test/Graphics/MeshShaders/SimpleAmplification.test
@@ -6,12 +6,13 @@ struct Payload {
     float3 color;
 };
 
+groupshared Payload gs_payload;
+
 [numthreads(1, 1, 1)]
 void main()
 {
-    Payload payload = (Payload)0;
-    payload.color = float3(1.0, 0.5, 1.0);
-    DispatchMesh(1, 1, 1, payload);
+    gs_payload.color = float3(1.0, 0.5, 1.0);
+    DispatchMesh(1, 1, 1, gs_payload);
 }
 
 #--- mesh.hlsl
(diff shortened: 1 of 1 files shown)


EmilioLaiso (approved):
Approving, but as discussed offline this doesn't really require a new golden image as it can reuse the simple triangle one.

test/Graphics/MeshShaders/SimpleAmplification.test
```diff
@@ -0,0 +1,110 @@
+// Test that has a really simple passthrough amplification shader that halfs the
```
MarijnS95:
```suggestion
// Test that has a really simple passthrough amplification shader that halves the
```

test/Graphics/MeshShaders/SimpleAmplification.test
```diff
@@ -0,0 +1,110 @@
...
+    SetMeshOutputCounts(3, 1);
+
+    float4 position;
+    float4 color = float4(payload.color, 1.0);
+    if (groupThreadID == 0) {
+        position = float4(0.0, 0.25, 0.0, 1.0);
+        color *= float4(1.0, 0.0, 0.0, 1.0);
+    } else if (groupThreadID == 1) {
+        position = float4(0.25, -0.25, 0.0, 1.0);
+        color *= float4(0.0, 2.0, 0.0, 1.0);
```
MarijnS95: Curious why this one is green `2.0`, to compensate?

manon-traverse: Yes, it's so we get back to the colors of the SimpleTriangle golden image when the amplification shader payload is applied.

MarijnS95: Might deserve a tiny comment to let readers notice this in-code, otherwise good to go.

MarijnS95 (approved):

PR #1244, opened 2026-05-28
Add ComputeEncoder::batchBuildAS() and shared AS-build orchestration helper

MarijnS95:

This PR contains two stacked commits:

### `Refactor AS sizing/alloc into a Device::getXBuildSizes() + createBLAS()/createTLAS() API`

A build request now strictly describes build *inputs* — its `AccelerationStructureSizes` field moves out and is queried via a new sizing-only entry point on `Device`. Allocation no longer accepts a build request:

- `Expected<AccelerationStructureSizes> getBLASBuildSizes(triangles)`
- `Expected<AccelerationStructureSizes> getBLASBuildSizes(aabbs)`
- `Expected<AccelerationStructureSizes> getTLASBuildSizes(instanceCount)`
- `Expected<unique_ptr<AS>>             createBLAS(sizes)`
- `Expected<unique_ptr<AS>>             createTLAS(sizes)`

The old `createTriangleBLASBuildRequest()` / `createAABBBLASBuildRequest()` / `createTLASBuildRequest()` / `createAccelerationStructure(Request)` entry points are removed. The `AccelerationStructure` base class additionally stores the sizes it was allocated for and exposes `getSizes()` so the GPU build path can size its scratch buffer without re-querying. The per-backend `createBLAS()` / `createTLAS()` implementations fold the existing "create request + allocate" pair into one allocation step parameterised solely on `AccelerationStructureSizes`.

### `Add ComputeEncoder::batchBuildAS() and shared AS-build orchestration helper`

Move acceleration-structure build commands behind the abstract `ComputeEncoder` interface so the orchestration (data upload, AS allocation, build recording) can live in one place rather than splitting across three backends.

`ComputeEncoder` gains a single `batchBuildAS(ArrayRef<ASBuildItem>)` method. `BLASBuildRequest` / `TLASBuildRequest` each carry their own target `AS *` field (filled in by the caller after `Dev.createBLAS()` / `createTLAS()`); `ASBuildItem` collapses to a `PointerUnion<const BLASBuildRequest *, const TLASBuildRequest *>` alias, and backends size their scratch via `Req->AS->getSizes()` rather than reaching into the request for sizes or into a separate item struct for the target. The caller guarantees no inter-item memory dependencies inside a batch — backends record the whole batch with one barrier slot, no per-element barriers.

- **Vulkan**: single `vkCmdBuildAccelerationStructuresKHR()` call covering the whole batch. TLAS items serialize `VkAccelerationStructureInstanceKHR` into a device-address upload buffer; BLAS items pull addresses from each `VulkanBuffer` (new `getDeviceAddress()` accessor). Storage buffers transparently gain `SHADER_DEVICE_ADDRESS` + `ACCEL_BUILD_INPUT_READ_ONLY` flags when ray tracing is supported, with the matching `VkMemoryAllocateFlagsInfo` chained on every allocation.
- **DX12**: loop calling `BuildRaytracingAccelerationStructure()` per item with no intermediate barriers; `D3D12_RAYTRACING_INSTANCE_DESC` is bit-identical to the Vulkan instance struct.
- **Metal**: lazy transition to `MTL::AccelerationStructureCommandEncoder`, deduplicates BLAS handles into the `MTL::InstanceAccelerationStructureDescriptor`'s `instancedAccelerationStructures` array (Metal references BLASes by index, not GPU address).

A shared helper `buildPipelineAccelerationStructures()` in `lib/API/Device.cpp` walks `Pipeline::AccelStructs`, uploads vertex/index data via the new `createBufferWithData()`, queries sizes via `Dev.getBLASBuildSizes()` / `Dev.getTLASBuildSizes()`, allocates handles via `Dev.createBLAS()` / `Dev.createTLAS()`, and issues two `batchBuildAS()` calls (BLAS batch then TLAS batch — VUID-03403 forbids referencing a sibling `dstAccelerationStructure` in one command). Each backend's `executeProgram()` calls this helper to build the pipeline's AS objects.

Descriptor binding for AS resources is intentionally not in this PR — the existing InlineRT tests still fail at the descriptor-creation step (an unbound AS resource trips `createSRV()` / equivalents in each backend) and stay caught by their `XFAIL: *` annotation. #1245 wires the AS descriptor write and flips them green.

### DIFF
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -264,20 +264,27 @@ public:
   virtual llvm::Expected<std::unique_ptr<CommandBuffer>>
   createCommandBuffer() = 0;
 
-  virtual llvm::Expected<BLASBuildRequest> createTriangleBLASBuildRequest(
-      llvm::ArrayRef<TriangleGeometryDesc> Triangles) = 0;
+  // Sizing queries: return the result/scratch sizes the backend needs to
+  // allocate an AS that can hold the given build inputs. The build-input
+  // pointers (geometry buffers, BLAS handles) are never consulted — only
+  // counts/strides — so these can be called before BLAS handles exist.
+  virtual llvm::Expected<AccelerationStructureSizes>
+  getBLASBuildSizes(llvm::ArrayRef<TriangleGeometryDesc> Triangles) = 0;
 
-  virtual llvm::Expected<BLASBuildRequest>
-  createAABBBLASBuildRequest(llvm::ArrayRef<AABBGeometryDesc> AABBs) = 0;
+  virtual llvm::Expected<AccelerationStructureSizes>
+  getBLASBuildSizes(llvm::ArrayRef<AABBGeometryDesc> AABBs) = 0;
 
-  virtual llvm::Expected<TLASBuildRequest> createTLASBuildRequest(
-      llvm::ArrayRef<AccelerationStructureInstance> Instances) = 0;
+  virtual llvm::Expected<AccelerationStructureSizes>
+  getTLASBuildSizes(uint32_t InstanceCount) = 0;
+
+  // Allocate the AS storage (no GPU build). The build is recorded later via
+  // ComputeEncoder::batchBuildAS(), which is the only place that consumes the
+  // associated *BuildRequest.
+  virtual llvm::Expected<std::unique_ptr<AccelerationStructure>>
+  createBLAS(const AccelerationStructureSizes &Sizes) = 0;
 
   virtual llvm::Expected<std::unique_ptr<AccelerationStructure>>
-  createAccelerationStructure(const BLASBuildRequest &Request) = 0;
-
-  virtual llvm::Expected<std::unique_ptr<AccelerationStructure>>
-  createAccelerationStructure(const TLASBuildRequest &Request) = 0;
+  createTLAS(const AccelerationStructureSizes &Sizes) = 0;
 
   virtual ~Device() = 0;
 
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -323,6 +323,25 @@ createBufferWithData(Device &Dev, std::string Name,
                      size_t SizeInBytes, ComputeEncoder *Encoder,
                      std::unique_ptr<offloadtest::Buffer> *OutUploadBuffer);
 
+// Builds all BLAS / TLAS objects defined in `P.AccelStructs` using the
+// supplied compute encoder. Uploads each BLAS's vertex/index data, queries
+// sizes via `Dev.getBLASBuildSizes` / `Dev.getTLASBuildSizes`, allocates
+// the handles via `Dev.createBLAS` / `Dev.createTLAS`, and records the GPU
+// builds via two `Enc.batchBuildAS` calls (BLAS batch then TLAS batch — so
+// the AS-build-write barrier between BLAS and TLAS is automatic).
+//
+// Built AS objects are pushed to `OutAS` (in declaration order: BLASes first,
+// then TLASes). Vertex/index buffers used as build inputs are pushed to
+// `OutInputBuffers`; both must outlive command-buffer submission.
+//
+// TODO: `Pipeline` belongs to the test framework, not the rendering backend
+// API. This helper lives here only because `executeProgram` is still on
+// `Device` — once that moves out, this helper should follow.
+llvm::Error buildPipelineAccelerationStructures(
+    Device &Dev, ComputeEncoder &Enc, Pipeline &P,
+    llvm::SmallVectorImpl<std::unique_ptr<AccelerationStructure>> &OutAS,
+    llvm::SmallVectorImpl<std::unique_ptr<Buffer>> &OutInputBuffers);
+
 } // namespace offloadtest
 
 #endif // OFFLOADTEST_API_DEVICE_H
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -587,13 +587,21 @@ public:
       override;
 };
 
+class DXDevice; // forward decl — defined below in this same anon ns
+
 class DXCommandBuffer : public offloadtest::CommandBuffer {
 public:
   ComPtr<ID3D12CommandAllocator> Allocator;
   ComPtr<ID3D12GraphicsCommandListX> CmdList;
+  /// Back-pointer to the owning device. Used by encoders that need access to
+  /// device-level resources (e.g. allocating AS scratch buffers).
+  DXDevice *Dev = nullptr;
   /// Whether a UAV barrier is pending from a prior compute command.
   bool PendingUAVBarrier = false;
   llvm::SmallVector<D3D12_RESOURCE_BARRIER> PendingTransitions;
+  /// Buffers that must outlive command-buffer submission (e.g. AS scratch
+  /// and TLAS instance buffers used during builds).
+  llvm::SmallVector<std::unique_ptr<offloadtest::Buffer>> KeepAliveOwned;
 
   static llvm::Expected<std::unique_ptr<DXCommandBuffer>>
   create(ComPtr<ID3D12DeviceX> Device) {
@@ -782,6 +790,10 @@ public:
 
     return llvm::Error::success();
   }
+
+  // Defined out-of-line below — needs DXDevice's full type for access to the
+  // ID3D12Device5 entry point and helper allocators.
+  llvm::Error batchBuildAS(llvm::ArrayRef<ASBuildItem> Items) override;
 
   void endEncodingImpl() override { popDebugGroup(); }
 };
@@ -1049,6 +1061,10 @@ DXCommandBuffer::createRenderEncoder(
 }
 
 class DXDevice : public offloadtest::Device {
+  // DXComputeEncoder needs access to Device5 for AS build commands and to the
+  // raw ID3D12Device for scratch buffer allocation.
+  friend class DXComputeEncoder;
+
 private:
   ComPtr<IDXCoreAdapter> Adapter;
   ComPtr<ID3D12DeviceX> Device;
@@ -1104,6 +1120,12 @@ private:
 
     llvm::SmallVector<DescriptorTable> DescTables;
     llvm::SmallVector<ResourcePair> RootResources;
+
+    // Built acceleration structures, kept alive for the pipeline lifetime.
+    llvm::SmallVector<std::unique_ptr<offloadtest::AccelerationStructure>>
+        AccelStructs;
+    // Vertex/index buffers consumed during AS builds; must outlive submission.
+    llvm::SmallVector<std::unique_ptr<offloadtest::Buffer>> ASInputBuffers;
   };
 
 public:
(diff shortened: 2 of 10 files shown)


lib/API/DX/Device.cpp
```diff
@@ -571,12 +571,21 @@ class DXQueue : public offloadtest::Queue {
...
   ComPtr<ID3D12GraphicsCommandListX> CmdList;
+  /// Back-pointer to the owning device. Used by encoders that need access to
+  /// device-level resources (e.g. allocating AS scratch buffers).
+  DXDevice *Dev = nullptr;
   /// Whether a UAV barrier is pending from a prior compute command.
   bool PendingUAVBarrier = false;
+  /// Resources that must outlive command-buffer submission (e.g. AS scratch
+  /// and TLAS instance buffers used during builds).
+  std::vector<ComPtr<ID3D12Resource>> KeepAliveResources;
+  llvm::SmallVector<std::unique_ptr<offloadtest::Buffer>> KeepAliveOwned;
```
EmilioLaiso: Could scratch be `offloadtest::Buffer` as well so that we can have a single keep alive struct?

manon-traverse: I agree with making those `offloadtest::Buffer`

EmilioLaiso (approved):
LGTM

EmilioLaiso (approved):

include/API/Device.h
```diff
@@ -316,6 +289,21 @@ createBufferWithData(Device &Dev, std::string Name,
...
+// builds via two `Enc.batchBuildAS` calls (BLAS batch then TLAS batch — so
+// the AS-build-write barrier between BLAS and TLAS is automatic).
+//
+// Built AS objects are pushed to `OutAS` (in declaration order: BLASes first,
+// then TLASes). Vertex/index buffers used as build inputs are pushed to
+// `OutInputBuffers`; both must outlive command-buffer submission.
+llvm::Error buildPipelineAccelerationStructures(
+    Device &Dev, ComputeEncoder &Enc, Pipeline &P,
+    llvm::SmallVectorImpl<std::unique_ptr<AccelerationStructure>> &OutAS,
+    llvm::SmallVectorImpl<std::unique_ptr<Buffer>> &OutInputBuffers);
```
manon-traverse, line 343: `Pipeline` is not a type we should be using the rendering backend API.

The test frame work should use the rendering backend API, but any new rendering backend code should not have the test framework as a dependency.

manon-traverse: Discussed IRL: Will leave it like this for now since `executeProgram` still cannot be moved out of the `Device` type.

A comment should be added that this is temporarily living in the wrong place until we can move `executeProgram` outside of the `Device` type.

lib/API/Device.cpp
```diff
@@ -94,6 +95,134 @@ offloadtest::createRenderTargetFromCPUBuffer(Device &Dev,
...
+    llvm::SmallVectorImpl<std::unique_ptr<Buffer>> &OutInputBuffers) {
+  if (P.AccelStructs.BLAS.empty() && P.AccelStructs.TLAS.empty())
+    return llvm::Error::success();
+
+  // Storage upload-heap buffer flags. Backends that need extra native flags
+  // for AS-build inputs (e.g. Vulkan SHADER_DEVICE_ADDRESS +
+  // ACCEL_BUILD_INPUT_READ_ONLY) add them implicitly when ray tracing is
+  // supported.
+  const BufferCreateDesc UploadDesc{MemoryLocation::CpuToGpu,
+                                    BufferUsage::Storage};
```
manon-traverse: Would there be an advantage to adding a `BufferUsage::AsInput` enum variant? Or in other words, are there downsides to adding the `SHADER_DEVICE_ADDRESS +
 ACCEL_BUILD_INPUT_READ_ONLY` flags?

If not, we can just say `Storage` is the usage you need for acceleration structure build input buffers.

lib/API/DX/Device.cpp
```diff
@@ -2805,6 +2823,157 @@ class DXDevice : public offloadtest::Device {
...
+
+  // BuildRaytracingAccelerationStructure() lives on ID3D12GraphicsCommandList4.
+  ComPtr<ID3D12GraphicsCommandList4> CmdList4;
+  if (auto Err = HR::toError(CB.CmdList.As(&CmdList4),
+                             "Failed to query ID3D12GraphicsCommandList4."))
+    return Err;
+
+  // Flush a pending barrier before reading, like dispatch(): a TLAS build must
+  // observe BLASes built in the previous batch.
+  CB.flushBarrier();
```
manon-traverse, line 2930: Don't we also need to add a dst barrier so we actually make the writes visible to the acceleration structure build?

lib/API/MTL/MTLDevice.cpp
```diff
@@ -373,9 +373,18 @@ class MTLRenderPass final : public offloadtest::RenderPass {
...
 class MTLCommandBuffer : public offloadtest::CommandBuffer {
 public:
   MTL::CommandBuffer *CmdBuffer = nullptr;
+  /// Back-pointer to the owning device; used by encoders that need to
+  /// allocate scratch / instance buffers for AS builds.
+  MTLDevice *Dev = nullptr;
+  /// MTL::Buffer objects (e.g. AS scratch and TLAS instance buffers) that
+  /// must outlive command-buffer submission. Released when the CB is
+  /// destroyed.
+  llvm::SmallVector<MTL::Buffer *> KeepAliveMTLBuffers;
```
manon-traverse: Is it possible to make the scratch and TLAS instance buffers actual `offloadtest::Buffer` type? And if possible, does that make sense to do? (Just thinking out loud)

manon-traverse (approved):
Just a couple of questions, but those could all be addressed in follow-up PRs. (Apart from a question about barriers)

lib/API/DX/Device.cpp
manon-traverse: Wait this is legacy barriers. Do we need to UAV barrier or a state transition?

include/API/Device.h
```diff
@@ -25,6 +25,7 @@
 #include "Support/Pipeline.h"
 
 #include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/StringMap.h"
```
bogner: I believe this is only used in the .cpp, so this include is redundant

lib/API/DX/Device.cpp
```diff
@@ -1719,32 +1706,38 @@ class DXDevice : public offloadtest::Device {
                 &HeapProps, D3D12_HEAP_FLAG_NONE, &BufferDesc,
                 D3D12_RESOURCE_STATE_RAYTRACING_ACCELERATION_STRUCTURE, nullptr,
                 IID_PPV_ARGS(&ASBuffer)),
-            "Failed to create BLAS resource."))
+            llvm::formatv("Failed to create {0} resource.", Kind)
+                .str()
+                .c_str()))
```
bogner: This is unsafe - the `std::string` returned from `.str()`'s lifetime isn't sufficient for the call.

We could probably change the signature of `HR::toError` to `llvm::Error toError(HRESULT HR, const llvm::Twine &Msg)` and just pass `llvm::formatv(...)` to it directly (the `formatv_object` itself would be the temporary then, rather than a raw pointer, and thus lifetime extended).

Note that this issue occurs several times throughout the patch.

lib/API/Device.cpp
MarijnS95: SDA is generally useful; `ACCEL_BUILD_INPUT_READ_ONLY` is also made default on Breda because there don't seem to be any repercussions.

These are "regular" buffers with index/vertex/instance/transform data.

lib/API/DX/Device.cpp
MarijnS95: Good catch. Besides, [this doc-snippet](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#synchronizing-acceleration-structure-memory-writesreads) also seems to answer our question from https://github.com/llvm/offload-test-suite/pull/1282#discussion_r3359146456:

> Given that acceleration structures must always be in [D3D12_RESOURCE_STATE_RAYTRACING_ACCELERATION_STRUCTURE](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#additional-resource-states) as described above, resource state transitions can’t be used to synchronize between writes and reads (and vice versa) of acceleration structure data. Instead, the way to accomplish this is using UAV barriers on resources holding acceleration structure data between operations that write to an acceleration structure (such as [BuildRaytracingAccelerationStructure()](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#buildraytracingaccelerationstructure) and operations that read from them (such as [DispatchRays()](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#dispatchrays) (and vice versa).

So there must be a UAV barrier. The tail of `batchBuildAS()` already covers this:
https://github.com/llvm/offload-test-suite/pull/1244/changes/BASE..d61084a909f24a9cf45f3c30519d77ad3bad2f17#diff-70fee035163d029347502582085e489e1c1cc30c901b95dbfa7aaf346a0f76afR2973-R2974

lib/API/DX/Device.cpp
MarijnS95: Seems like I need to read up on C++ lifetimes again. I'm used to Rust where all lifetimes in a statement are extended until the end of that statement. Thus this `formatv().str()` was expected to live until the end of the `if` that surrounds it, even if `.c_str()` is called on top of the temporary.

However the `c_str()` pattern introduced by me here is arbitrary and wrong. There are other APIs / call-sites where an `llvm::StringRef` is _received_, but a `formatv().str()` (**without `.c_str()`**) is passed. That should have equally solved the lifetime if the resulting `.str()` was passed into `toError()` right?

Although `llvm::Twine` sounds nicer since `toError()` is going to reformat the string anyway, at least on Vulkan (where it should simply append the twine and flatten the string only once).

MarijnS95: Looking through the codebase there are quite a few places where `c_str()` is used to pass `std::string` into the `createStringError()` too; doesn't that support formatting the strings already (that `llvm::formatv()` call) without it?

lib/API/DX/Device.cpp
manon-traverse: Thanks for double checking this! Sounds like it's good to go!

lib/API/DX/Device.cpp
bogner: Ugh, I was wrong here - The `char *` being passed around triggered a red flag for me and I crossed some wires on why. The temporary is indeed destroyed at the end of the full expression ([[class.temporary] 4](https://timsong-cpp.github.io/cppwp/n4659/class.temporary#4)), so such usage was perfectly fine as long as we don't store the pointer.

Even so, the Twine is more efficient and clearer about what we're doing.

MarijnS95: Awesome! I'm glad we managed to clean this up either way, the `.c_str()` call was redundant to pass an `llvm::StringRef` and indeed `Twine` is more efficient and makes more logical sense.

PR #1245, opened 2026-05-28
Bind acceleration structures and enable the InlineRT tests

### LINKED ISSUE #1158: Build Acceleration Structure infrastructure for Ray Tracing tests

MarijnS95:
Closes https://github.com/llvm/offload-test-suite/issues/1158 🥳

## Summary
Wire up acceleration-structure descriptor binding end-to-end across all three backends so shaders can actually consume the TLAS that `buildPipelineAccelerationStructures()` produced — completing the stack and promoting the three InlineRT tests from XFAIL to passing.

Per-resource AS handling lands in a new per-backend `createAS()` (paired with `createSRV()` / `createUAV()` / `createCBV()`): a pure single-create that queries TLAS sizes via `Dev.getTLASBuildSizes()` and allocates the handle via `Dev.createTLAS()`, returning the `unique_ptr` to the caller. No `InvocationState` or `Pipeline` access — the multi-create (`createBuffers()` / `createResources()`) records the handle in `InvocationState::TLASes` (a `StringMap` keyed by `TLASDesc::Name`) and wires a non-owning AS pointer into the per-resource bundle the binding loop reads. The shared AS-build helper picks up that map and walks `P.AccelStructs.TLAS` to pair each YAML descriptor with its pre-allocated handle by name (TLASes without a map entry are skipped, i.e. declared but unbound). BLAS handles are still allocated by the helper itself since BLASes aren't user-bindable.

`executeProgram()` in each backend now runs as:

- `createBuffers` / `createResources` (`createAS()` allocates TLAS handles)
- open encoder → `buildPipelineAccelerationStructures()` → end

- **Vulkan**: `createDescriptorPool()` counts AS descriptors in a separate scalar (the KHR enum value `1000150000` doesn't fit in the indexed array used for the core types) and emits one `VkDescriptorPoolSize` for them. `createDescriptorSets()` reads the resolved `VulkanAccelerationStructure` handle from `ResourceRef.AS` (populated by `createResources()`) and writes it through a `VkWriteDescriptorSetAccelerationStructureKHR` chained on the descriptor write's `pNext`. The dispatch's pre-barrier dst access now includes `VK_ACCESS_ACCELERATION_STRUCTURE_READ_BIT_KHR` so the prior AS-build's writes are visible to the shader's RayQuery reads. Device creation enables `VK_KHR_ray_query` using the same chain-pre-query + error-on-flag-mismatch pattern that #1232 set up for the AS / BDA extensions — without `VK_KHR_ray_query` enabled the shader's `OpRayQueryProceedKHR` instructions silently no-op and `Output` reads back zero. `copyResourceDataToDevice()` short-circuits AS bundles via a new `ResourceBundle::isAccelerationStructure()` predicate (no host buffer to barrier).
- **DX12**: writes a `D3D12_SRV_DIMENSION_RAYTRACING_ACCELERATION_STRUCTURE` SRV with the AS GPU virtual address as `Location` into the heap slot that `createBuffers()` reserved (`CreateShaderResourceView()` with a null resource — the AS data lives in the buffer pointed to by `Location`).
- **Metal**: the Metal shader converter doesn't bind the AS directly; the shader reads a buffer containing an `IRRaytracingAccelerationStructureGPUHeader` that holds the AS's `gpuResourceID` plus a pointer to an instance-contributions array. `createBuffers()` allocates and fills both buffers per AS-descriptor entry, then points the descriptor at the header buffer's GPU address. The TLAS itself is built with the `UserID` instance-descriptor variant so HLSL `CommittedInstanceID()` returns the YAML-specified per-instance ID instead of the array index.

The three InlineRT tests now actually exercise the AS end-to-end: `TraceRayInline()` issues a RayQuery against `Scene` and writes a hit-dependent value into `Output` (the instance ID for `multi-instance`, 1/0 otherwise). The catch-all `XFAIL: *` is dropped; `XFAIL: Clang` remains. The test shaders also gain explicit `[[vk::binding]]` annotations because dxc's default HLSL→SPIR-V binding mapping collides `Scene`'s `t0` with `Output`'s `u0` at binding 0, which VVL flags as a descriptor type mismatch.

## Test plan

Local on an NVIDIA RTX 3060:

CI (RT-capable runners):
- [ ] macOS Metal (`supportsRaytracing`)

### DIFF
--- a/lib/API/VK/Device.cpp
+++ b/lib/API/VK/Device.cpp
@@ -838,8 +838,11 @@ public:
                        uint32_t GroupCountX, uint32_t GroupCountY,
                        uint32_t GroupCountZ) override {
     const auto &VKPSO = llvm::cast<VulkanPipelineState>(PSO);
+    // Include AS_READ so dispatches that issue RayQuery against a TLAS get a
+    // proper hand-off from the prior AS-build's AS_WRITE.
     addDstBarrier(VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT,
-                  VK_ACCESS_SHADER_WRITE_BIT | VK_ACCESS_SHADER_READ_BIT);
+                  VK_ACCESS_SHADER_WRITE_BIT | VK_ACCESS_SHADER_READ_BIT |
+                      VK_ACCESS_ACCELERATION_STRUCTURE_READ_BIT_KHR);
     insertDebugSignpost(llvm::formatv("Dispatch [{0},{1},{2}]", GroupCountX,
                                       GroupCountY, GroupCountZ)
                             .str());
@@ -1199,12 +1202,15 @@ private:
   };
 
   struct ResourceRef {
-    ResourceRef(BufferRef H, BufferRef D) : Host(H), Device(D) {}
-    ResourceRef(BufferRef H, ImageRef I) : Host(H), Image(I) {}
+    explicit ResourceRef(BufferRef H, BufferRef D) : Host(H), Device(D) {}
+    explicit ResourceRef(BufferRef H, ImageRef I) : Host(H), Image(I) {}
+    explicit ResourceRef(VulkanAccelerationStructure *AS) : AS(AS) {}
 
-    BufferRef Host;
-    BufferRef Device;
-    ImageRef Image;
+    BufferRef Host{};
+    BufferRef Device{};
+    ImageRef Image{};
+    // AS-only; mutually exclusive with the buffer/image fields above.
+    VulkanAccelerationStructure *AS = nullptr;
   };
 
   struct ResourceBundle {
@@ -1220,6 +1226,10 @@ private:
 
     bool isSampler() const {
       return DescriptorType == VK_DESCRIPTOR_TYPE_SAMPLER;
+    }
+
+    bool isAccelerationStructure() const {
+      return DescriptorType == VK_DESCRIPTOR_TYPE_ACCELERATION_STRUCTURE_KHR;
     }
 
     bool isBuffer() const {
@@ -1266,9 +1276,11 @@ private:
     llvm::SmallVector<VkBufferView> BufferViews;
     llvm::SmallVector<VkImageView> ImageViews;
 
-    // Built acceleration structures, kept alive for the pipeline lifetime.
+    // Parallel-indexed to `P.AccelStructs.BLAS`.
     llvm::SmallVector<std::unique_ptr<offloadtest::AccelerationStructure>>
-        AccelStructs;
+        BLASes;
+    // Keyed by `TLASDesc::Name`.
+    llvm::StringMap<std::unique_ptr<offloadtest::AccelerationStructure>> TLASes;
     // Vertex/index buffers consumed during AS builds; must outlive submission.
     llvm::SmallVector<std::unique_ptr<offloadtest::Buffer>> ASInputBuffers;
   };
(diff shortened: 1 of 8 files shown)


test/Feature/InlineRT/triangle-setup.test
```diff
@@ -1,12 +1,19 @@
 #--- source.hlsl
 
-RaytracingAccelerationStructure Scene : register(t0);
-RWStructuredBuffer<uint> Output : register(u0);
+[[vk::binding(0, 0)]] RaytracingAccelerationStructure Scene : register(t0);
```
EmilioLaiso, line 3:
```suggestion
[[vk::binding(0, 0)]] RaytracingAccelerationStructure Tlas : register(t0);
```

Not a required change, just a preference. Feel free to ignore 😬

lib/API/VK/Device.cpp
```diff
@@ -838,8 +838,11 @@ class VKComputeEncoder : public offloadtest::ComputeEncoder {
                        uint32_t GroupCountX, uint32_t GroupCountY,
                        uint32_t GroupCountZ) override {
     const auto &VKPSO = llvm::cast<VulkanPipelineState>(PSO);
+    // Include AS_READ so dispatches that issue RayQuery against a TLAS get a
+    // proper hand-off from the prior AS-build's AS_WRITE.
```
EmilioLaiso, line 842: Does the Dx12 backend require any resource transition?

test/Feature/InlineRT/triangle-setup.test
MarijnS95: Will keep `Scene` — it's the conventional name in the DXR samples and HLSL spec text, and the variable holds the whole scene (TLAS + indirected BLASes), not just a TLAS. Happy to revisit if there's a stronger reason.

lib/API/VK/Device.cpp
MarijnS95: Per the [DXR spec on synchronizing AS memory writes/reads](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#synchronizing-acceleration-structure-memory-writesreads): AS resources are permanently in `D3D12_RESOURCE_STATE_RAYTRACING_ACCELERATION_STRUCTURE`, so resource-state transitions can't be used to synchronize AS — only UAV barriers can.

The DX backend already issues that UAV barrier: in the now-merged #1244, the tail of `DXComputeEncoder::batchBuildAS()` calls `CB.addPendingUAVBarrier()`, which the next dispatch flushes via `flushBarrier()`. Same shape as the VK code at line 842, just expressed as a UAV barrier instead of an access-mask barrier.

EmilioLaiso (approved):

manon-traverse (approved):

PR #1246, opened 2026-05-28
Add the mesh shader pipeline creation function to the Device base class.

manon-traverse:
The mesh shader pipeline creation function was not part of the Device base class. This PR adds it to the base class, and also introduces a create description similar to the traditional raster pipeline creation function.

Additionally support for specifying the topology has also been added. This is only needed in DX12 since both Vulkan and Metal just extract the topology type from the shader byte code.
Technically DX12 only cares about the topology type (triangles vs lines, instead triangle list vs triangle strip), but I think it is overkill to introduce a new enum for topology types to the .yaml file if we can just use a topology type instead.

### DIFF
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -124,6 +124,36 @@ struct TraditionalRasterPipelineCreateDesc {
   }
 };
 
+struct MeshShaderRasterPipelineCreateDesc {
+  llvm::SmallVector<Format> RTFormats;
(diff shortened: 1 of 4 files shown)


EmilioLaiso (approved):
LGTM

MarijnS95 (approved):

bogner (approved):

PR #1250, opened 2026-05-28
[NFC][offloader] Mark run() definition static to match its declaration

alsepkow:
Minor nit: `run()` in `tools/offloader/offloader.cpp` is forward-declared `static` but the definition omits the specifier. Adding it matches the other two helpers in this file and the readability rationale in the [LLVM Coding Standards](https://llvm.org/docs/CodingStandards.html#restrict-visibility). NFC.

Assisted by Claude Opus 4.7.

### DIFF
--- a/tools/offloader/offloader.cpp
+++ b/tools/offloader/offloader.cpp
@@ -96,7 +96,7 @@ static bool matchesRegexIgnoreCase(StringRef GPUDescription,
   return R.isValid() && R.match(GPUDescription);
 }
 
-int run() {
+static int run() {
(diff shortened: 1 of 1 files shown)


alsepkow:
CI builds are failing. Need [this fix](https://github.com/llvm/offload-test-suite/pull/1251)

bob80905 (approved):

bogner (approved):
Arguably we should also move `main` to the end of the file - the forward declaration of run here is pretty silly. But in any case, this change as is is correct, so that can be some later cleanup.

PR #1254, opened 2026-06-01
Full traditional raster pipeline test

EmilioLaiso:
With the recent work landing Hull/Domain and Geometry shader support, the traditional raster pipeline is now complete. 
This PR adds a single test that exercises every stage of it in one pass and validates the result against the existing `SimpleTriangle.png` golden image.

### Pipeline

Starting from a single point, the test drives all five raster stages:

1 point (triangle centroid "G" for a "big" triangle mapping to the one used in `SimpleTriangle.png`)
→ Vertex passthrough
→ Hull quad domain, all tess factors = 1, outputtopology("point")
→ Tessellator → 4 points - (topology change: patch → points)
→ Domain tag each point as sub-triangle 0..3, place at its centroid
→ Geometry expand each centroid → the sub-triangles (topology change: points → triangles)
→ Pixel outputs colour

Here's a very non-precise drawing to visualize. The red dot is G, our input point.
image

### DIFF
--- a/test/Graphics/FullTraditionalRasterPipelineSimpleTriangle.test
+++ b/test/Graphics/FullTraditionalRasterPipelineSimpleTriangle.test
@@ -232,8 +232,10 @@ DescriptorSets: []
 #--- end
 
 # Metal has no native geometry shader stage, and its tessellation maps HS/DS
-# onto a compute kernel + post-tessellation vertex function that isn't wired up
-# here. Skip this backend entirely.
+# onto a compute kernel + post-tessellation vertex function. Metal Shader
+# Converter can emulate both stages
+# (https://developer.apple.com/metal/shader-converter/#emulating-geometry), so
+# this is technically possible, but that emulation path isn't wired up here.
 # UNSUPPORTED: Metal
 
 # XFAIL: Clang
--- a/test/Graphics/FullTraditionalRasterPipelineSimpleTriangle.test
+++ b/test/Graphics/FullTraditionalRasterPipelineSimpleTriangle.test
@@ -1,0 +1,249 @@
+# Exercises the full traditional raster pipeline in one pass:
+#   Vertex -> Hull -> Domain -> Geometry -> Pixel
+# starting from a single point. The tessellator amplifies that point into four
+# points (one per sub-triangle of the midpoint subdivision) and the geometry
+# shader expands each into a triangle -- two primitive-topology transitions
+# (patch -> points, points -> triangles). The four sub-triangles tile the
+# original triangle, so the result is validated against the same golden image.
+
+#--- vertex.hlsl
+struct VSOutput
+{
+    float4 position : POSITION;
+};
+
+VSOutput main(float4 position : POSITION)
+{
+    VSOutput o;
+    o.position = position;
+    return o;
+}
+
+#--- hull.hlsl
+struct HSInput
+{
+    float4 position : POSITION;
+};
+
+struct HSOutput
+{
+    float4 position : POSITION;
+};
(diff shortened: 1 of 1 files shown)


test/Graphics/FullTraditionalRasterPipelineSimpleTriangle.test
```diff
@@ -0,0 +1,249 @@
...
+---
+- Type: PixelPercent
+  Val: 0.2 # No more than 0.2% of pixels may be visibly different.
+...
+#--- end
+
+# Metal has no native geometry shader stage, and its tessellation maps HS/DS
+# onto a compute kernel + post-tessellation vertex function that isn't wired up
+# here. Skip this backend entirely.
+# UNSUPPORTED: Metal
```
MarijnS95, line 253: We use Metal IRConverter, which technically enables this:

https://developer.apple.com/metal/shader-converter/#emulating-geometry

I think we can update the comment with this link, clarifying your statement that it's technically possible via emulation but not hooked up?

MarijnS95 (approved):

test/Graphics/FullTraditionalRasterPipelineSimpleTriangle.test
```diff
@@ -232,8 +232,10 @@ DescriptorSets: []
 #--- end
 
 # Metal has no native geometry shader stage, and its tessellation maps HS/DS
-# onto a compute kernel + post-tessellation vertex function that isn't wired up
-# here. Skip this backend entirely.
+# onto a compute kernel + post-tessellation vertex function. Metal Shader
+# Converter can emulate both stages
+# (https://developer.apple.com/metal/shader-converter/#emulating-geometry), so
+# this is technically possible, but that emulation path isn't wired up here.
 # UNSUPPORTED: Metal
```
MarijnS95, line 253: On a side-note (not a blocker, just a question) I'm curious if this should remain `UNSUPPORTED` (i.e. no intention to support this, don't even try to run it) or `XFAIL` (run the test on this platform, currently failing, might want to look into enabling/supporting it).

At least that's my understanding of the two 😅

bogner (approved):
Generally looks good. It would be nice to clang-format each of the hlsl inputs so that it matches LLVM style (though do note that you'll have to remove or temporarily modify the shader attributes to (eg) `[[domain("quad")]]` since clang-format stumbles over MS style attributes in hlsl still)

test/Graphics/FullTraditionalRasterPipelineSimpleTriangle.test
manon-traverse: I think it should be XFAIL as adding support for this at some point would make sense.

test/Graphics/FullTraditionalRasterPipelineSimpleTriangle.test
```diff
@@ -0,0 +1,265 @@
...
+    Entry: main
+  - Stage: Domain
+    Entry: main
+  - Stage: Geometry
+    Entry: main
+  - Stage: Pixel
+    Entry: main
+Buffers:
+  # A single control point at the big triangle centroid: ((0 + 0.25 - 0.25)/3,
+  # (0.25 - 0.25 - 0.25)/3) = (0, -1/12). Everything downstream is rebuilt from
```
manon-traverse, line 215:
```
  -1/12
```
The sum of all natural numbers!
https://en.wikipedia.org/wiki/1_%2B_2_%2B_3_%2B_4_%2B_%E2%8B%AF

manon-traverse (approved):

PR #1261, opened 2026-06-02
Fix stride and fill size of Out buffer in cbuffer/vectors-16bit.test.

manon-traverse:
The stride and fill size were incorrect in the `Feature/CBuffer/vectors-16bit.test` test.

On my local machine (where I am doing lots of fixes on buffer and texture creation/management) this is not an issue on WARP either. If it fails CI on this I will convert it to draft and we can then merge it at a later point.


bogner (approved):
Yeah I don't remember the context but I'm reasonably sure that comment is outdated

EmilioLaiso (approved):

PR #1269, opened 2026-06-03
[VK] Gate debug utils messenger on runtime config instead of NDEBUG

MarijnS95:
`registerDebugUtilCallback` was only invoked inside `#ifndef NDEBUG`, leaving the function unused (and triggering `-Wunused-function`) in `NDEBUG` builds. The extension registration itself had already been converted to runtime config flags in #382; the call site was missed.

Tie the callback to whether `VK_EXT_debug_utils` was actually enabled, and pull the extension in whenever either `EnableDebugLayer` or `EnableValidationLayer` is set — so validation messages route through the callback. This matches how the DX backend couples the two flags in `lib/API/DX/Device.cpp`.

### DIFF
--- a/lib/API/VK/Device.cpp
+++ b/lib/API/VK/Device.cpp
@@ -3864,10 +3864,13 @@ llvm::Error offloadtest::initializeVulkanDevices(
   }
   const llvm::SmallVector<VkExtensionProperties, 0> AvailableExtensions =
       queryInstanceExtensions(nullptr);
-  if (Config.EnableDebugLayer) {
+  bool DebugUtilsEnabled = false;
+  if (Config.EnableDebugLayer || Config.EnableValidationLayer) {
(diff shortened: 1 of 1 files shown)


MarijnS95:
I.e. I've been seeing this on `cmake`-default debug builds:

```
LLVM/OffloadTest/lib/API/VK/Device.cpp:287:33: warning: ‘VkDebugUtilsMessengerEXT_T* registerDebugUtilCallback(VkInstance)’ defined but not used [-Wunused-function]
  287 | static VkDebugUtilsMessengerEXT registerDebugUtilCallback(VkInstance Instance) {
      |                                 ^~~~~~~~~~~~~~~~~~~~~~~~~
```

EmilioLaiso (approved):

bogner (approved):

PR #1270, opened 2026-06-03
Add RayTracing pipeline kind, shader stages, and YAML schema

MarijnS95:
Depends on #1245

## Summary

Foundational PR in the PSO-based raytracing bring-up series tracked in #1268. Stacks on top of #1245 (which depends on #1244, which depends on #1232) — only the top commit on this branch is new; the rest are the inline-RT bring-up already in review.

Lays out the framework-side surface needed by the upcoming backend PRs:

- `ShaderPipelineKind::RayTracing` plus six new `Stages` — `RayGeneration`, `Miss`, `ClosestHit`, `AnyHit`, `Intersection`, `Callable` — with `isRayTracingStage` / `Pipeline::isRayTracing()` helpers.
- YAML schema for an RT pipeline: `HitGroup` (Triangles | Procedural, ClosestHit + optional AnyHit / Intersection), `RayTracingPipelineConfig` (MaxTraceRecursionDepth, MaxPayloadSizeInBytes, MaxAttributeSizeInBytes, optional PipelineFlags), and `ShaderBindingTable` (raygen / miss / hit-group / callable records, each with optional reserved LocalRootData bytes).
- `validatePipelineKind` allows duplicate RT stages (a pipeline can have several miss / hit-group shaders, which the existing duplicate check would have rejected), requires at least one RayGeneration, and rejects mixing with Compute/Vertex/Mesh. The reverse check rejects HitGroups / RTConfig / SBT on any non-RT pipeline. `validateDispatchParameters` reinterprets `DispatchGroupCount` as `{Width, Height, Depth}` for the upcoming DispatchRays and forbids VertexCount on RT.
- Existing `Stages` switches across the backends grow the six RT cases — Vulkan maps each one to its `VK_SHADER_STAGE_*_KHR` bit ready for PR 2; Metal unreachables on RT (`metal_irconverter` takes a different route); raster pipeline `setShader` (Traditional + MeshShader variants) adds them to the existing unreachable group.
- Each backend's `executeProgram` gets a terminal `else if (P.isRayTracing())` that returns a "not yet supported on <backend>" error so PR2/3/4 just have to replace it.
- `%dxc_target_lib` lit substitution (same compiler binary, separate name for `-T lib_6_x` library targets); `raytracing-pipeline` available-feature gated on DX `RaytracingTier >= 1.0` and the Vulkan `VK_KHR_ray_tracing_pipeline` extension being reported by the device.
[...]

Issue #1268, opened 2026-06-03
[EPIC]: Bring-up and test coverage for PSO-based raytracing


MarijnS95:
Companion to the inline-RT epic in #1258. That issue is scoped to `RayQuery` / `TraceRayInline` inside a regular compute shader; the test framework already treats it as a `Compute` pipeline that happens to bind an acceleration structure. PSO-based raytracing is a different beast — it introduces an entirely new pipeline kind, new shader stages, a shader binding table, and the `DispatchRays` command. The framework has none of that today, so this issue tracks both the bring-up work and the shader-visible test surface that becomes unlockable as bring-up progresses.

Reuses the AS YAML / abstraction layers from the inline-RT bring-up (#1214, #1232, #1245), so this work starts above that line.

Legend:
- [ ] Not started
- [ ] In progress 🏗️
- [ ] In review 👀
- [x] Finished (merged in main)
- 👍 testable on top of the bring-up below
- 🏗️ requires additional framework work first

## Bring-up PRs (foundation for everything below)
- [ ] Add `ShaderPipelineKind::RayTracing`, six new RT `Stages`, HitGroups / RayTracingPipelineConfig / ShaderBindingTable YAML schema, `%dxc_target_lib` substitution, the `raytracing-pipeline` lit feature, and a foundational `XFAIL: *` smoke test 👀
    - #1270
- [ ] Vulkan RT pipeline + SBT + `dispatchRays` (drops Vulkan from the smoke test's XFAIL list) 👀
    - #1273
- [ ] D3D12 state object + SBT + `DispatchRays` (drops DirectX from the smoke test's XFAIL list) 👀
    - #1275
- [ ] Metal RT pipeline + SBT + `dispatchRays` via `metal_irconverter` (drops Metal from the smoke test's XFAIL list) 👀
    - #1281
- [ ] First PSO RT test batch: DispatchRays index/dimensions + SBT miss/hit-group routing 👀
    - #1277
- [ ] ClosestHit ray system values: barycentrics, PrimitiveIndex, WorldRay 👀
    - #1278
- [ ] PSO-only RT surface: TraceRay recursion, RAY_FLAG_SKIP_CLOSEST_HIT_SHADER, callable shaders 👀
    - #1279

### Pipeline / stage model
- [ ] Add `ShaderPipelineKind::RayTracing` alongside `Compute` / `TraditionalRaster` / `MeshShaderRaster` in `Support/Pipeline.h` 👀
[...]

PR #1273, opened 2026-06-03
[VK] Add ray tracing pipeline, SBT, and DispatchRays bring-up

MarijnS95:
## Summary

First per-backend bring-up in the PSO raytracing series (#1268). Stacks on top of #1270 (foundational schema + lit infrastructure + XFAILed test). Adds the API surface needed by the upcoming D3D12 and Metal PRs plus the Vulkan implementation behind it.

API surface:

- `ComputeEncoder::dispatchRays(PSO, SBT, W, H, D)` virtual on the existing compute encoder (no separate `RayTracingEncoder`).
- `Device::createPipelineRT` + `Device::createShaderBindingTable` virtuals with a new `RayTracingPipelineCreateDesc` carrying the DXIL library blob, the shader entry points (Stage + EntryPoint), the hit-group list, and the `RayTracingPipelineConfig`.
- `include/API/ShaderBindingTable.h` holding the abstract runtime base; backend SBT classes derive from it with LLVM-style `classof` / `cast<>`.
- Rename: PR #1270's YAML struct `ShaderBindingTable` → `ShaderBindingTableDesc` so the bare name is free for the runtime class (parallel to `BLASDesc` / `TLASDesc` vs `AccelerationStructure`). YAML key stays `ShaderBindingTable:`.
- D3D12 and Metal stub the new methods with not-yet-supported errors; their bring-up lands in follow-up PRs.

Vulkan implementation:

- The pre-existing `RaytracingFunctions RT` struct lumped AS and RT-pipeline entry points together; they split into `ASFunctions AS` + `RTPipelineFunctions RT` so the names match the actual feature-gate split (AS + ray-query is a complete configuration; RT pipeline layers on top). `HasRayTracingSupport` renames to `HasASSupport`; `HasRTPipelineSupport` tracks the new extension.
- `VK_KHR_ray_tracing_pipeline` is requested when reported, with `VkPhysicalDeviceRayTracingPipelineFeaturesKHR` chained pre-query and the gating `rayTracingPipeline` bool checked post-query (matches the AS / BDA pattern from #1232). Sub-features the tests don't exercise (capture-replay / indirect-trace / traversal-primitive-culling) are cleared.
- Function pointers `vkCreateRayTracingPipelinesKHR`, `vkGetRayTracingShaderGroupHandlesKHR`, `vkCmdTraceRaysKHR` resolve once at device creation. `VkPhysicalDeviceRayTracingPipelinePropertiesKHR` is cached at the same time for SBT handle size / alignment / base alignment.
- `VKRayTracingPipelineState` derives from `VulkanPipelineState`; an `IsRayTracing` flag on the base lets the existing Vulkan `cast<>` path stay polymorphic without adding a new `GPUAPI` value. The derived class also carries a `StringMap<uint32_t>` resolving each shader `EntryPoint` or hit-group `Name` to its index in the pipeline's group array, plus per-bucket counts so the SBT builder can slice the contiguous handle blob into raygen / miss / hit / callable regions.
- `createPipelineRT` builds a single `VkShaderModule` (the DXIL library compiles to one SPIR-V module with multiple `OpEntryPoint`s), one `VkPipelineShaderStageCreateInfo` per `Shader` entry, and one `VkRayTracingShaderGroupCreateInfoKHR` per general shader / hit group. Pipeline layout uses the same `createPipelineLayout` helper as the compute path, gated on all six RT stage flags so any binding can be consumed from any RT shader.
- `createShaderBindingTable` allocates a host-visible coherent buffer big enough for four regions, then lays out each entry as `[handle bytes][LocalRootData bytes][padding-to-stride]`. Per-region stride = `align(handleSize + max-LocalRootData-in-region, handleAlignment)`; per-region size = `align(count * stride, baseAlignment)`. LocalRootData support comes for free from PR #1270's SBT schema; the test doesn't exercise it yet. Each region's `VkStridedDeviceAddressRegionKHR` derives from the buffer's `vkGetBufferDeviceAddress`.
- `dispatchRays` binds the pipeline at `VK_PIPELINE_BIND_POINT_RAY_TRACING_KHR`, emits a pre-barrier with `ACCELERATION_STRUCTURE_READ_BIT_KHR | SHADER_READ_BIT | SHADER_WRITE_BIT` dst access into `RAY_TRACING_SHADER_BIT_KHR`, then calls `vkCmdTraceRaysKHR` with the SBT's four region structs.
- `createCommands` picks the new bind point for RT pipelines so `vkCmdBindDescriptorSets` binds to the right point. `executeProgram`'s `isRayTracing` branch builds a `RayTracingPipelineCreateDesc` from the `Pipeline`, calls `createPipelineRT` then `createShaderBindingTable`, and keeps both on `InvocationState` for the dispatch.

Test side: `raygen-roundtrip.test`'s `XFAIL` becomes `Clang, DirectX, Metal`. On a DXC + Vulkan combo with the device reporting `VK_KHR_ray_tracing_pipeline` this should PASS; the Clang token still catches the compile failure on the Linux + `clang-dxc` loop where `[shader("raygeneration")]` doesn't yet lower to SPIR-V.

## Test plan

Local on an NVIDIA RTX 3060:

CI (RT-capable runners):

### DIFF
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -20,6 +20,7 @@
 #include "API/Capabilities.h"
 #include "API/CommandBuffer.h"
 #include "API/RenderPass.h"
+#include "API/ShaderBindingTable.h"
 #include "API/Texture.h"
 
 #include "Support/Pipeline.h"
@@ -82,6 +83,21 @@ struct ShaderContainer {
   std::string EntryPoint;
   const llvm::MemoryBuffer *Shader;
   llvm::SmallVector<SpecializationConstant> SpecializationConstants;
+};
+
+struct RayTracingShader {
+  Stages Stage;
+  std::string EntryPoint;
+};
+
+struct RayTracingPipelineCreateDesc {
+  // All RT shaders are compiled into a single DXIL library; every entry in
+  // `Shaders` references this same blob via the backend's library-loading
+  // path.
+  const llvm::MemoryBuffer *Library = nullptr;
+  llvm::SmallVector<RayTracingShader> Shaders;
+  llvm::SmallVector<HitGroup> HitGroups;
+  RayTracingPipelineConfig Config;
 };
 
 struct TraditionalRasterPipelineCreateDesc {
@@ -258,6 +274,14 @@ public:
   createMeshShaderRasterPipeline(
       llvm::StringRef Name, const BindingsDesc &BindingsDesc,
       const MeshShaderRasterPipelineCreateDesc &Desc) = 0;
+
+  virtual llvm::Expected<std::unique_ptr<PipelineState>>
+  createPipelineRT(llvm::StringRef Name, const BindingsDesc &BindingsDesc,
+                   const RayTracingPipelineCreateDesc &Desc) = 0;
+
+  virtual llvm::Expected<std::unique_ptr<ShaderBindingTable>>
+  createShaderBindingTable(const PipelineState &PSO,
+                           const ShaderBindingTableDesc &Desc) = 0;
 
   virtual llvm::Expected<std::unique_ptr<Fence>>
   createFence(llvm::StringRef Name) = 0;
--- a/include/API/Device.h
+++ b/include/API/Device.h
@@ -91,9 +91,9 @@ struct RayTracingShader {
 };
 
 struct RayTracingPipelineCreateDesc {
-  // All RT shaders are compiled into a single DXIL library; every entry in
-  // `Shaders` references this same blob via the backend's library-loading
-  // path.
+  // All RT shaders are compiled into a single shader library.
+  // Every entry in `Shaders` references this same blob via the backend's
+  // library-loading path.
   const llvm::MemoryBuffer *Library = nullptr;
   llvm::SmallVector<RayTracingShader> Shaders;
   llvm::SmallVector<HitGroup> HitGroups;
--- a/tools/offloader/offloader.cpp
+++ b/tools/offloader/offloader.cpp
@@ -113,16 +113,27 @@ static int run() {
   YIn >> PipelineDesc;
   ExitOnErr(llvm::errorCodeToError(YIn.error()));
 
-  // Read in the shaders
-  for (size_t I = 0; I < InputShader.size(); ++I) {
-    PipelineDesc.Shaders[I].Shader = readFile(InputShader[I]);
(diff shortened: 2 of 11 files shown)


include/API/Device.h
```diff
@@ -84,6 +85,21 @@ struct ShaderContainer {
...
+struct RayTracingShader {
+  Stages Stage;
+  std::string EntryPoint;
+};
+
+struct RayTracingPipelineCreateDesc {
+  // All RT shaders are compiled into a single DXIL library; every entry in
+  // `Shaders` references this same blob via the backend's library-loading
+  // path.
+  const llvm::MemoryBuffer *Library = nullptr;
```
manon-traverse: How does this work on Vulkan and Metal? There we don't have DXIL. Are they still stored in a single `Library`?

manon-traverse (approved):
Just one question but that might just be because of a badly written comment.

tools/offloader/offloader.cpp
```diff
@@ -113,17 +113,28 @@ static int run() {
...
+  // backend reads the blob from Shaders.front() and uses the rest only for
+  // their (Stage, Entry) metadata. Other pipelines expect one object file per
+  // stage.
+  if (PipelineDesc.isRayTracing()) {
+    if (InputShader.size() != 1)
+      ExitOnErr(createStringError(
+          std::errc::invalid_argument,
+          "RayTracing pipeline expects a single shader library, %d provided",
+          InputShader.size()));
+    PipelineDesc.Shaders.front().Shader = readFile(InputShader[0]);
```
manon-traverse, line 127: One more question, when we talk about a single shader library, does that mean one .hlsl file? We should definitely have support for splitting these shaders over multiple hlsl files.

include/API/Device.h
EmilioLaiso: The comment was just wrong / too DX12 specific, reworded it.

tools/offloader/offloader.cpp
EmilioLaiso: Indeed for now the limitation is one hlsl file

manon-traverse: Let's save this for a future PR then.

manon-traverse (approved):

EmilioLaiso (approved):

manon-traverse:
Only an irrelevant tests are failing, two people approved it, let's merge it.

PR #1275, opened 2026-06-03
[DX] Add ray tracing state object, SBT, and DispatchRays bring-up

MarijnS95:
Depends on #1273

## Summary

Second per-backend bring-up in the PSO raytracing series (#1268). Stacks on top of #1273 (the Vulkan bring-up + the API surface). Mirrors that PR for D3D12: builds an `ID3D12StateObject` from the YAML schema, hands out shader identifiers via `ID3D12StateObjectProperties`, lays out the SBT in an upload heap, and routes `DispatchRays` through `ID3D12GraphicsCommandList4` (same query path the AS-build already uses).

D3D12 implementation:

- `DXRayTracingPipelineState` derives from `DXPipelineState` with an `IsRayTracing` flag on the base for `classof` — matching the `VulkanPipelineState` pattern. It carries the `ID3D12StateObject`, a cached `ID3D12StateObjectProperties`, and a `StringMap<const void *>` resolving each raygen / miss / callable shader's `EntryPoint` and each hit-group's `Name` to its 32-byte shader identifier blob. The identifiers are driver-owned and stay alive for the `Properties` COM lifetime, so the PSO keeps `Properties` alive. Closest-hit / any-hit / intersection shaders are *not* directly addressable via `GetShaderIdentifier` — they're imported through a hit-group subobject — so the eager-cache loop skips them.
- `DXShaderBindingTable` holds a single upload-heap buffer plus four pre-built `D3D12_DISPATCH_RAYS_DESC` ranges (raygen, miss, hit-group, callable) — `RANGE` for raygen since it's always one record, and `RANGE_AND_STRIDE` for the others.
- `createPipelineRT` builds a `CD3DX12_STATE_OBJECT_DESC` with subobjects for the DXIL library (one export per `Shader` entry), per-hit-group subobjects with closest-hit / any-hit / intersection imports, the pipeline shader config (max payload + max attribute bytes), pipeline config (max recursion depth), and a global root signature subobject. The root signature comes from the library's embedded `RTS0` part when present, falling back to the `BindingsDesc` path (matching the existing compute / raster pipeline behaviour). Wide strings for the subobject exports live in a `SmallVector` that outlives the `SODesc`, since the helper classes store pointers into the strings rather than copying.
- `createShaderBindingTable` lays out each entry as `[identifier][LocalRootData][padding-to-stride]` with per-region stride = `align(D3D12_SHADER_IDENTIFIER_SIZE_IN_BYTES + max-LocalRootData-in-region, D3D12_RAYTRACING_SHADER_RECORD_BYTE_ALIGNMENT)` and per-region size = `align(count * stride, D3D12_RAYTRACING_SHADER_TABLE_BYTE_ALIGNMENT)`. The buffer lives in an upload heap with `D3D12_RESOURCE_STATE_GENERIC_READ` — PR3 simplification; a staging copy into a default heap is a follow-up.
- `dispatchRays` queries the underlying `CommandListX` for `ID3D12GraphicsCommandList4` (matching the AS-build path), binds the global root signature via `SetComputeRootSignature`, calls `SetPipelineState1` with the state object, and issues `DispatchRays` with a `D3D12_DISPATCH_RAYS_DESC` populated from the SBT's four ranges plus the dispatch dimensions. The descriptor heap + descriptor-table bindings are set up by the existing `createComputeCommands` helper before the encoder is created.
- `createComputeCommands` grows an `isRayTracing` branch at the dispatch point so it calls `dispatchRays` instead of `dispatch`, reusing all of the descriptor-heap and root-signature wiring. `InvocationState` carries a `ShaderBindingTable` `unique_ptr` that's only populated for RT pipelines.
- `executeProgram`'s `isRayTracing` branch builds a `RayTracingPipelineCreateDesc` from `Pipeline.Shaders` / `HitGroups` / `RTConfig`, calls `createPipelineRT` then `createShaderBindingTable`, then re-enters `createComputeCommands` which dispatches via the new RT path.

Test side: `raygen-roundtrip.test`'s `XFAIL` becomes `Clang, Metal` — DirectX should PASS via this implementation on Windows CI (and via Wine + vkd3d-proton locally on Linux). The Clang token still catches the compile failure on `clang-dxc` since `[shader("raygeneration")]` doesn't yet lower to either DXIL libraries or SPIR-V on that path.

## Test plan

Local on an NVIDIA RTX 3060:
- [ ] Linux Vulkan (native `offloader`)
- [x] Linux D3D12 (Wine + vkd3d-proton + cross-compiled `offloader.exe`)

CI (RT-capable runners):

### DIFF
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -452,15 +452,69 @@ public:
   ComPtr<ID3D12PipelineState> PSO;
   // Only set for graphics pipelines.
   std::optional<D3D_PRIMITIVE_TOPOLOGY> Topology;
+  // True for pipelines created via createPipelineRT — used by SBT / dispatch
+  // code to safely downcast to DXRayTracingPipelineState (parallel to
+  // VulkanPipelineState::IsRayTracing).
+  bool IsRayTracing = false;
 
   DXPipelineState(llvm::StringRef Name, ComPtr<ID3D12RootSignature> RootSig,
                   ComPtr<ID3D12PipelineState> PSO,
-                  std::optional<D3D_PRIMITIVE_TOPOLOGY> Topology)
+                  std::optional<D3D_PRIMITIVE_TOPOLOGY> Topology,
+                  bool IsRT = false)
       : offloadtest::PipelineState(GPUAPI::DirectX), Name(Name),
-        RootSig(RootSig), PSO(PSO), Topology(Topology) {}
+        RootSig(RootSig), PSO(PSO), Topology(Topology), IsRayTracing(IsRT) {}
 
   static bool classof(const offloadtest::PipelineState *B) {
     return B->getAPI() == GPUAPI::DirectX;
+  }
+};
+
+/// RT pipeline state: holds the ID3D12StateObject + cached
+/// ID3D12StateObjectProperties for SBT identifier queries plus a
+/// shader-name → identifier-pointer map. The `void *` identifiers are
+/// owned by Properties — keep it alive for the SBT's lifetime.
+class DXRayTracingPipelineState : public DXPipelineState {
+public:
+  ComPtr<ID3D12StateObject> StateObject;
+  ComPtr<ID3D12StateObjectProperties> Properties;
+  llvm::StringMap<const void *> ShaderIdentifiers;
+
+  DXRayTracingPipelineState(llvm::StringRef Name,
+                            ComPtr<ID3D12RootSignature> RootSig,
+                            ComPtr<ID3D12StateObject> SO,
+                            ComPtr<ID3D12StateObjectProperties> Props)
+      : DXPipelineState(Name, RootSig, /*PSO=*/nullptr, std::nullopt,
+                        /*IsRT=*/true),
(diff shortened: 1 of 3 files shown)


lib/API/DX/Device.cpp
```diff
@@ -1538,18 +1594,232 @@ class DXDevice : public offloadtest::Device {
     return std::make_unique<DXPipelineState>(Name, RootSig, PSO, std::nullopt);
   }
 
+  static std::wstring widen(llvm::StringRef S) {
+    // Entry-point names and hit-group names are ASCII; a straight 1:1 widen
+    // is sufficient.
+    return std::wstring(S.begin(), S.end());
+  }
```
manon-traverse, line 1601: This can be a free function.

manon-traverse: Perhaps we fix this in a follow-up PR where we also refactor out all the `std::wstring` calls to this  helper.

manon-traverse (approved):
Just needs to be clang-tidied and pass CI, other than that this looks fine to me.

EmilioLaiso (approved):

PR #1279, opened 2026-06-03
Cover PSO-only RT surface: recursion, RAY_FLAG_SKIP_CLOSEST_HIT_SHADER, callables

MarijnS95:
Three tests stacked on #1275 covering features that inline RT (`RayQuery` in compute) physically can't express — they're only reachable through a `DispatchRays`-driven RT pipeline.

- `trace-ray-recursion.test` — closest-hit fires a secondary `TraceRay` from an above-the-triangle origin. First-level CH sees payload=0 → bumps to 0x1 → calls `TraceRay`. Second-level CH sees payload!=0 → writes 0x10. Unwinds: first-level OR's in 0x100. Final payload 0x110 (272 decimal). `RayTracingPipelineConfig.MaxTraceRecursionDepth: 2` so both TraceRay calls are within budget.
- `ray-flag-skip-closest-hit.test` — two lanes fire identical rays at the same triangle. Lane 0 uses `RAY_FLAG_NONE` so CH runs and writes 0xBEEF. Lane 1 uses `RAY_FLAG_SKIP_CLOSEST_HIT_SHADER` (PSO-only — inline RT has no equivalent) so CH is skipped and payload keeps its initial 0xAAAA. Output `[0xBEEF, 0xAAAA]`.
- `callable-shader.test` — two callable shaders writing distinct sentinels (0xAAAA / 0xBBBB). Each lane calls `CallShader(Idx, ...)` so the SBT callable region's per-record routing is exercised independently of the hit-group / miss routing already covered in #1277. Callable shaders themselves don't exist in inline RT.

  Only `# XFAIL: Clang` — the clang HLSL path doesn't implement the RT pipeline yet (#1268). On Vulkan the current DXC `-spirv` output is valid: each callable's `IncomingCallableDataKHR` is listed only in its own entry point's interface and `spirv-val` is clean, so the module loads and the test passes (confirmed by an XPASS on the windows-intel VK CI runner). An earlier DXC over-listed those interface variables in every callable entry point, which has since been fixed upstream — so no Vulkan XFAIL or `spirv-opt` cleanup is needed anymore.

All three pass on Metal once the Metal bring-up (#1281) sets `setMaxCallStackDepth(MaxTraceRecursionDepth)` so nested `TraceRay` actually unwinds; with the default of 1 the second TraceRay was silently dropped and the recursion test produced 0x1 instead of 0x110. With #1281 rebased underneath, `Metal` is dropped from the XFAIL list on all three tests.

## Test plan

Local on an NVIDIA RTX 3060:
- [x] Linux D3D12 (Wine + vkd3d-proton + cross-compiled `offloader.exe`)

CI (RT-capable runners):
- [x] windows-intel VK (`VK_KHR_ray_tracing_pipeline`) — all three RT tests pass; callable previously surfaced as an XPASS against the stale Vulkan XFAIL that this revision removes

### DIFF
--- a/test/Feature/RT/callable-shader.test
+++ b/test/Feature/RT/callable-shader.test
@@ -1,0 +1,87 @@
+#--- source.hlsl
+
+// Callable shaders receive an inout user-defined parameter — Vulkan / DXR
+// don't have a Built-in convention beyond "size matches what the caller
+// declared". One uint of payload is enough to confirm routing.
+struct CallableParams {
+  uint Value;
+};
+
+[[vk::binding(0, 0)]] RWStructuredBuffer<uint> Output : register(u0);
+
+[shader("raygeneration")]
+void RayGen() {
+  // Each lane calls a different callable shader index. CallShader picks
+  // the callable record at the given index in the SBT's Callable region
+  // — inline RT has no equivalent of this stage.
+  const uint Idx = DispatchRaysIndex().x;
+  CallableParams P;
+  P.Value = 0;
+  CallShader(Idx, P);
+  Output[Idx] = P.Value;
+}
+
+[shader("callable")]
+void CallableA(inout CallableParams P) {
+  P.Value = 0xAAAA;
+}
+
+[shader("callable")]
+void CallableB(inout CallableParams P) {
+  P.Value = 0xBBBB;
+}
+//--- pipeline.yaml
+---
+Shaders:
+  - Stage: RayGeneration
+    Entry: RayGen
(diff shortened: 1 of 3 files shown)


EmilioLaiso (approved):

MarijnS95:
[Only lavapipe fails](https://github.com/llvm/offload-test-suite/actions/runs/30822528462/job/91738625920?pr=1279) specifically on all RT tests and existing InlineRT tests.

PR #1282, opened 2026-06-03
[DX12] Apply automatic resource transitions on Buffer and Texture objects in the CommandEncoders

manon-traverse:
If we want to do readbacks via the generic buffer and texture objects using the command encoders we will need to do state transitions. Currently all transitions are being done manually but that is very error prone. By introducing this basic state tracking system we always put resources in a preferred state which allows us to bind them as SRV or UAV, and only transition them into other states when we need to do things like copying or executing a render pass.

### DIFF
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -360,11 +360,13 @@ public:
   std::string Name;
   BufferCreateDesc Desc;
   size_t SizeInBytes;
+  D3D12_RESOURCE_STATES PreferredState;
 
   DXBuffer(ComPtr<ID3D12Resource> Buffer, llvm::StringRef Name,
-           BufferCreateDesc Desc, size_t SizeInBytes)
+           BufferCreateDesc Desc, size_t SizeInBytes,
+           D3D12_RESOURCE_STATES PreferredState)
       : offloadtest::Buffer(GPUAPI::DirectX), Buffer(Buffer), Name(Name),
-        Desc(Desc), SizeInBytes(SizeInBytes) {}
+        Desc(Desc), SizeInBytes(SizeInBytes), PreferredState(PreferredState) {}
 
   size_t getSizeInBytes() const override { return SizeInBytes; }
 
@@ -389,6 +391,7 @@ public:
 class DXTexture : public offloadtest::Texture {
 public:
   ComPtr<ID3D12Resource> Resource;
+  D3D12_RESOURCE_STATES PreferredState;
 
   // TODO:
   // Ideally SRVs/UAVs would also live here, but they currently require a
@@ -404,9 +407,9 @@ public:
   TextureCreateDesc Desc;
 
   DXTexture(ComPtr<ID3D12Resource> Resource, llvm::StringRef Name,
-            TextureCreateDesc Desc)
-      : offloadtest::Texture(GPUAPI::DirectX), Resource(Resource), Name(Name),
-        Desc(Desc) {}
+            TextureCreateDesc Desc, D3D12_RESOURCE_STATES PreferredState)
+      : offloadtest::Texture(GPUAPI::DirectX), Resource(Resource),
+        PreferredState(PreferredState), Name(Name), Desc(Desc) {}
 
   const TextureCreateDesc &getDesc() const override { return Desc; }
 
@@ -577,6 +580,7 @@ public:
   ComPtr<ID3D12GraphicsCommandListX> CmdList;
   /// Whether a UAV barrier is pending from a prior compute command.
   bool PendingUAVBarrier = false;
+  llvm::SmallVector<D3D12_RESOURCE_BARRIER> PendingTransitions;
 
   static llvm::Expected<std::unique_ptr<DXCommandBuffer>>
   create(ComPtr<ID3D12DeviceX> Device) {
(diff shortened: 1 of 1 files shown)


lib/API/DX/Device.cpp
```diff
@@ -389,6 +391,7 @@ class DXBuffer : public offloadtest::Buffer {
 class DXTexture : public offloadtest::Texture {
 public:
   ComPtr<ID3D12Resource> Resource;
+  D3D12_RESOURCE_STATES PreferredState;
```
MarijnS95, line 406: So the precedent/contract is for code to keep/leave the texture in this state at all times. We do that in Breda too, but I think it deserves a comment that anyone transitioning away from it is required to transition back too.

lib/API/DX/Device.cpp
```diff
@@ -693,10 +717,28 @@ class DXComputeEncoder : public offloadtest::ComputeEncoder {
...
     auto &DXDst = static_cast<DXBuffer &>(Dst);
-    addUAVBarrier();
+
+    if (DXSrc.PreferredState != D3D12_RESOURCE_STATE_COPY_SOURCE)
+      CB.addResourceTransition(DXSrc.Buffer.Get(), DXSrc.PreferredState,
+                               D3D12_RESOURCE_STATE_COPY_SOURCE);
+    if (DXDst.PreferredState != D3D12_RESOURCE_STATE_COPY_DEST)
+      CB.addResourceTransition(DXDst.Buffer.Get(), DXDst.PreferredState,
+                               D3D12_RESOURCE_STATE_COPY_DEST);
+    CB.flushBarrier();
```
MarijnS95, line 767: What if the preferred state is actually what's listed here? This barrier will be skipped. If there was any operation before or after that also checked against the same `PreferredState`, there's no longer a buffer-specific nor global memory barrier (UAV barrier?) enqueued **between those commands**, possibly allowing those to be overlapping or reordered.

I think layout transitions, access patterns (cache flushes + exec-deps) and pipeline stages (execution dependencies) are different, and exactly why Enhanced Barriers (and what Vulkan shipped with from the get-go 😏) keeps them separate to better represent what's actually happening.

In other words, perhaps we just need to take the state transition from and into the same state? Or have I missed something about state promotion/decay and whatnot scary features that legacy resource transitions have?

lib/API/DX/Device.cpp
```diff
@@ -1406,7 +1485,8 @@ class DXDevice : public offloadtest::Device {
                                "Failed to create texture."))
       return Err;
 
-    auto Tex = std::make_unique<DXTexture>(DeviceTexture, Name, Desc);
+    auto Tex =
+        std::make_unique<DXTexture>(DeviceTexture, Name, Desc, InitialState);
```
MarijnS95: So `InitialState` becomes `PreferredState`?

lib/API/DX/Device.cpp
manon-traverse: Correct that is the contract I am aiming for. Fair to add some comments explaining that!

lib/API/DX/Device.cpp
manon-traverse: Yeah.... I don't know what goes on in this case. I much prefer the enhanced barriers but I was thinking that would become another refactor on it's own since all the existing code relies on legacy barriers.
I suppose we could UAV barrier just in case, but we are not actually writing to a UAV in this case.

lib/API/DX/Device.cpp
manon-traverse: Correct, I'll clarify this in the code.

lib/API/DX/Device.cpp
manon-traverse: I have added a comment explaining the edge case, but I am saving the fix for when we switch to enhanced barriers. That is a change we should make anyway and will also fix this issue.

lib/API/DX/Device.cpp
```diff
@@ -710,10 +746,39 @@ class DXComputeEncoder : public offloadtest::ComputeEncoder {
...
+    // - The Dst Buffer having a PreferredState of
+    // D3D12_RESOURCE_STATE_COPY_DEST
+    // - Each Src Buffer having a PreferredState of
+    // D3D12_RESOURCE_STATE_COPY_SOURCE
+    // In that case no barrier would be emitted
+    // and a race condition would occur. There are ways to solve this with
+    // legacy barriers, but switching to enhanced barriers is a better solution
+    // to this problem.
+
+    if (DXSrc.PreferredState != D3D12_RESOURCE_STATE_COPY_SOURCE)
```
MarijnS95, line 761: After-thought: maybe `addResourceTransition()` should skip if `src==dst`? But that's technically slightly different from merely checking the preferred state.

(But validation could also catch that as "unexpected", because we don't really want to hit this state)

MarijnS95 (approved):

lib/API/DX/Device.cpp
MarijnS95: Yeah, I'm mostly not entirely sure how this case is supposed to be handled with legacy resource states, if it can or is expected to occur at all.

lib/API/DX/Device.cpp
manon-traverse: Hmmmm good point, but we can rely on the debug layer to catch src==dst. So I'll keep this as-is.

bogner (approved):

lib/API/DX/Device.cpp
manon-traverse: The only way to handle this is to transition to COMMON and back, which is... far from ideal. This case won't happen in practice at all since you would need to do multiple memcpys of upload buffers into one readback buffer. That's a case we are not going to hit at all.

PR #1283, opened 2026-06-03
[Utils] Add explicit encoding option to subprocess.run() for github API responses that are non-ASCII

bob80905:
This PR adds an explicit encoding option to subprocess.run() to process non-ASCII characters in Github API responses.
Improves reliability of the script.


Icohedron:
LGTUS @alsepkow @inbelic

alsepkow (approved):

PR #1284, opened 2026-06-04
Add quad domain tessellation test

EmilioLaiso:
Tessellate a viewport-filling quad with the quad domain and integer partitioning, then read back a 2x2 render target whose four texels must carry the four quadrant colours.


manon-traverse (approved):

bogner (approved):
Obligatory clang-format comment, but otherwise LGTM.

PR #1285, opened 2026-06-04
Add isoline domain tessellation test

EmilioLaiso:
Adds a graphics test covering the **isoline** tessellation domain — tessellate one isoline into four segments and capture each generated vertex's along-line coordinate (`SV_DomainLocation.x`) from the domain shader into a UAV. 
The five evenly spaced values `[0, 0.25, 0.5, 0.75, 1]` prove the isoline domain ran and subdivided the line as expected.

The two isoline `SV_TessFactor` slots are ordered as so:
- `Edges[0]` = number of lines, 
- `Edges[1]` = segments per line 

on real D3D12 hardware and on Vulkan, which agree. 
WARP orders them the opposite way (a bug?), so it tessellates this patch into four single-segment lines and produces `[0, 0, 0, 0, 1]` instead.

### DIFF
--- a/test/Graphics/IsolineDomainTessellation.test
+++ b/test/Graphics/IsolineDomainTessellation.test
@@ -59,18 +59,26 @@ struct DSOutput {
   float4 position : SV_POSITION;
 };
 
-// Record each generated vertex's along-line coordinate (SV_DomainLocation.x)
-// into a UAV slot keyed by its segment. Writing is idempotent (overwrite, not
-// add), so the result does not depend on how many times the domain shader runs
-// -- the per-vertex invocation count is implementation-defined. With four
-// segments the vertices land at u = {0, 0.25, 0.5, 0.75, 1}.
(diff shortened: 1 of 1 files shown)


manon-traverse (approved):

test/Graphics/IsolineDomainTessellation.test
```diff
@@ -0,0 +1,167 @@
...
+...
+#--- end
+
+# UNSUPPORTED: Metal
+
+# Clang's HLSL frontend does not support tessellation. 
+# WARP orders the isoline tessellation factors opposite to real 
+# D3D12 hardware and Vulkan, so it tessellates this into one-segment 
+# lines and fails the check.
+# XFAIL: Clang || WARP
```
bogner: Please use two XFAILs here, since these are different reasons. Ideally we would also link to an issue for each as well.

bogner (approved):
Clang-format and XFAIL hygiene, then LGTM.

Issue #1258, opened 2026-06-02
[EPIC]: Improve inline raytracing test coverage


MarijnS95:
This issue tracks expanding the inline raytracing (`RayQuery`) test suite. The bring-up PRs introduce a baseline of three scenarios under `test/Feature/InlineRT/` — a single-triangle hit, an indexed triangle hit, and instance selection via translated transforms. The list below walks the HLSL inline-RT spec surface — limited to behavior observable from a shader, since that is what this suite tests — and notes what becomes testable after those PRs land versus what still needs further framework / YAML work.

### Bring-up PRs (foundation for everything below)
- [x] YAML-driven acceleration structure definitions in `.test` files
    - https://github.com/llvm/offload-test-suite/pull/1214
- [ ] RT acceleration structure abstraction (size queries, resource allocation) 👀
    - https://github.com/llvm/offload-test-suite/pull/1232
- [ ] Bind acceleration structures and enable the InlineRT tests 👀
    - https://github.com/llvm/offload-test-suite/pull/1245

Legend:
- [ ] Not started
- [ ] In progress 🏗️
- [ ] In review 👀
- [x] Finished (merged in main)
- 👍 testable on top of the bring-up PRs above
- 🏗️ requires additional backend or YAML work first

### `TraceRayInline` parameters
- [ ] `RAY_FLAG_NONE`, default mask `0xFF` 👀
    - covered by the bring-up tests in #1245
- [ ] Per-call `RayFlags` argument != template flags (runtime-OR with template) — 👍
- [ ] `InstanceInclusionMask` filtering — TLAS instances with various `InstanceMask` values, query with overlapping/disjoint masks 👀
    - covered by `instance-mask.test` in #1274
- [ ] `RayDesc` with non-axis-aligned direction (no auto-normalization) — 👍
- [ ] `RayDesc` with `TMin > 0` excluding a near hit 👀
    - covered by `tmin-tmax-clip.test` in #1271
- [ ] `RayDesc` with `TMax` clipping a far hit 👀
    - covered by `tmin-tmax-clip.test` in #1271
- [ ] Multiple `TraceRayInline` calls on the same `RayQuery` object — 👍
- [ ] Multiple distinct `RayQuery<Flags>` objects in the same shader — 👍

### Template `<RayFlags>` (compile-time)
- [ ] `RAY_FLAG_FORCE_OPAQUE` (skips any-hit / non-opaque candidate path) — 👍
[...]

PR #1289, opened 2026-06-04
Wire AABB/procedural geometry through the Pipeline-level AS builder

MarijnS95:
Depends on #1245

## Summary
Routes `BLASDesc.AABBs` through the Pipeline → API conversion in `buildPipelineAccelerationStructures`, picking the appropriate `BLASBuildRequest` variant (`SmallVector<AABBGeometryDesc>` or `SmallVector<TriangleGeometryDesc>` in `Geometry`) based on which geometry list is populated (mixed and empty cases now error explicitly). The DX/VK/MTL AABB build paths were already in place from the bring-up; this PR just hooks them up to YAML.

The covering test (`Feature/InlineRT/aabb-procedural.test`) defines a single AABB, fires a ray through it, and in the candidate loop calls `CommitProceduralPrimitiveHit()`; `CommittedStatus()` must return `COMMITTED_PROCEDURAL_PRIMITIVE_HIT`.

Part of the inline-RT test coverage epic (https://github.com/llvm/offload-test-suite/issues/1258).

## Test plan

Local on an NVIDIA RTX 3060:

CI (RT-capable runners):
- [x] windows-nvidia D3D12 (`RaytracingTier 1.2`)

### DIFF
--- a/lib/API/Device.cpp
+++ b/lib/API/Device.cpp
@@ -166,51 +166,90 @@ llvm::Error offloadtest::buildPipelineAccelerationStructures(
   llvm::StringMap<AccelerationStructure *> BLASesByName;
 
   for (const auto &BD : P.AccelStructs.BLAS) {
-    llvm::SmallVector<TriangleGeometryDesc> Triangles;
-    Triangles.reserve(BD.Triangles.size());
-    for (const auto &T : BD.Triangles) {
-      assert(T.VertexBufferPtr && "VertexBufferPtr not resolved");
-      auto VBOrErr = createBufferWithData(
-          Dev, "AS-Vertices", UploadDesc, T.VertexBufferPtr->Data[0].get(),
-          T.VertexBufferPtr->size(), nullptr, nullptr);
-      if (!VBOrErr)
-        return VBOrErr.takeError();
+    if (!BD.Triangles.empty() && !BD.AABBs.empty())
+      return llvm::createStringError(std::errc::invalid_argument,
+                                     "BLAS '%s' mixes triangle and AABB "
+                                     "geometry; must be either, not both.",
+                                     BD.Name.c_str());
+    if (BD.Triangles.empty() && BD.AABBs.empty())
+      return llvm::createStringError(std::errc::invalid_argument,
+                                     "BLAS '%s' has no geometry.",
+                                     BD.Name.c_str());
 
-      TriangleGeometryDesc TGD;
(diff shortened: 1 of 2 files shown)


lib/API/Device.cpp
```diff
@@ -118,50 +118,89 @@ llvm::Error offloadtest::buildPipelineAccelerationStructures(
...
-        if (!IBOrErr)
-          return IBOrErr.takeError();
-        TGD.IndexBuffer = IBOrErr->get();
-        TGD.IndexCount = T.IndexCount;
-        TGD.IdxFormat = T.IdxFormat;
-        OutInputBuffers.push_back(std::move(*IBOrErr));
+    if (!BD.Triangles.empty() && !BD.AABBs.empty())
+      return llvm::createStringError(
+          std::errc::invalid_argument,
+          "BLAS '%s' mixes triangle and AABB geometry; pick one.",
```
manon-traverse:
```suggestion
          "BLAS '%s' mixes triangle and AABB geometry; must be either not both.",
```

test/Feature/InlineRT/aabb-procedural.test
```diff
@@ -0,0 +1,86 @@
...
+  Ray.Direction = float3(0, 0, -1);
+  Ray.TMin = 0.0;
+  Ray.TMax = 100.0;
+  RayQuery<RAY_FLAG_NONE> Q;
+  Q.TraceRayInline(Scene, RAY_FLAG_NONE, 0xFF, Ray);
+  while (Q.Proceed()) {
+    if (Q.CandidateType() == CANDIDATE_PROCEDURAL_PRIMITIVE)
+      Q.CommitProceduralPrimitiveHit(1.5);
+  }
+  Output[0] = (uint)Q.CommittedStatus();
```
manon-traverse: I am not quite sure about this one.
The RT API doesn't require the ray to actually hit the AABB to register it as a hit (due to AABB quantization).
I suppose for this specific example it's fine since we only need a ray to hit the box, but it might be better if we have multiple rays where some hit and some miss so that the AABB data is uploaded as expected.
This means we will also need to do a manual AABB test after we register an intersection.

bogner (approved):

bogner:
I've put up #1395 to resolve the conflicts and get this patch merged.

MarijnS95:
@bogner thanks but that's really odd; I thought I rebased and pushed all my branches after you mentioned it in Discord - trying again.

PR #1296, opened 2026-06-08
[Vulkan] Fix xpass cases

farzonl:
Suprisingly the spirv-val error impacting intel is not impacting QC,NV, or AMD. Also looks like QC loop peeling is fixed for dxc.


V-FEXrt (approved):

tex3d:
Did you mean to say `XFAIL` in the title instead of `xpass`? `[Vulkan] Fix xpass cases #1296`

farzonl:
> Did you mean to say `XFAIL` in the title instead of `xpass`? `[Vulkan] Fix xpass cases #1296`

No the ones I am fixing are unexpectedly passing the lit suite  prints these as `XPASS` when I look at the logs.

PR #1299, opened 2026-06-09
[VK] Apply automatic layout transitions on Texture objects in the CommandEncoders

manon-traverse:
This is the Vulkan equivalent of #1282 

Since Vulkan barriers allow us to be more specific we only need to keep track of image layouts. However, Vulkan doesn't allow specifying an initial layout other than `UNDEFINED` or `PREINITIALIZED`, meaning we need to add some extra logic there to handle `UNDEFINED` layout cases.

Currently only the render target and depth stencil buffer use this system but in the future all vulkan textures will use this system with minimal tracking and layout transitions.

### DIFF
--- a/lib/API/VK/Device.cpp
+++ b/lib/API/VK/Device.cpp
@@ -486,16 +486,27 @@ public:
   std::string Name;
   TextureCreateDesc Desc;
 
+  VkImageLayout PreferredLayout = VK_IMAGE_LAYOUT_GENERAL;
+  VkImageSubresourceRange FullRange;
+  bool IsInUndefinedLayout = true;
+
   VulkanTexture(VkDevice Dev, VkImage Image, VkDeviceMemory Memory,
-                llvm::StringRef Name, TextureCreateDesc Desc)
+                llvm::StringRef Name, TextureCreateDesc Desc,
+                VkImageLayout PreferredLayout,
+                VkImageSubresourceRange FullRange)
       : offloadtest::Texture(GPUAPI::Vulkan), Dev(Dev), Image(Image),
-        Memory(Memory), Name(Name), Desc(Desc) {}
+        Memory(Memory), Name(Name), Desc(Desc),
+        PreferredLayout(PreferredLayout), FullRange(FullRange) {}
 
   ~VulkanTexture() override {
     if (View)
       vkDestroyImageView(Dev, View, nullptr);
     vkDestroyImage(Dev, Image, nullptr);
     vkFreeMemory(Dev, Memory, nullptr);
+  }
+
+  VkImageLayout preferredLayoutOrUndefined() {
+    return IsInUndefinedLayout ? VK_IMAGE_LAYOUT_UNDEFINED : PreferredLayout;
   }
 
   const TextureCreateDesc &getDesc() const override { return Desc; }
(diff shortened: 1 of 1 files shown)


lib/API/VK/Device.cpp
```diff
@@ -691,28 +702,44 @@ class VulkanCommandBuffer : public offloadtest::CommandBuffer {
...
-      PendingDstAccess = 0;
-      return;
+    if (PendingSrcStage != 0 || !PendingImageTransitions.empty()) {
+      VkMemoryBarrier Barrier = {};
+      Barrier.sType = VK_STRUCTURE_TYPE_MEMORY_BARRIER;
+      Barrier.srcAccessMask = PendingSrcAccess;
+      Barrier.dstAccessMask = PendingDstAccess;
+      vkCmdPipelineBarrier(CmdBuffer, PendingSrcStage, PendingDstStage, 0, 1,
+                           &Barrier, 0, nullptr, PendingImageTransitions.size(),
+                           PendingImageTransitions.data());
```
MarijnS95: Nit: I've seen some drivers not being happy with a non-null ptr even if size==0... (blegh)

lib/API/VK/Device.cpp
```diff
@@ -1094,6 +1146,38 @@ VulkanCommandBuffer::createRenderEncoder(
     ClearValues.push_back(CV);
   }
 
+  for (size_t I = 0; I < Desc.ColorAttachments.size(); ++I) {
+    auto &Tex = llvm::cast<VulkanTexture>(*Desc.ColorAttachments[I]);
+    if (PassDesc.ColorAttachments[I].Load == LoadAction::Load) {
+      this->addImageTransition(
```
MarijnS95, line 1159: Can we have a "TODO these belong in the render pass setup"?

lib/API/VK/Device.cpp
```diff
@@ -1094,6 +1146,38 @@ VulkanCommandBuffer::createRenderEncoder(
...
+
+    if (PassDesc.DepthStencil->DepthLoad == LoadAction::Load ||
+        PassDesc.DepthStencil->StencilLoad == LoadAction::Load) {
+      this->addImageTransition(
+          this->PendingSrcAccess, /*SrcAccessMask*/
+          VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT |
+              VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT, /*DstAccessMask*/
+          Tex.preferredLayoutOrUndefined(),                 /*OldLayout*/
+          VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL, /*NewLayout*/
+          Tex.Image, Tex.FullRange);
```
MarijnS95: Crazy clang-format for not aligning _all_ parameter comments 🤮

lib/API/VK/Device.cpp
```diff
@@ -498,6 +505,10 @@ class VulkanTexture : public offloadtest::Texture {
     vkFreeMemory(Dev, Memory, nullptr);
   }
 
+  VkImageLayout preferredLayoutOrUndefined() {
+    return IsInUndefinedLayout ? VK_IMAGE_LAYOUT_UNDEFINED : PreferredLayout;
```
MarijnS95, line 509: In breda we immediately set undefined=false here, assuming that that the caller is going to transition away at this point... (requires some function docs and maybe a rename)

lib/API/VK/Device.cpp
```diff
@@ -486,10 +486,17 @@ class VulkanTexture : public offloadtest::Texture {
   std::string Name;
   TextureCreateDesc Desc;
 
+  VkImageLayout PreferredLayout = VK_IMAGE_LAYOUT_GENERAL;
+  VkImageSubresourceRange FullRange;
+  bool IsInUndefinedLayout = true;
```
MarijnS95, line 491: ~No compute shaders use textures yet? Otherwise I'm missing where they transition away from UNDEFINED.~

EDIT: Right they're not used interchangeably and currently open-code the transitions manually?

MarijnS95 (commented):
My beloved Breda pattern! Just curious how this works for compute?

EDIT: Ah, it's a "system" and compute still uses a manual transition path/call? Then I understand the rationale.

lib/API/VK/Device.cpp
manon-traverse: This is the render pass setup right?

MarijnS95: Yes. Render passes natively support transitions when setting up attachments, but here plain pipeline barriers are used.

lib/API/VK/Device.cpp
manon-traverse: Yes, I am not sure what the best approach here is. Setting the layout to defined by calling `undefinedOrPreferredLayout` sounds like a bad idea since that names screams 'getter' to me, and I don't like it when getters change state.

Perhaps we should pass the actual `Texture` object to the `addImageTransition` funciton and let that set the `Texture` to defined state? 🤔 

Let me play around with that idea!

lib/API/VK/Device.cpp
manon-traverse: clang-format is gonna clang-format ¯\\\_(ツ)\_/¯

lib/API/VK/Device.cpp
manon-traverse: Does that path provide any benefits over the way we do it now? Right now we get the flexibility and control over the transition into the COLOR_ATTACHMENT_OPTIMAL layout no matter what the previous layout was. If we go your route we cannot support textures that are `Storage | RenderTarget` since those have a different `PreferredLayout`.

lib/API/VK/Device.cpp
```diff
@@ -691,28 +702,47 @@ class VulkanCommandBuffer : public offloadtest::CommandBuffer {
...
+        DstAccessMask,
+        OldLayout,
+        NewLayout,
+        0, /*srcQueueFamilyIndex*/
+        0, /*dstQueueFamilyIndex*/
+        Texture.Image,
+        Texture.FullRange,
+    });
+
+    Texture.IsInUndefinedLayout = false;
```
EmilioLaiso, line 727: What would happen if `NewLayout` is undefined? Why would I do that? Good question 😆 But I guess an assert would make the API safer to use and would prevent a wrong value here.

manon-traverse: Good point, added an assert.

There are no good reasons to ever transition to an `UNDEFINED` layout.
However, transitioning from an `UNDEFINED` layout is valid, but does discard the contents of the image.

EmilioLaiso (approved):

Issue #1300, opened 2026-06-09
Track XFails across the repo effectively


bob80905:
We need a way to more efficiently track XFAILS across the test suite.
It is troublesome to discover which XFAILS need to be added, and which ones need to be removed.
A script to quickly collect and categorize test failures that need XFAILS, or tests with XFAILS that need to be removed, would assist in this organization effort.

PR #1301, opened 2026-06-09
[Utils] Add xfail sync script

bob80905:
This PR implements a way to track the XFAILs that have been added to the offload test suite.
There are 2 failure cases this script is able to catch.
1. A test that has consistent failures is missing an XFAIL annotation, and should have an XFAIL annotation added.
2. A test that already has an XFAIL annotation is consistently passing, and should have its XFAIL annotation removed.

Here is the output of the script when run:
image

Addresses https://github.com/llvm/offload-test-suite/issues/1300
Assisted by: Github Copilot

### DIFF
--- a/utils/xfail_sync.py
+++ b/utils/xfail_sync.py
@@ -1,0 +1,762 @@
+#!/usr/bin/env python3
+"""Synchronize XFAIL/UNSUPPORTED annotations with actual CI results.
+
+Compares the XFAIL and UNSUPPORTED annotations in .test/.yaml files against
+recent GitHub Actions CI results to identify:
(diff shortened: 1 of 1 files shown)


bogner:
I'm not really convinced this is useful. Most of the time the first time something fails or xpasses it's pretty obvious what we should do with it, so I expect the frequency with which this gives us any extra information to be low.

bogner:
I've put up https://github.com/llvm/offload-test-suite/pull/1303 which I think gives a better overview of what's currently happening and might be sufficient instead of this.

bob80905:
Talked with Justin offline, I'll close this in favor of
https://github.com/llvm/offload-test-suite/pull/1303

PR #1302, opened 2026-06-10
Add support for 64-bit formats.

manon-traverse:
DX12, Vulkan and Metal do support 64-bit formats for a couple of use cases like Storage(UAV) Textures.
This PR adds the 64-bit formats that are supported to the list of available formats and adds the necessary code to convert to API specific formats.
Note that 64-bit formats are not always supported for vertex inputs or sampled textures.

### DIFF
--- a/include/API/FormatConversion.h
+++ b/include/API/FormatConversion.h
@@ -88,6 +88,7 @@ inline llvm::Expected<Format> toFormat(DataFormat Format, int Channels) {
       return Format::D32Float;
     break;
   case DataFormat::UInt64:
+    // Only 1 and 2 channels of 64-bit integers are supported.
     switch (Channels) {
     case 1:
       return Format::R64Uint;
@@ -96,6 +97,7 @@ inline llvm::Expected<Format> toFormat(DataFormat Format, int Channels) {
     }
     break;
   case DataFormat::Int64:
+    // Only 1 and 2 channels of 64-bit integers are supported.
     switch (Channels) {
     case 1:
       return Format::R64Sint;
--- a/include/API/FormatConversion.h
+++ b/include/API/FormatConversion.h
@@ -87,13 +87,27 @@ inline llvm::Expected<Format> toFormat(DataFormat Format, int Channels) {
     if (Channels == 1)
       return Format::D32Float;
     break;
+  case DataFormat::UInt64:
+    switch (Channels) {
+    case 1:
+      return Format::R64Uint;
+    case 2:
+      return Format::RG64Uint;
+    }
+    break;
+  case DataFormat::Int64:
+    switch (Channels) {
+    case 1:
+      return Format::R64Sint;
+    case 2:
+      return Format::RG64Sint;
+    }
+    break;
   // No Format mapping for these DataFormats.
   case DataFormat::Hex8:
   case DataFormat::Hex16:
   case DataFormat::Hex32:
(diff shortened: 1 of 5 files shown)


include/API/Resources.h
```diff
@@ -199,6 +223,11 @@ inline bool isTextureCompatible(Format Format) {
   case Format::RGBA32Float:
   case Format::D32Float:
   case Format::D32FloatS8Uint:
+  // Only for RWTextures
+  case Format::R64Uint:
+  case Format::R64Sint:
+  case Format::RG64Uint:
+  case Format::RG64Sint:
```
EmilioLaiso, line 230: Then should we have `isRwTextureCompatible` ?

EmilioLaiso (commented):
Is there a plan to add a test exercising these new formats?

manon-traverse:
> Is there a plan to add a test exercising these new formats?

There are already tests that use these formats. However, the resources for those tests still use raw dx12, vulkan, and metal objects, not the generic types. In a future PR we will be creating these resources with the generic types and thus need support for these 64-bit formats. Since that change is huge I am splitting it up in smaller steps that can be reviewed individually so it's easier to see what is going on.

include/API/Resources.h
manon-traverse: I don't think so. Like you can still have an SRV for these and load it as a regular texture, but you can't use a sampler on it from what I understand. It's one of these edge cases where there is no clear yes/no answer on 'can we create a texture with this format'.

MarijnS95 (approved):

lib/API/DX/DXResources.h
```diff
@@ -65,6 +65,14 @@ inline DXGI_FORMAT getDXGIFormat(Format Format) {
     return DXGI_FORMAT_R32G32B32A32_UINT;
   case Format::RGBA32Float:
     return DXGI_FORMAT_R32G32B32A32_FLOAT;
+  case Format::R64Uint:
```
bob80905, line 68: This block confuses me, though I am not sure of all the options available to choose from.
Is it possible to add a format that more readily fits these cases, like in VKResources.h?
e.g., DXGI_FORMAT_R64_UINT would be perfect, if it exists, right?

include/API/FormatConversion.h
```diff
@@ -87,13 +87,27 @@ inline llvm::Expected<Format> toFormat(DataFormat Format, int Channels) {
     if (Channels == 1)
       return Format::D32Float;
     break;
+  case DataFormat::UInt64:
+    switch (Channels) {
+    case 1:
+      return Format::R64Uint;
+    case 2:
```
bob80905, line 95: Also, is there a reason we stop at 2 channels?

manon-traverse: Because the hardware and APIs don't support more than 2 channels for 64-bit types.

lib/API/DX/DXResources.h
manon-traverse: `DXGI_FORMAT_R64_UINT ` does not exist. The API expects you to use `DXGI_FORMAT_R32G32_UINT` instead. When you specify the `uint64_t` in the shader it will load it as a `uint64_t` though.

Vulkan does distinguish between R32G32 and R64, so we make that distinction in our formats as well.

bob80905 (approved):

Issue #1146, opened 2026-05-04
[EPIC]: Improve rasterization tests coverage


EmilioLaiso:
#### Semantics
- [x] SV_VertexID
    - https://github.com/llvm/offload-test-suite/pull/1142
- [x] SV_Position
    - Already covered by all graphics tests
- [ ] SV_ClipDistance[n]
    - https://github.com/llvm/offload-test-suite/pull/1168
- [ ] SV_CullDistance[n]
    - https://github.com/llvm/offload-test-suite/pull/1168
- [ ] SV_InstanceID - 🏗️
- [ ] SV_StartVertexLocation - 🏗️
- [ ] SV_StartInstanceLocation - 🏗️
- [ ] SV_RenderTargetArrayIndex - 🏗️
- [ ] SV_ViewportArrayIndex - 🏗️
- [ ] SV_ViewID - 🏗️

#### Functional
- [ ] Non-`Float32` vertex attribute formats (`Int32`, `UInt32`, `Float16`, `Int16`, `UInt16`) - 👍
- [ ] Normalized integer attribute formats (`UNorm8`, `SNorm8`, `UNorm16`, `SNorm16`) - 👍
- [x] Multiple varyings to PS — many simultaneous user semantics, mixed integer/float types
    - https://github.com/llvm/offload-test-suite/pull/1255
- [ ] Struct-typed varyings between VS and PS - 👍
- [x] CBuffer read in VS (real `mul(MVP, position)` instead of pass-through)
    - https://github.com/llvm/offload-test-suite/pull/1255
- [ ] Push constants in VS - 👍
- [x] `StructuredBuffer<T>` read from VS
    - https://github.com/llvm/offload-test-suite/pull/1255
- [x] `ByteAddressBuffer` read from VS
    - https://github.com/llvm/offload-test-suite/pull/1255
- [ ] `RWStructuredBuffer` write from VS (different UAV path than PS-side write) - 👍
- [ ] `Texture2D.SampleLevel(...)` from VS (no derivatives in VS) - 👍
- [ ] `Texture2D.Load(...)` from VS - 👍
- [x] `SV_Position` with W ≠ 1 (verify perspective divide downstream)
    - https://github.com/llvm/offload-test-suite/pull/1255
- [ ] Multiple vertex buffer slots (stream 0 + stream 1) - 🏗️
- [ ] Per-instance attributes (input rate `INSTANCE`) - 🏗️
[...]

PR #1304, opened 2026-06-11
Hull and Domain shader system-value tests

EmilioLaiso:
Both tests validate the tessellation system values directly (capturing them into a `RWStructuredBuffer` and checking with `Results: BufferExact`), rather than inferring them from a rendered image.

### `HullSystemValues.test`
Validates:
- `SV_PrimitiveID` in both the patch-constant function and HS `main`.
- `SV_OutputControlPointID`. Each invocation sees its own index `{0,1,2}`.
- `SV_TessFactor` / `SV_InsideTessFactor`. Round-tripped HS -> DS and checked.

### `DomainSystemValues.test`
Validates:
- `SV_PrimitiveID` as a DS input. Not exercised by any other test.
- `SV_DomainLocation` — exact `(u,v)` per pixel (same mapping proven in `QuadDomainTessellation.test`, captured here as exact floats).
- `SV_Position` (DS output → rasterizer → PS). Exact pixel centers.

Part of #1146

### DIFF
--- a/test/Feature/Semantics/DomainSystemValues.test
+++ b/test/Feature/Semantics/DomainSystemValues.test
@@ -3,11 +3,6 @@
 #   * SV_DomainLocation - DS input (the tessellator's (u, v) for this vertex)
 #   * SV_Position       - DS output -> rasterizer -> PS input (pixel center)
 #
-# SV_PrimitiveID: https://github.com/llvm/wg-hlsl/issues/160
-# SV_DomainLocation / SV_Position (DS): https://github.com/llvm/wg-hlsl/issues/141
-# Clang's HLSL -> DXIL lowering does not yet implement the tessellation stages,
-# so this is XFAIL on Clang. Metal has no Hull/Domain stages (see note below).
-#
 # Geometry: two independent quad patches drawn back-to-back (8 control points,
 # PatchControlPoints = 4). The input-assembler hands each patch a distinct
 # SV_PrimitiveID (0, 1). Each patch covers one half of a 4x2 render target:
@@ -119,7 +114,7 @@ struct PSInput {
 };
 
 struct Record {
-  uint  PrimID;
+  uint PrimID;
   float DomU;
   float DomV;
   float PosX;
@@ -131,17 +126,16 @@ struct Record {
 RWStructuredBuffer<Record> Output : register(u0);
 
 float4 main(PSInput input) : SV_TARGET {
-  // 4x2 target, top-left origin, row-major.
   uint idx = (uint)input.position.y * 4u + (uint)input.position.x;
 
   Record r;
   r.PrimID = input.primID;
-  r.DomU   = input.uv.x;
-  r.DomV   = input.uv.y;
-  r.PosX   = input.position.x;
-  r.PosY   = input.position.y;
-  r.PosZ   = input.position.z;
-  r.PosW   = input.position.w;
+  r.DomU = input.uv.x;
+  r.DomV = input.uv.y;
+  r.PosX = input.position.x;
+  r.PosY = input.position.y;
+  r.PosZ = input.position.z;
+  r.PosW = input.position.w;
   Output[idx] = r;
 
   return float4(1.0, 0.0, 0.0, 1.0);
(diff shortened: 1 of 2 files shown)


test/Feature/Semantics/DomainSystemValues.test
```diff
@@ -0,0 +1,245 @@
...
+      Width: 4
+      Depth: 1
+  - Name: ResultBuffer
+    Format: Hex32
+    Stride: 28 # sizeof(Record): 1 uint + 6 float
+    FillSize: 224 # 8 records * 28 bytes
+    FillValue: 0
+  - Name: ResultBuffer_Expected
+    Format: Hex32
+    Stride: 28
```
manon-traverse, line 187: Why are we using floats values written down as hex values instead of just floating point values?

test/Feature/Semantics/HullSystemValues.test
```diff
@@ -0,0 +1,250 @@
...
+    Stride: 28
+    # Per-pixel expected records:
+    #   PCPrimID, HSMainPrimID, Cpid0, Cpid1, Cpid2, EdgeFactor, InsideFactor
+    # 1.0 -> 0x3F800000
+    Data: [
+      # idx0: patch 0
+      0x0, 0x0, 0x0, 0x1, 0x2, 0x3F800000, 0x3F800000,
+      # idx1: patch 1
+      0x1, 0x1, 0x0, 0x1, 0x2, 0x3F800000, 0x3F800000,
+    ]
```
manon-traverse, line 210: Perhaps we should use two buffers if we are missing floats and integers? What do you think?

manon-traverse (approved):

test/Feature/Semantics/DomainSystemValues.test
```diff
@@ -0,0 +1,245 @@
...
+#   * SV_Position       - DS output -> rasterizer -> PS input (pixel center)
+#
+# Geometry: two independent quad patches drawn back-to-back (8 control points,
+# PatchControlPoints = 4). The input-assembler hands each patch a distinct
+# SV_PrimitiveID (0, 1). Each patch covers one half of a 4x2 render target:
+#
+#     patch 0  -> NDC x in [-1, 0]  -> columns 0,1
+#     patch 1  -> NDC x in [ 0, 1]  -> columns 2,3
+#
+# With integer partitioning and all tess factors = 2, each patch is subdivided
```
bogner: Might be better to spell out "tesselation" rather than "tess" in the comment, for clarity.

test/Feature/Semantics/DomainSystemValues.test
bogner: Presumably it's because the ResultBuffer has a mix of integer and floating point, and the YAML doesn't have a way to do mixed

test/Feature/Semantics/HullSystemValues.test
```diff
@@ -0,0 +1,250 @@
...
+    Rule: BufferExact
+    Actual: ResultBuffer
+    Expected: ResultBuffer_Expected
+...
+#--- end
+
+# Metal has tessellation but no Hull/Domain stages: HS is a compute kernel
+# writing per-patch factors, DS is a post-tessellation vertex function tagged
+# [[patch(...)]]. Mapping HLSL HS/DS onto that isn't wired up, so skip Metal.
+# UNSUPPORTED: Metal
```
bogner, line 241: There's an argument for XFAIL here, since we could in theory wire it up if we decided to spend the effort on that. In any case, UNSUPPORTED is probably fine for hull and domain in particular as I don't really see getting those working in metal to be a priority. Could use a similar comment in the other test.

bogner (approved):

test/Feature/Semantics/DomainSystemValues.test
EmilioLaiso: Indeed! 😃

test/Feature/Semantics/HullSystemValues.test
EmilioLaiso: Left it as UNSUPPORTED for now (and aligned the comments). But indeed it could be an XFAIL candidate.

PR #1305, opened 2026-06-11
Support arrays of TLAS bindings

MarijnS95:
Mirror the existing buffer/texture array pattern (`ArraySize` plus an `ArraySize`-driven YAML field) for top-level acceleration structures so shaders can declare `RaytracingAccelerationStructure Scenes[N]` and bind N distinct TLASes through a single resource entry.

- **Schema** (`Support/Pipeline.{h,cpp}`): `TLASDesc` gains `ArraySize` (default 1) and reshapes `Instances` from `SmallVector<InstanceDesc>` to `SmallVector<SmallVector<InstanceDesc>>`. `MappingTraits<TLASDesc>` dispatches on `ArraySize` exactly like `setData()` does for `CPUBuffer`: flat `Instances: [...]` for the scalar case, list-of-lists for arrays, with an `ActualSize != ArraySize` validation error on mismatch. `Resource::getArraySize()` returns `TLASPtr->ArraySize` for AS resources (moved out-of-line so it can dereference the still-forward-declared `TLASDesc`).
- **createAS()**: takes `uint32_t InstanceCount` directly — pure single-create that sizes and allocates one TLAS. No `Pipeline` / `Resource` / `InvocationState` parameters. The multi-create (`createBuffers` / `createResources`) loops `TD.ArraySize` and pushes one bundle entry per element plus N handles into `InvocationState::TLASes`, which becomes `StringMap<SmallVector<unique_ptr<AccelerationStructure>>>` (one vector per `TLASDesc::Name`, sized to `ArraySize`).
- **`buildPipelineAccelerationStructures()`** walks `P.AccelStructs.TLAS` and, for each name with a pre-allocated vector, builds one `TLASBuildRequest` per element using `TD.Instances[Elt]` and `Handles[Elt]`.
- **Vulkan**: descriptor write iterates the bundle's `ResourceRefs` to fill N entries of `pAccelerationStructures` and sets `descriptorCount = R.getArraySize()` so the descriptor set sees the full array. Pool sizing (`ASDescriptorCount`) and binding layout (`ResourceBinding.DescriptorCount`) already used `R.getArraySize()` for AS resources — they just start returning the right number once `TLASDesc::ArraySize` flows through.
- **DX12**: `bindSRV()`'s AS branch already loops `ResBundle`; with the new multi-entry bundle it now writes N `RAYTRACING_ACCELERATION_STRUCTURE` SRVs into consecutive heap slots automatically. Heap sizing already uses `getDescriptorCountWithFlattenedArrays()`.
- **Metal**: the AS descriptor-binding loop now builds one `IRRaytracingAccelerationStructureGPUHeader` + instance-contributions buffer pair per array element and writes one descriptor entry per element via `IRDescriptorTableSetAccelerationStructure()`. `MarkASResident` descends into the per-name vector.

New test `test/Feature/InlineRT/tlas-array.test` declares `Scenes[2]`, each TLAS carries one triangle instance with a distinct `InstanceID` (10 and 20), and the shader writes each `CommittedInstanceID()` into `Output[i]`. Gated on `acceleration-structure`, `XFAIL: Clang`.

## Test plan

Local on an NVIDIA RTX 3060:

CI (RT-capable runners):
- [ ] macOS Metal (`supportsRaytracing`)

### DIFF
--- a/include/Support/Pipeline.h
+++ b/include/Support/Pipeline.h
@@ -398,11 +398,7 @@ struct Resource {
     return isByteAddressBuffer() ? 4 : BufferPtr->getElementSize();
   }
 
-  uint32_t getArraySize() const {
-    if (isSampler() || isAccelerationStructure())
-      return 1;
-    return BufferPtr->ArraySize;
-  }
+  uint32_t getArraySize() const;
 
   uint32_t size() const {
     assert(!isSampler() && !isAccelerationStructure() &&
@@ -582,7 +578,11 @@ struct InstanceDesc {
 
 struct TLASDesc {
   std::string Name;
-  llvm::SmallVector<InstanceDesc> Instances;
+  uint32_t ArraySize = 1;
+  // Outer vector has ArraySize entries (one per descriptor-array element);
+  // inner vector lists the instances for that element. Mirrors
+  // CPUBuffer::Data's ArraySize-driven layout.
+  llvm::SmallVector<llvm::SmallVector<InstanceDesc>, 1> Instances;
 };
 
 struct AccelerationStructureDescs {
@@ -623,6 +623,14 @@ struct ShaderBindingTableDesc {
   llvm::SmallVector<SBTEntry> HitGroup;
   llvm::SmallVector<SBTEntry> Callable;
 };
+
+inline uint32_t Resource::getArraySize() const {
+  if (isSampler())
+    return 1;
+  if (isAccelerationStructure())
+    return TLASPtr->ArraySize;
+  return BufferPtr->ArraySize;
+}
 
 struct Pipeline {
   ShaderPipelineKind Kind;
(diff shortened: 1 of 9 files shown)


include/Support/Pipeline.h
```diff
@@ -338,11 +338,7 @@ struct Resource {
     return isByteAddressBuffer() ? 4 : BufferPtr->getElementSize();
   }
 
-  uint32_t getArraySize() const {
-    if (isSampler() || isAccelerationStructure())
-      return 1;
-    return BufferPtr->ArraySize;
-  }
+  uint32_t getArraySize() const; // out-of-line: needs complete TLASDesc.
```
bogner: Explaining why this is out of line isn't terribly useful I think. If somebody for some reason really wanted it to be inline they'd figure it out in a hurry.
```suggestion
  uint32_t getArraySize() const;
```

bogner (approved):

PR #1306, opened 2026-06-11
Improve buffer creation API to cover all needed current use cases.

manon-traverse:
Improve the buffer creation API:
* Add a `BufferShaderAccessType` enum to distinguish between buffers that do raw access, typed access, or are just as structured buffers.
* Add `BufferShaderAccessTypeParams` to provide a format or stride in the cases of typed access or a structured buffer
* Add `MemoryBacking` to handle reserved resources
* Add `BufferUsage` variants for constant buffers, index buffers and indirect arguments
* Add `HasCounter` bool for structured buffers with a counter value.

Use the extra API features to:
* Create SRV, UAV, and CBV descriptors in DX12. These are not used anywhere, but are now ready to replace the buffer creation using raw dx12 with the generic graphics interface
* Allocate a counter value if needed to the buffer (DX12 and Vulkan)

Drive by fix:
* Add a helper function for a scratch buffer `BufferCreateDesc`
* Use the existing `BufferCreateDesc` helper functions in the acceleration structure code.

### DIFF
--- a/include/API/Resources.h
+++ b/include/API/Resources.h
@@ -33,7 +33,7 @@ enum class MemoryBacking {
   // DX: CreateReservedResource + UpdateTileMappings.
   // VK: VK_IMAGE_CREATE_SPARSE_BINDING_BIT + vkQueueBindSparse.
   // Metal: MTLTextureDescriptor.sparseLevel + heap tile mapping
-  // (requires Apple Silicon).
+  //        (requires Apple Silicon).
   Sparse,
 };
 
--- a/include/API/Resources.h
+++ b/include/API/Resources.h
@@ -25,6 +25,18 @@ enum class MemoryLocation {
   GpuToCpu,
 };
 
+enum class MemoryBacking {
+  // Allocates all memory for this resource.
+  Automatic,
+
+  // No memory allocated; physical pages mapped manually on demand.
+  // DX: CreateReservedResource + UpdateTileMappings.
+  // VK: VK_IMAGE_CREATE_SPARSE_BINDING_BIT + vkQueueBindSparse.
+  // Metal: MTLTextureDescriptor.sparseLevel + heap tile mapping
+  // (requires Apple Silicon).
+  Sparse,
+};
+
 enum class IndexFormat { Uint16, Uint32 };
 
 // TODO: Add Unorm types (e.g. R8Unorm, RGBA8Unorm) which can be sampled as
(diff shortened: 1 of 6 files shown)


lib/API/DX/Device.cpp
```diff
@@ -1480,36 +1498,159 @@ class DXDevice : public offloadtest::Device {
...
+      case BufferShaderAccessType::Structured:
+        assert(Desc.AccessTypeParams.StructureStride > 0 &&
+               "Structured buffers must have a Structure Stride.");
+        UAVDesc.Format = DXGI_FORMAT_UNKNOWN;
+        UAVDesc.Buffer.NumElements = static_cast<uint32_t>(
+            SizeInBytes / Desc.AccessTypeParams.StructureStride);
+        UAVDesc.Buffer.StructureByteStride =
+            Desc.AccessTypeParams.StructureStride;
+        break;
+      }
```
EmilioLaiso, line 1625: There's a bunch of duplicated code in here that could be streamlined

manon-traverse: We discussed this offline a bit and came to the conclusion that deduplicating the shared code here would either be just caching a `SizeInBytes / 4` or making the code very complicated with either templated code or macros.
In the end, we decided to leave the code as-is, as it's fairly easy to read in its current state.

EmilioLaiso (approved):

include/API/Resources.h
```diff
@@ -25,6 +25,18 @@ enum class MemoryLocation {
...
 
+enum class MemoryBacking {
+  // Allocates all memory for this resource.
+  Automatic,
+
+  // No memory allocated; physical pages mapped manually on demand.
+  // DX: CreateReservedResource + UpdateTileMappings.
+  // VK: VK_IMAGE_CREATE_SPARSE_BINDING_BIT + vkQueueBindSparse.
+  // Metal: MTLTextureDescriptor.sparseLevel + heap tile mapping
+  // (requires Apple Silicon).
```
bogner: For whatever reason this reads to me like the parenthetical applies to the whole comment even though that doesn't make any sense. Suggest some indentation to clarify.
```suggestion
  // Metal: MTLTextureDescriptor.sparseLevel + heap tile mapping
  //        (requires Apple Silicon).
```

bogner (approved):

bogner:
There are a few places where we didn't update initialization of `BufferCreateDesc` objects. This breaks the build for metal (There's a fix here: https://github.com/llvm/offload-test-suite/pull/1312), but I also see a few places that zero initialize the structure like here:
https://github.com/llvm/offload-test-suite/blob/e8ed152123d03788d8b13ebcf3044a54010bfd92/lib/API/VK/Device.cpp#L3268-L3270

Are these correct? Should we be explicitly specifying the new fields?

PR #1309, opened 2026-06-15
Create, upload, and read back buffers using generic API in DX12.

manon-traverse:
This is a major step in the generic API that wraps around DX12, Metal and Vulkan.

In the DX12 version of `executeProgram` all buffer creation, uploading and reading back is now done using only the generic interface. Only binding of resources to the root signature and setting up descriptor tables/heaps is done using DX12 specific code.

This brings the project another step closer to moving `executeProgram` outside of the graphics API wrapper and into the offloader executable.

Key changes:
- Create CBV, SRV, and UAVs when resources are created
- Create all buffers and textures using `createBuffer` and `createTexture` calls, no more raw DX12 calls
- Perform readbacks and uploads using CommandEncoders
- Remove all DX12 specific code for creating resources and descriptors.

Extra changes:
- VK and Metal implementations of buffer and texture copy functions

Drive by fixes/changes:
- Remove log checks from counter buffer tests that are no longer valid
- Delete the CounterBuffer in Vulkan in the destructor
- Select correct tiling mode for textures that use the `CpuToGpu` or `GpuToCpu` `MemoryLocation`


lib/API/DX/Device.cpp
```diff
@@ -201,155 +143,53 @@ static uint64_t getAlignedTextureBufferSize(const CPUBuffer &B) {
...
-  const D3D12_RESOURCE_FLAGS Flags =
-      R.isReadWrite() ? D3D12_RESOURCE_FLAG_ALLOW_UNORDERED_ACCESS
-                      : D3D12_RESOURCE_FLAG_NONE;
-  const D3D12_RESOURCE_DESC ResDesc = {Dimension, 0,      Width,  Height, 1, 1,
-                                       Format,    {1, 0}, Layout, Flags};
-  return ResDesc;
+    return BufferUsage::ConstantBuffer;
+    break;
+  default:
+    llvm_unreachable("Invalid case, ResourceKind is not a buffer.");
```
bogner: I have a slight preference for listing the non-buffer types here so we have a covered switch, and dropping the debug label. That way if new kinds are added and we forget to update here they'll trigger a covered switch warning rather than potentially hitting this unreachable, saving some headache.

lib/API/DX/Device.cpp
```diff
@@ -201,155 +143,53 @@ static uint64_t getAlignedTextureBufferSize(const CPUBuffer &B) {
...
-  case ResourceKind::SampledTexture2D:
-    llvm_unreachable("Sampled textures aren't supported in DirectX!");
-  case ResourceKind::AccelerationStructure:
-    llvm_unreachable("Acceleration structures use a separate descriptor path!");
-  }
-  return Desc;
+    return BufferShaderAccessType::Raw;
+  default:
+    llvm_unreachable(
+        "Invalid case, non-buffers should have been filtered out.");
```
bogner, line 198: similarly here.

lib/API/DX/Device.cpp
```diff
@@ -646,10 +489,26 @@ class DXCommandBuffer : public offloadtest::CommandBuffer {
...
+    for (size_t I = 0, N = PendingTransitions.size(); I < N; ++I) {
+      auto &Trans = PendingTransitions[I];
       if (Trans.Transition.pResource == Resource) {
         assert(StateBefore == Trans.Transition.StateAfter);
-        Trans.Transition.StateAfter = StateAfter;
+
+        if (StateAfter == Trans.Transition.StateBefore) {
+          // We actually need the resource to be in the original state!
+          // Remove the transition from the list of Pending Transitions.
+          PendingTransitions.erase(PendingTransitions.begin() + I);
```
bogner: Style nit: it's probably slightly clearer to use iterators for the loop, rather than using an index and calculating the correct iterator here.

lib/API/DX/Device.cpp
```diff
@@ -646,10 +489,26 @@ class DXCommandBuffer : public offloadtest::CommandBuffer {
                              D3D12_RESOURCE_STATES StateBefore,
                              D3D12_RESOURCE_STATES StateAfter) {
 
-    for (auto &Trans : PendingTransitions) {
+    for (size_t I = 0, N = PendingTransitions.size(); I < N; ++I) {
+      auto &Trans = PendingTransitions[I];
       if (Trans.Transition.pResource == Resource) {
```
bogner: Would it be clearer to reverse this and continue? ie,
```c++
      if (Trans.Transition.pResource != Resource) {
        continue;
```

This avoids some indentation and I think it's clearer that the `.erase()` call is safe if the `return` isn't nested in a conditional block at all.

lib/API/DX/Device.cpp
```diff
@@ -1761,19 +1743,19 @@ class DXDevice : public offloadtest::Device {
...
       if (!HandleOrErr)
         return HandleOrErr.takeError();
       Tex->DSVHandle = *HandleOrErr;
-      Device->CreateDepthStencilView(DeviceTexture.Get(), nullptr,
+      Device->CreateDepthStencilView(TextureObject.Get(), nullptr,
                                      Tex->DSVHandle);
     }
 
-    return Tex;
+    return std::move(Tex);
```
bogner: Is this necessary? I would expect to get NRVO here without the `std::move`, though if that's the case I'd also expect to see a pessimizing move warning from the compiler

lib/API/MTL/MTLDevice.cpp
```diff
@@ -1413,8 +1450,7 @@ class MTLDevice : public offloadtest::Device {
           for (const auto &Inst : R.first->TLASPtr->Instances)
             Contributions.push_back(Inst.InstanceContributionToHitGroupIndex &
                                     0xFFFFFFu);
-          const BufferCreateDesc Desc{MemoryLocation::GpuToCpu,
-                                      BufferUsage::Storage};
+          const BufferCreateDesc DescBufferCreateDesc::uploadBuffer();
```
bogner: Typo? I guess this is meant to be
```suggestion
          const BufferCreateDesc Desc = BufferCreateDesc::uploadBuffer();
```

In any case, this presumably conflicts with https://github.com/llvm/offload-test-suite/issues/1312 at the moment.

lib/API/DX/Device.cpp
```diff
@@ -2637,21 +2337,29 @@ class DXDevice : public offloadtest::Device {
...
+          if (!RootDescIt->second.back().Buffer) {
+            return llvm::createStringError(
+                std::errc::value_too_large,
+                "Root descriptor can only refer to buffers.");
+          }
+
+          const auto &BufferDX =
+              llvm::cast<DXBuffer>(*RootDescIt->second.back().Buffer);
+          const D3D12_GPU_VIRTUAL_ADDRESS VirtualAddress =
+              BufferDX.Buffer->GetGPUVirtualAddress();
```
bogner: Might be clearer that the expression is repeated here if we use `cast_if_present` and cast the pointer rather than the reference:
```suggestion
          const DXBuffer *BufferDX =
              llvm::cast_if_present<DXBuffer>(RootDescIt->second.back().Buffer);
          if (!BufferDX) {
            return llvm::createStringError(
                std::errc::value_too_large,
                "Root descriptor can only refer to buffers.");
          }

          const D3D12_GPU_VIRTUAL_ADDRESS VirtualAddress =
              BufferDX->Buffer->GetGPUVirtualAddress();
```

bogner (commented):
Looks like the Feature/HLSLLib/PartiallyMappedResources.test test fails with this change, which will need looking into

lib/API/DX/Device.cpp
manon-traverse: Cool I didn't know about that one!

lib/API/DX/Device.cpp
manon-traverse: Fair point!

lib/API/DX/Device.cpp
manon-traverse: Hmmm, not sure why I wrote it like this, because you are right about NRVO.

lib/API/MTL/MTLDevice.cpp
manon-traverse: Oh that is 100% a typo. I think this happened somewhere in a big rebase.

manon-traverse:
> Looks like the Feature/HLSLLib/PartiallyMappedResources.test test fails with this change, which will need looking into

It seems that test is being skipped on my machine due to my Intel GPU. I have added a code path for sparse textures and buffers. They are fully mapped just like before, but there is now also a generic API present for mapping specific tiles.

manon-traverse:
Looks like there is a bug in the sparse buffer impl. I'll fix that tomorrow.

lib/API/DX/Device.cpp
```diff
@@ -1116,7 +1164,7 @@ class DXDevice : public offloadtest::Device {
   // raw ID3D12Device for scratch buffer allocation.
   friend class DXComputeEncoder;
 
-private:
+public:
```
bogner, line 1160: Is this what we want? It seems suspicious to make all of the members of a class specialization public. In any case, if we do want this class to be entirely public the friend declaration above seems redundant

lib/API/MTL/MTLDevice.cpp
```diff
@@ -955,6 +1012,7 @@ class MTLDevice : public offloadtest::Device {
   // and instance buffer allocation.
   friend class MTLComputeEncoder;
 
+public:
```
bogner, line 1011: Similarly here

lib/API/DX/Device.cpp
manon-traverse: To me private has never been that useful of a feature in programming languages. It's more something that gets in the way, and if you really want encapsulation you probably want a proper abstraction anway.

Although I'm fine with either:
1. Writing a getter for the fields I need
2. Just making the whole thing public and get rid of the `friend class`

What decision do you think fits best in the llvm / offload-test-suite code base?

bogner: It's probably fine for these classes to have all public members - use of the derived class is mostly in places where we need to break the general abstraction anyway, and simply adding getters for all the fields isn't super valuable

bogner (approved):
LGTM once the redundant friend declarations have been cleaned up

include/API/Device.h
```diff
@@ -205,6 +205,47 @@ struct SubmitResult {
...
+  GPUAPI getAPI() const { return API; }
+
+protected:
+  explicit MemoryHeap(GPUAPI API) : API(API) {}
+};
+
+/// A region of a Sparse-backed resource, expressed in tiles (not bytes). For
+/// buffers only TileOffsetX / NumTilesX are meaningful; textures additionally
+/// use Subresource and the Y/Z extents. Query the backend tile shape to convert
+/// between texels and tiles.
```
EmilioLaiso: Might be worth mentioning an actual struct/function here rather than just "Query the backend tile shape"

lib/API/DX/Device.cpp
```diff
@@ -201,158 +143,67 @@ static uint64_t getAlignedTextureBufferSize(const CPUBuffer &B) {
...
+  switch (Kind) {
   case ResourceKind::Buffer:
   case ResourceKind::StructuredBuffer:
   case ResourceKind::ByteAddressBuffer:
-  case ResourceKind::RWStructuredBuffer:
   case ResourceKind::RWBuffer:
+  case ResourceKind::RWStructuredBuffer:
   case ResourceKind::RWByteAddressBuffer:
+    return BufferUsage::Storage;
+    break;
```
EmilioLaiso: `break;` can be removed in those `return` statements, right?

EmilioLaiso (approved):

PR #1311, opened 2026-06-15
Remove upload buffer from resource set

manon-traverse:
No longer store the upload buffer in the `ResourceSet`. We don't need to access it anymore after resource creation, we just need to keep it alive.


EmilioLaiso (approved):
Nice! :D

bogner (approved):

manon-traverse:
CI only fails on an irrelevant test unexpectedly passing, safe to merge.

PR #1312, opened 2026-06-15
[Metal] MTLDevice BufferCreateDesc needs to be fully initalized

farzonl:
The Metal\Mac runner builds are failing because we can't create a buffer description. This change full initalize it so we fix the warnings and build errors.

### DIFF
--- a/lib/API/MTL/MTLDevice.cpp
+++ b/lib/API/MTL/MTLDevice.cpp
@@ -1414,7 +1414,11 @@ class MTLDevice : public offloadtest::Device {
             Contributions.push_back(Inst.InstanceContributionToHitGroupIndex &
                                     0xFFFFFFu);
           const BufferCreateDesc Desc{MemoryLocation::GpuToCpu,
-                                      BufferUsage::Storage};
+                                      MemoryBacking::Automatic,
+                                      BufferUsage::Storage,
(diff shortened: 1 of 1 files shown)


lib/API/MTL/MTLDevice.cpp
```diff
@@ -1414,7 +1414,11 @@ class MTLDevice : public offloadtest::Device {
             Contributions.push_back(Inst.InstanceContributionToHitGroupIndex &
                                     0xFFFFFFu);
           const BufferCreateDesc Desc{MemoryLocation::GpuToCpu,
-                                      BufferUsage::Storage};
+                                      MemoryBacking::Automatic,
+                                      BufferUsage::Storage,
+                                      BufferShaderAccessType::Raw,
+                                      {},
+                                      false};
```
bogner, line 1421: Should this use one of the static `BufferCreateDesc` helper functions instead? `BufferCreateDesc::readBackBuffer` I guess?

bogner (approved):
This looks fine to fix the build, but I think we need to follow up on this - I also see a couple of places where we're creating BufferCreateDesc with `{}` elsewhere. I'll follow up with some post-commit comments on https://github.com/llvm/offload-test-suite/pull/1306

PR #1313, opened 2026-06-15
Use the internal lit shell

bogner:
The external lit shell is deprecated as of llvm/llvm-project#203689


bob80905 (approved):

bob80905:
I suppose that variable isn't set in the macos env, so not False = true, and that triggers the error 🤔

bogner:
> I suppose that variable isn't set in the macos env, so not False = true, and that triggers the error 🤔

It defaults to true on windows, which is why most of our CI is happy. I guess the lit config variable will go away at some point.

hekota (approved):

PR #1317, opened 2026-06-17
Pass warp architecture through to lit configuration

llvm-beanz:
This passes the warp architecture through to the lit configurations so that the features are configured correctly.

### DIFF
--- a/.github/workflows/build-and-test-callable.yaml
+++ b/.github/workflows/build-and-test-callable.yaml
@@ -441,9 +441,18 @@ jobs:
             # bare name elsewhere.
             dxc_path="$GITHUB_WORKSPACE/dxc-dist/bin/dxc"
             if [ -x "${dxc_path}.exe" ]; then dxc_path="${dxc_path}.exe"; fi
+            warp_arch_args=()
+            if [ '${{ inputs.OS }}' = 'windows' ]; then
+              if [ '${{ runner.arch }}' = 'ARM64' ]; then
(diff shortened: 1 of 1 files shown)


manon-traverse (approved):

.github/workflows/build-and-test-callable.yaml
```diff
@@ -441,9 +441,18 @@ jobs:
             # bare name elsewhere.
             dxc_path="$GITHUB_WORKSPACE/dxc-dist/bin/dxc"
             if [ -x "${dxc_path}.exe" ]; then dxc_path="${dxc_path}.exe"; fi
+            warp_arch_args=()
+            if [ '${{ inputs.OS }}' = 'windows' ]; then
```
farzonl, line 445: Why do we need to check windows. If the user wants to pick warp windows should be assumed no?

farzonl (approved):

.github/workflows/build-and-test-callable.yaml
llvm-beanz: In the workflow we don't actually know that we're configuring for warp. It probably wouldn't harm anything to set this always (I just didn't trace through all the places where `WARP_ARCHITECTURE` might impact something). It also might be better to just have a more generalized way of feeding lit the host architecture, but I think that would be a bigger change.

bob80905 (approved):

Icohedron (approved):

PR #1318, opened 2026-06-17
Improve logging on mismatched values

llvm-beanz:
This change is just intended to improve the logging for mismatched values by logging the specific value that mismatched (helpful for large comparison buffers) and printing floathex or hex depending on of the value is a floating point type.

### DIFF
--- a/lib/Support/Check.cpp
+++ b/lib/Support/Check.cpp
@@ -19,6 +19,7 @@
 #include <cmath>
 #include <cstring>
 #include <sstream>
+#include <type_traits>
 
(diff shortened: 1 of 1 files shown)


V-FEXrt (approved):

manon-traverse (approved):

bob80905 (approved):
This LGTM, but I must say I do like the early return like before.
If we encounter a large buffer array like you mention, then we may see "Mismatch at element..." repeated hundreds of times, when only the first mismatch might be enough for our purposes.
It could potentially blow up the log size on the pipelines.

PR #1325, opened 2026-06-22
[DX][Metal] Enable array-of-constant-buffers.test

farzonl:
with https://github.com/llvm/llvm-project/pull/204732 merged this test is now passing. Remove the DX and Metal xfails.

### DIFF
--- a/test/Feature/ResourceArrays/array-of-constant-buffers.test
+++ b/test/Feature/ResourceArrays/array-of-constant-buffers.test
@@ -67,11 +67,6 @@ DescriptorSets:
 ...
 #--- end
 
-# CBuffer<T> not yet implemented for DirectX
-# Unimplemented https://github.com/llvm/llvm-project/issues/122791
-# XFAIL: Clang && DirectX
-# XFAIL: Clang && Metal
-
 # RUN: split-file %s %t
 # RUN: %dxc_target -T cs_6_0 -Fo %t.o %t/source.hlsl
 # RUN: %offloader %t/pipeline.yaml %t.o


bogner (approved):

test/Feature/ResourceArrays/array-of-constant-buffers.test
```diff
@@ -67,11 +67,6 @@ DescriptorSets:
 ...
 #--- end
 
-# CBuffer<T> not yet implemented for DirectX
-# Unimplemented https://github.com/llvm/llvm-project/issues/122791
-# XFAIL: Clang && DirectX
-# XFAIL: Clang && Metal
```
farzonl, line 73: We probably need to leave this XFAIL and just do DX. because Metal runs are green right now.

Icohedron (approved):

test/Feature/ResourceArrays/array-of-constant-buffers.test
farzonl: Nevermind looks like removing the xfail has no impact clang metal runner is passing with this change: https://github.com/llvm/offload-test-suite/actions/runs/27965920506/job/82759548461?pr=1325

Might investigate a bit more to see if this test is disabled by other means.

Icohedron: Test seems to be passing on DX. So both XFAIL lines should be removed.
Intel Clang DX showing XPASS for example: https://github.com/llvm/offload-test-suite/actions/runs/27957974299/job/82742409306#step:12:176

farzonl: Looks like it is passing on Metal in this pr. Not sure why mac runners are green without the change.
```
PASS: OffloadTest-clang-mtl :: Feature/ResourceArrays/array-of-constant-buffers.test (309 of 545)
```

farzonl: hmm nevermind it was failing on metal too https://github.com/llvm/offload-test-suite/actions/runs/27961998346
looks like my browser was just caching the  pass sticker.

V-FEXrt (approved):

PR #1327, opened 2026-06-22
[NFC] Express GPU texture formats via CPUBuffer GpuFormat field

### LINKED ISSUE #1326: Express GPU texture formats directly via a CPUBuffer GpuFormat field
## Problem

Today, special texture formats are folded into the `DataFormat` enum (e.g. `DataFormat::Depth32`) and the GPU `Format` is inferred from `DataFormat` + `Channels` via `toFormat()`. This has two drawbacks:

1. **Doesn't scale** — every new special format requires extending `DataFormat`.
2. **Can't express depth-stencil formats** — `D32FloatS8Uint`'s stencil component is a separate `uint8`, not a second channel, so it is not representable as `DataFormat` + `Channels`.

`DataFormat` and the GPU `Format` are really two different axes:

- **`DataFormat`** (e.g. `Float32`) — how to interpret the **CPU bytes**.
- **GPU `Format`** (e.g. `D32Float`) — the **GPU resource format**.

## Proposal

Add an optional `GpuFormat` field to `CPUBuffer` that names a GPU `Format` directly, bypassing the `DataFormat -> Format` round-trip, and remove the special-cased `DataFormat::Depth32`. Backends can then just check `isDepth` on the format rather than special-casing an enum value.

```yaml
# Before
- Name: DepthTex
  Format: Depth32

# After
- Name: DepthTex
  Format: Float32        # how to read the CPU bytes
  GpuFormat: D32Float    # explicit GPU resource format
```

This is cleaner and extensible to any future GPU format, including depth-stencil formats like `D32FloatS8Uint`.

## Credit

Based on review feedback from @manon-traverse on #1229, who suggested using the `Format` enum directly in the YAML and checking `isDepth` instead of converting, to support depth formats that include a stencil buffer.

alsepkow:
Resolves #1326.

Replaces the special-cased `DataFormat::Depth32` with an optional `CPUBuffer::GpuFormat` field that names a GPU `Format` directly. The Vulkan backend now derives the image/view format and depth/color aspect from `GpuFormat`; the `Check` path drops the unused `Depth32` cases. Existing `GatherCmp`/`SampleCmp` depth tests migrate from `Format: Depth32` to `Format: Float32` + `GpuFormat: D32Float`.

See #1326 for full motivation.

### DIFF
--- a/include/Support/Pipeline.h
+++ b/include/Support/Pipeline.h
@@ -107,7 +107,6 @@ enum class DataFormat {
   Float16,
   Float32,
   Float64,
-  Depth32,
   Bool,
 };
 
@@ -198,7 +197,6 @@ static inline uint32_t getFormatSize(DataFormat Format) {
   case DataFormat::UInt32:
   case DataFormat::Int32:
   case DataFormat::Float32:
-  case DataFormat::Depth32:
   case DataFormat::Bool:
     return 4;
   case DataFormat::Hex64:
@@ -216,6 +214,11 @@ struct CPUBuffer {
   int Channels;
   int Stride;
   uint32_t ArraySize;
+  // When set, names the GPU texture format directly (e.g. D32Float) instead of
+  // inferring it from DataFormat + Channels via toFormat(). This lets depth
+  // buffers and other special formats be expressed without extending
+  // DataFormat.
+  std::optional<offloadtest::Format> GpuFormat;
   // Data can contain one block of data for a singular resource
   // or multiple blocks for a resource array.
   llvm::SmallVector<std::unique_ptr<char[]>> Data;
@@ -931,7 +934,6 @@ template <> struct ScalarEnumerationTraits<offloadtest::DataFormat> {
     ENUM_CASE(Float16);
     ENUM_CASE(Float32);
     ENUM_CASE(Float64);
-    ENUM_CASE(Depth32);
     ENUM_CASE(Bool);
 #undef ENUM_CASE
   }
--- a/include/Support/Pipeline.h
+++ b/include/Support/Pipeline.h
@@ -218,7 +218,7 @@ struct CPUBuffer {
   // inferring it from DataFormat + Channels via toFormat(). This lets depth
   // buffers and other special formats be expressed without extending
   // DataFormat.
-  std::optional<offloadtest::Format> GpuFormat;
+  std::optional<offloadtest::Format> GPUFormat;
   // Data can contain one block of data for a singular resource
   // or multiple blocks for a resource array.
   llvm::SmallVector<std::unique_ptr<char[]>> Data;
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -2560,8 +2560,8 @@ public:
                                          "supported for DirectX textures.");
 
         auto FormatOrErr =
-            R.BufferPtr->GpuFormat
-                ? llvm::Expected<Format>(*R.BufferPtr->GpuFormat)
+            R.BufferPtr->GPUFormat
+                ? llvm::Expected<Format>(*R.BufferPtr->GPUFormat)
                 : toFormat(R.BufferPtr->Format, R.BufferPtr->Channels);
         if (!FormatOrErr)
           return FormatOrErr.takeError();
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -2565,7 +2565,6 @@ public:
                 : toFormat(R.BufferPtr->Format, R.BufferPtr->Channels);
         if (!FormatOrErr)
           return FormatOrErr.takeError();
-        const Format Fmt = *FormatOrErr;
 
         LLVM_ENABLE_BITMASK_ENUMS_IN_NAMESPACE();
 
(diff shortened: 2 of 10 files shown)


include/Support/Pipeline.h
```diff
@@ -216,6 +214,11 @@ struct CPUBuffer {
   int Channels;
   int Stride;
   uint32_t ArraySize;
+  // When set, names the GPU texture format directly (e.g. D32Float) instead of
+  // inferring it from DataFormat + Channels via toFormat(). This lets depth
+  // buffers and other special formats be expressed without extending
+  // DataFormat.
+  std::optional<offloadtest::Format> GpuFormat;
```
inbelic: nit:
```suggestion
  std::optional<offloadtest::Format> GPUFormat;
```
to match CPU elsewhere

lib/API/VK/Device.cpp
```diff
@@ -3395,13 +3391,15 @@ class VulkanDevice : public offloadtest::Device {
   llvm::Expected<ResourceRef> createImage(Resource &R, BufferRef &Host,
                                           int UsageOverride = 0) {
     const offloadtest::CPUBuffer &B = *R.BufferPtr;
-    if (B.Format == DataFormat::Depth32 && R.isReadWrite())
+    const bool IsDepth = B.GpuFormat.has_value() && isDepthFormat(*B.GpuFormat);
```
inbelic: nit:
```suggestion
    const bool IsDepth = B.GpuFormat.has_value() && isDepthFormat(B.GpuFormat.value());
```
or
```suggestion
    const bool IsDepth = B.GpuFormat && isDepthFormat(*B.GpuFormat);
```

test/Feature/Textures/Texture2D.SampleCmp.test.yaml
```diff
@@ -107,7 +107,8 @@ Shaders:
 
 Buffers:
   - Name: Tex
-    Format: Depth32
+    Format: Float32
+    GpuFormat: D32Float
     Channels: 1
```
inbelic, line 112: IIUC, if GpuFormat is specified then both Format and Channels are ignored. Is there any easy way to represent that in the struct/yaml so they are not silently ignored. Maybe a `std::variant` works?

alsepkow: Your general intuition for the `std::variant` makes sense. But `GPUFormat` only implies that `Channels` are ignored. `Format` is still used for how we interpret the CPU bytes.

include/Support/Pipeline.h
```diff
@@ -216,6 +214,11 @@ struct CPUBuffer {
   int Channels;
   int Stride;
   uint32_t ArraySize;
+  // When set, names the GPU texture format directly (e.g. D32Float) instead of
+  // inferring it from DataFormat + Channels via toFormat(). This lets depth
+  // buffers and other special formats be expressed without extending
+  // DataFormat.
+  std::optional<offloadtest::Format> GPUFormat;
```
manon-traverse, line 221: I love this change. In a future PR we should completely get rid of the old way of doing formats and just use the ones that are actually GPU API compatible.

test/Feature/Textures/Texture2D.SampleCmp.test.yaml
manon-traverse: Why wouldn't we be able to also use GPUFormat to interpret the human-readable data when reading the YAML?

manon-traverse (approved):

test/Feature/Textures/Texture2D.GatherCmp.test.yaml
```diff
@@ -68,7 +68,8 @@ Shaders:
 
 Buffers:
   - Name: Tex
-    Format: Depth32
+    Format: Float32
+    GPUFormat: D32Float
```
bogner, line 72: ~Is the `Format: Float32` just entirely ignored here since `GPUFormat` is specified? This feels a bit confusing.~

I guess it isn't ignored but it's how we're interpreting the values in the YAML?

lib/API/DX/Device.cpp
```diff
@@ -2559,7 +2559,10 @@ class DXDevice : public offloadtest::Device {
                                          "Multiple mip levels are not yet "
                                          "supported for DirectX textures.");
 
-        auto FormatOrErr = toFormat(R.BufferPtr->Format, R.BufferPtr->Channels);
+        auto FormatOrErr =
+            R.BufferPtr->GPUFormat
+                ? llvm::Expected<Format>(*R.BufferPtr->GPUFormat)
+                : toFormat(R.BufferPtr->Format, R.BufferPtr->Channels);
```
bogner, line 2565: I think it would be more sound to update `toFormat` to take all three of the GPUFormat, Format, and Channels and reconcile them. This way we can at least assert if we've somehow specified both in an incompatible way.

lib/API/VK/Device.cpp
```diff
@@ -3395,13 +3391,15 @@ class VulkanDevice : public offloadtest::Device {
...
+    const bool IsDepth = B.GPUFormat && isDepthFormat(*B.GPUFormat);
+    if (IsDepth && R.isReadWrite())
       return llvm::createStringError(std::errc::invalid_argument,
                                      "Image memory allocation failed.");
     VkImageCreateInfo ImageCreateInfo = {};
     ImageCreateInfo.sType = VK_STRUCTURE_TYPE_IMAGE_CREATE_INFO;
     ImageCreateInfo.imageType = getVKImageType(R.Kind);
-    ImageCreateInfo.format = getVKFormat(B.Format, B.Channels);
+    ImageCreateInfo.format = B.GPUFormat ? getVulkanFormat(*B.GPUFormat)
+                                         : getVKFormat(B.Format, B.Channels);
```
bogner, line 3402: Similarly here, if we update `getVKFormat` to reconcile the specification first then we don't need to have all of these places with their own checks that need to be aware of which way we've specified the format

bogner (commented):
I worry that this change as is makes things pretty confusing. There are now two slightly incompatible ways to specify the format, and as far as I can tell nothing stops you from doing both.

I realize it would be a massive change to all of the tests to just switch to GPUFormat entirely, but if this is the direction we're going I think we might be better off just ripping off the bandaid and changing it all.

PR #1329, opened 2026-06-23
[CI] Add a VK and DX only test all tag

farzonl:
The test-all tag is a bit to expensive when we are only testing DIXL or SPIRV related changes. This PR allows us to pick only the runtime we need to test.


bob80905 (approved):
LGTM. Though I would mention at least in the PR description why one label has a different set of SKUs than the other.

PR #1330, opened 2026-06-24
Move resource creation, readbacks, and InvocationState from DX backend to shared file.

manon-traverse:
Move code that doesn't need to live in the DX backend to a shared file.

Why make this change?
1. All this code should eventually live the offloader application. However, it can't live there right now because the rendering backends are too divergent.
2. In a follow-up PR we can reuse this code in the Vulkan and Metal backends resulting in a very significant reduction in code size.

This PR doesn't make any other code changes other than splitting the creation of the descriptor tables into a separate function from resource creation, due to no abstraction existing for building descriptor sets/tables and binding those yet.

### DIFF
--- a/test/Feature/Textures/Texture2D.OperatorIndex.test.yaml
+++ b/test/Feature/Textures/Texture2D.OperatorIndex.test.yaml
@@ -60,9 +60,6 @@ Results:
 ...
 #--- end
 
-# Unimplemented: Clang + DX: https://github.com/llvm/llvm-project/issues/101558
-# XFAIL: DirectX
-
 # RUN: split-file %s %t
 # RUN: %dxc_target -T cs_6_0 -Fo %t.o %t/source.hlsl
 # RUN: %offloader %t/pipeline.yaml %t.o
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -50,6 +50,8 @@
 
 #include "../Util.h"
 
+#include "../Support/OffloadMigration.h"
+
 #include <atomic>
 #include <codecvt>
 #include <locale>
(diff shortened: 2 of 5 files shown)


test/Feature/Textures/Texture2D.OperatorIndex.test.yaml
```diff
@@ -60,9 +60,6 @@ Results:
 ...
 #--- end
 
-# Unimplemented: Clang + DX: https://github.com/llvm/llvm-project/issues/101558
-# XFAIL: DirectX
```
bogner, line 64: Shouldn't this change be NFC? Why does this start passing?

bogner (approved):
Will this stay in `Support/` long term, or is the intent to eventually move it again once everything is migrated?

Overall LGTM on the assumption that the follow ups are where doing this will really shine.

manon-traverse:
> Will this stay in `Support/` long term, or is the intent to eventually move it again once everything is migrated?
> 
> Overall LGTM on the assumption that the follow ups are where doing this will really shine.

The eventual goal is that all of this will live in the offload application, since this is all just code that is specific to running a .test file, i.e. what the offloader is for.

test/Feature/Textures/Texture2D.OperatorIndex.test.yaml
manon-traverse: This stems from the fact that DX12 already supported mip levels, but the mip levels were forced to 1. By using the mip level count from the `Resource` instead of clamping it this test started succeeding.

manon-traverse:
This PR is only failing on irrelevant tests due to a bug in the CI system itself. Safe to merge.

PR #1331, opened 2026-06-24
[DX12] Add support for Samplers in DX12.

### LINKED ISSUE #664: [DX][MTL] Implement Samplers for the Directx and Metal backends.
I have added the ability to define samplers in the YAML test, and the Vulkan backend is able to create and bind them. However, I did not implement this for any other API.

manon-traverse:
Adds support for samplers in DX12.
This change is a lot larger than you would expect from adding samplers.
This is because samplers need to live in a separate DescriptorHeap which cascaded into a couple of issues:
* Building the descriptor tables requires building tables for both sampler and CBV/SRV/UAV descriptors.
* The binding of descriptor tables can no longer be implied based on the DescriptorSets defined in .test file
* A layout must be generated when the root signature is generated based on the bindings
* In case of root signatures defined in the shader, the layout must be extracted from that root signature by deserializing it.

Adding this feature does allow more tests to succeed on DX12.

resolves #664

### DIFF
--- a/include/API/Sampler.h
+++ b/include/API/Sampler.h
@@ -1,0 +1,67 @@
+//===- Sampler.h - Offload API Texture ------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef OFFLOADTEST_API_SAMPLER_H
+#define OFFLOADTEST_API_SAMPLER_H
+
+#include "API/API.h"
+#include "API/Resources.h"
+
+namespace offloadtest {
+
+enum class FilterMode { Nearest, Linear };
+
+enum class AddressMode { Clamp, Repeat, Mirror, Border, MirrorOnce };
+
+enum class CompareFunction {
+  Never,
+  Less,
+  Equal,
+  LessEqual,
+  Greater,
+  NotEqual,
+  GreaterEqual,
+  Always
+};
+
+enum class SamplerKind { Sampler, SamplerComparison };
+
+struct SamplerCreateDesc {
+  FilterMode MinFilter = FilterMode::Linear;
+  FilterMode MagFilter = FilterMode::Linear;
+  AddressMode Address = AddressMode::Clamp;
+  float MinLOD = 0.0f;
+  float MaxLOD = std::numeric_limits<float>::max();
+  float MipLODBias = 0.0f;
+  CompareFunction ComparisonOp = CompareFunction::Never;
+  SamplerKind Kind = SamplerKind::Sampler;
+};
+
+class Sampler {
+  GPUAPI API;
+
+public:
+  virtual ~Sampler();
+  Sampler(const Sampler &) = delete;
+  // Sampler(Sampler &&) = delete;
+  Sampler &operator=(const Sampler &) = delete;
+  // Sampler &operator=(Sampler &&) = delete;
+
+  GPUAPI getAPI() const { return API; }
+  virtual const SamplerCreateDesc &getDesc() const = 0;
+
+protected:
+  explicit Sampler(GPUAPI API) : API(API) {}
+};
+
+} // namespace offloadtest
+
+#endif // OFFLOADTEST_API_SAMPLER_H
--- a/include/API/Sampler.h
+++ b/include/API/Sampler.h
@@ -1,11 +1,8 @@
-//===- Sampler.h - Offload API Texture ------------------------------------===//
+//===- Sampler.h - Offload API Sampler ------------------------------------===//
 //
 // Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-//
-//===----------------------------------------------------------------------===//
-//
 //
 //===----------------------------------------------------------------------===//
 
@@ -51,9 +48,7 @@ class Sampler {
 public:
   virtual ~Sampler();
   Sampler(const Sampler &) = delete;
-  // Sampler(Sampler &&) = delete;
   Sampler &operator=(const Sampler &) = delete;
-  // Sampler &operator=(Sampler &&) = delete;
 
   GPUAPI getAPI() const { return API; }
   virtual const SamplerCreateDesc &getDesc() const = 0;
--- a/include/API/Texture.h
+++ b/include/API/Texture.h
@@ -105,12 +105,6 @@ inline llvm::Error validateTextureCreateDesc(const TextureCreateDesc &Desc) {
         std::errc::not_supported,
         "DepthStencil combined with Storage is not yet supported.");
 
-  // Depth formats require DepthStencil usage; non-depth formats forbid it.
-  if (IsDepth && !IsDS)
-    return llvm::createStringError(
-        std::errc::invalid_argument,
-        "Depth format '%s' requires DepthStencil usage.",
-        getFormatName(Desc.Fmt).data());
   if (!IsDepth && IsDS)
     return llvm::createStringError(
         std::errc::invalid_argument,
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -141,6 +141,67 @@ getDXPrimitiveTopology(PrimitiveTopology Topology,
   llvm_unreachable("All PrimitiveTopology cases handled");
 }
 
+static D3D12_FILTER getDXFilterMode(FilterMode MinFilter, FilterMode MagFilter,
+                                    bool IsComparison) {
+  if (IsComparison) {
+    if (MinFilter == FilterMode::Nearest)
+      return MagFilter == FilterMode::Nearest
+                 ? D3D12_FILTER_COMPARISON_MIN_MAG_MIP_POINT
+                 : D3D12_FILTER_COMPARISON_MIN_POINT_MAG_LINEAR_MIP_POINT;
+    else
+      return MagFilter == FilterMode::Nearest
+                 ? D3D12_FILTER_COMPARISON_MIN_LINEAR_MAG_MIP_POINT
+                 : D3D12_FILTER_COMPARISON_MIN_MAG_LINEAR_MIP_POINT;
+  } else {
+    if (MinFilter == FilterMode::Nearest)
+      return MagFilter == FilterMode::Nearest
+                 ? D3D12_FILTER_MIN_MAG_MIP_POINT
+                 : D3D12_FILTER_MIN_POINT_MAG_LINEAR_MIP_POINT;
+    else
+      return MagFilter == FilterMode::Nearest
+                 ? D3D12_FILTER_MIN_LINEAR_MAG_MIP_POINT
+                 : D3D12_FILTER_MIN_MAG_LINEAR_MIP_POINT;
+  }
+}
+
+static D3D12_TEXTURE_ADDRESS_MODE getDXTextureAddressMode(AddressMode Mode) {
+  switch (Mode) {
+  case AddressMode::Clamp:
+    return D3D12_TEXTURE_ADDRESS_MODE_CLAMP;
+  case AddressMode::Repeat:
+    return D3D12_TEXTURE_ADDRESS_MODE_WRAP;
+  case AddressMode::Mirror:
+    return D3D12_TEXTURE_ADDRESS_MODE_MIRROR;
+  case AddressMode::Border:
+    return D3D12_TEXTURE_ADDRESS_MODE_BORDER;
+  case AddressMode::MirrorOnce:
+    return D3D12_TEXTURE_ADDRESS_MODE_MIRROR_ONCE;
+  }
+  llvm_unreachable("All cases handled.");
+}
+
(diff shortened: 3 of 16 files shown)


lib/API/DX/Device.cpp
```diff
@@ -2537,17 +2861,44 @@ class DXDevice : public offloadtest::Device {
...
+          SamplerHeap->GetGPUDescriptorHandleForHeapStart());
+
+      const uint32_t Inc = Device->GetDescriptorHandleIncrementSize(
+          D3D12_DESCRIPTOR_HEAP_TYPE_CBV_SRV_UAV);
+      const uint32_t SamplerInc = Device->GetDescriptorHandleIncrementSize(
+          D3D12_DESCRIPTOR_HEAP_TYPE_SAMPLER);
+
+      for (uint32_t I = 0, N = DXPipeline.Layout.size(); I < N; ++I) {
+        const auto &Layout = DXPipeline.Layout[I];
+        llvm::outs() << "Layout.Count: " << Layout.Count << "\n";
```
EmilioLaiso: This debug print can be removed right?

lib/API/DX/Device.cpp
```diff
@@ -243,10 +303,47 @@ class DXTexture : public offloadtest::Texture {
...
+
+struct RoogtSignatureLayout {
+  RootParameterType ParameterType : 3;
+  uint32_t Count : 29;
+
+  RoogtSignatureLayout(RootParameterType ParameterType, uint32_t Count)
+      : ParameterType(ParameterType), Count(Count) {}
+  RoogtSignatureLayout() = delete;
+};
+
```
EmilioLaiso, line 341: Could've probably been two PRs, but at least this way it doesn't only introduces new API but it also shows how it's used

EmilioLaiso (approved):

include/API/Texture.h
```diff
@@ -105,12 +105,6 @@ inline llvm::Error validateTextureCreateDesc(const TextureCreateDesc &Desc) {
         std::errc::not_supported,
         "DepthStencil combined with Storage is not yet supported.");
 
-  // Depth formats require DepthStencil usage; non-depth formats forbid it.
-  if (IsDepth && !IsDS)
-    return llvm::createStringError(
-        std::errc::invalid_argument,
-        "Depth format '%s' requires DepthStencil usage.",
-        getFormatName(Desc.Fmt).data());
```
MarijnS95, line 113: Out of scope / unrelated?

lib/API/DX/Device.cpp
```diff
@@ -1279,12 +1446,49 @@ class DXDevice : public offloadtest::Device {
...
+          continue;
+
+        Ranges.get()[RangeIdx].RangeType = D3D12_DESCRIPTOR_RANGE_TYPE_SAMPLER;
+        Ranges.get()[RangeIdx].NumDescriptors = Binding.DescriptorCount;
+        Ranges.get()[RangeIdx].BaseShaderRegister = Binding.DXBinding.Register;
+        Ranges.get()[RangeIdx].RegisterSpace = Binding.DXBinding.Space;
+        Ranges.get()[RangeIdx].OffsetInDescriptorsFromTableStart =
+            SamplerDescriptorIdx;
+
+        assert(Binding.DescriptorCount == 1 && "Manon expected this to be 1.");
```
MarijnS95: Debug text?

lib/API/DX/Device.cpp
```diff
@@ -243,10 +303,47 @@ class DXTexture : public offloadtest::Texture {
...
+enum class RootParameterType : uint32_t {
+  DescriptorTable = 0,
+  SamplerTable,
+  Constant,
+  CBV,
+  SRV,
+  UAV,
+};
+
+struct RoogtSignatureLayout {
```
MarijnS95: I am groot (roogt?)



```suggestion
struct RootSignatureLayout {
```

include/API/Sampler.h
```diff
@@ -0,0 +1,67 @@
+//===- Sampler.h - Offload API Texture ------------------------------------===//
```
MarijnS95:
```suggestion
//===- Sampler.h - Offload API Sampler ------------------------------------===//
```

include/API/Sampler.h
```diff
@@ -0,0 +1,67 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+//
+//===----------------------------------------------------------------------===//
```
MarijnS95:
```suggestion
```

include/API/Sampler.h
```diff
@@ -0,0 +1,67 @@
...
+
+class Sampler {
+  GPUAPI API;
+
+public:
+  virtual ~Sampler();
+  Sampler(const Sampler &) = delete;
+  // Sampler(Sampler &&) = delete;
+  Sampler &operator=(const Sampler &) = delete;
+  // Sampler &operator=(Sampler &&) = delete;
```
MarijnS95:
```suggestion
  Sampler &operator=(const Sampler &) = delete;
```

lib/API/DX/Device.cpp
```diff
@@ -2331,15 +2612,26 @@ class DXDevice : public offloadtest::Device {
   }
 
   void buildDescriptorTables(llvm::ArrayRef<DescriptorTable> DescTables,
-                             const ComPtr<ID3D12DescriptorHeap> &DescHeap) {
+                             const ComPtr<ID3D12DescriptorHeap> &DescHeap,
+                             const ComPtr<ID3D12DescriptorHeap> &SamplerHeap) {
     // Bind descriptors in descriptor tables.
-    if (!DescHeap)
+    if (!DescHeap && !SamplerHeap)
```
MarijnS95:
```suggestion
    if (!DescHeap || !SamplerHeap)
```

Because both are unconditionally dereferenced afterwards.

lib/API/DX/Device.cpp
```diff
@@ -2488,9 +2799,22 @@ class DXDevice : public offloadtest::Device {
       // If no explicit root parameters are provided, fall back to using the
       // descriptor set layout. This is to make it easier to write tests that
       // don't need complicated root signatures.
-      for (uint32_t Idx = 0u; Idx < P.Sets.size(); ++Idx) {
-        DXCB.CmdList->SetComputeRootDescriptorTable(Idx, Handle);
-        Handle.Offset(P.Sets[Idx].Resources.size(), Inc);
+      for (uint32_t I = 0, N = DXPipeline.Layout.size(); I < N; ++I) {
```
MarijnS95, line 2811: Claude thinks this loop can be deduplicated leaving only the graphics vs compute-specific calls on the API.

lib/API/DX/Device.cpp
```diff
@@ -2444,6 +2753,7 @@ class DXDevice : public offloadtest::Device {
           break;
         }
         case dx::RootParamKind::DescriptorTable:
+          // TODO(manon): Add support for descriptor tables containing samplers
           DXCB.CmdList->SetComputeRootDescriptorTable(RootParamIndex++, Handle);
           Handle.Offset(P.Sets[DescriptorTableIndex++].Resources.size(), Inc);
           break;
```
MarijnS95, line 2768: Samplers live in a separate heap, so a `DescriptorTable` mixing a sampler with CBV/SRV/UAV resources binds the wrong heap and `Handle.Offset()` over-advances past the descriptors actually written (no `SamplerTable` case here either). Hard-error until it's supported (`case` needs braces for the local):

```suggestion
        case dx::RootParamKind::DescriptorTable: {
          const DescriptorSet &Set = P.Sets[DescriptorTableIndex++];
          // TODO(manon): Add support for descriptor tables containing samplers.
          for (const auto &R : Set.Resources)
            if (R.isSampler())
              return llvm::createStringError(
                  "Descriptor tables containing samplers are not yet "
                  "supported with explicit RootParameters.");
          DXCB.CmdList->SetComputeRootDescriptorTable(RootParamIndex++, Handle);
          Handle.Offset(Set.Resources.size(), Inc);
          break;
        }
```

MarijnS95 (commented):
A few more, on top of the inline notes:

- **`Texture.h`** — dropping the "depth format requires DepthStencil usage" check is unrelated to samplers, and the depth-as-SRV path it enables still isn't correct (typed `D32_FLOAT` can't back an SRV without `TYPELESS`; those tests stay `XFAIL`). Maybe split out.
- **Implicit fallback** now hard-errors on shader-embedded inline descriptors / root constants when no explicit `RootParameters` are given (previously bound each set as a table). No test hits it, but it's a behavior change.
- **`assert(... "Manon expected this to be 1.")`** silently caps sampler arrays at 1 — prefer a real `createStringError` stating the invariant.
- **`RoogtSignatureLayout`** (and the error string `RootParamters`) is misspelled.

Cleared on inspection: `SamplerCreateDesc` brace-init order matches the struct, the filter/comparison/address enum mappings are correct, and the 256-cap `SamplerAllocator` matches the existing allocator pattern.

lib/API/DX/Device.cpp
```diff
@@ -1269,7 +1435,8 @@ class DXDevice : public offloadtest::Device {
           Ranges.get()[RangeIdx].RangeType = D3D12_DESCRIPTOR_RANGE_TYPE_CBV;
           break;
         case DescriptorKind::SAMPLER:
-          llvm_unreachable("Not implemented yet."); // Requires a separate heap
+          llvm_unreachable("Sampler should have been filtered out.");
```
bob80905, line 1438: I would think this is dead code because of the continue on line 1425

manon-traverse: Yes, which is why I put the unreachable statement there. I want all cases to be handled instead of relying on the default case label.

lib/API/DX/Device.cpp
manon-traverse: We actually force there to be a SamplerHeap if we are having a DescHeap, so I will change it to just checking for the DescHeap. This is code that is being removed in follow-up PRs though.

lib/API/DX/Device.cpp
manon-traverse: I added an error check a bit earlier to check there are no samplers when RootParams are specified.

lib/API/DX/Device.cpp
manon-traverse: Yes, but we are tackling this code in a follow-up PR anyway so I don't see a lot of value in deduplicating this.

lib/API/DX/Device.cpp
MarijnS95: Bumping this — still `RoogtSignatureLayout` (~20 uses) and the `RootParamters` typo in the error strings (lines 2825, 2907). Both are merge-once-and-it-sticks: renaming a 20-site symbol in a follow-up is pure churn, and `RootParamters` is user-facing error text. It's a mechanical `s/Roogt/Root/` + `s/RootParamters/RootParameters/` while the branch is still open — would prefer to land it here.

lib/API/DX/Device.cpp
```diff
@@ -2488,9 +2808,22 @@ class DXDevice : public offloadtest::Device {
...
+          Handle.Offset(Layout.Count, Inc);
+          break;
+        case RootParameterType::SamplerTable:
+          DXCB.CmdList->SetComputeRootDescriptorTable(I, SamplerHandle);
+          SamplerHandle.Offset(Layout.Count, SamplerInc);
+          break;
+        default:
+          return llvm::createStringError(
+              "Root Signatures that contain constants and inline descriptors "
+              "require custom RootParamters to be defined.");
```
MarijnS95:
```suggestion
              "require custom RootParameters to be defined.");
```

lib/API/DX/Device.cpp
```diff
@@ -2537,17 +2870,43 @@ class DXDevice : public offloadtest::Device {
...
+          Handle.Offset(Layout.Count, Inc);
+          break;
+        case RootParameterType::SamplerTable:
+          DXCB.CmdList->SetGraphicsRootDescriptorTable(I, SamplerHandle);
+          SamplerHandle.Offset(Layout.Count, SamplerInc);
+          break;
+        default:
+          return llvm::createStringError(
+              "Root Signatures that contain constants and inline descriptors "
+              "require custom RootParamters to be defined.");
```
MarijnS95:
```suggestion
              "require custom RootParameters to be defined.");
```

MarijnS95 (approved):

include/API/Texture.h
manon-traverse: Some of the sampler tests failed on textures with a depth format, but without a depth stencil format being created. While this is unusual, it's not incorrect.

damyanp:
Tagging @alsepkow for awareness.

bogner (approved):

bogner:
Looks like we need to update one more test here

```diff
--- a/test/Feature/Textures/Texture2D.SampleBias.test.yaml
+++ b/test/Feature/Textures/Texture2D.SampleBias.test.yaml
@@ -151,7 +151,8 @@ Results:
 #--- end

 # Unimplemented: https://github.com/llvm/offload-test-suite/issues/664
-# XFAIL: DirectX || Metal || Vulkan && Darwin
+# XFAIL: Metal || Vulkan && Darwin
+# XFAIL: Clang && DirectX

 # Unimplemented: https://github.com/llvm/llvm-project/issues/175630
 # XFAIL: Clang && !Intel
```

bogner:
I've put up https://github.com/llvm/offload-test-suite/pull/1403 to resolve the merge conflicts and update newer tests as needed. We'll finish this PR there.

bogner:
Merged in https://github.com/llvm/offload-test-suite/pull/1403

PR #1332, opened 2026-06-24
Add mipmap support to createTextureWithData.

### LINKED ISSUE #1039: [DirectX] Add multiple mip level support
This is required to support testing of the `.mips[][]` operator.

This functionality is already supported in Vulkan. Placeholder errors are currently generated when targeting DirectX.

This issue is completed when we have equivalent support for DirectX.

The vulkan implementation for reference [here](https://github.com/llvm/offload-test-suite/pull/708).

The first step to completion is to update this issue with a comprehensive plan.

AC:
- [ ] Implement equivalent functionality for DirectX
- [ ] Add independent testing of functionality
- [ ] All existing tests referencing this ticket are passing for DXC and clang (where applicable)

manon-traverse:
createTextureWithData now also uploads mipmap when creating a texture.

This required introducing a new function to query the texture footprint, as well as adding support for copying mips to the copyBufferToTexture function.

Resolves https://github.com/llvm/offload-test-suite/issues/1039

### DIFF
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -789,15 +789,20 @@ public:
                                D3D12_RESOURCE_STATE_COPY_DEST);
     CB.flushBarrier();
 
-    const uint32_t ElementSize = getFormatSizeInBytes(DXDst.Desc.Fmt);
-    const D3D12_PLACED_SUBRESOURCE_FOOTPRINT Footprint{
-        0,
-        CD3DX12_SUBRESOURCE_FOOTPRINT(
-            getDXGIFormat(DXDst.Desc.Fmt), DXDst.Desc.Width, DXDst.Desc.Height,
-            1, getAlignedTexturePitch(DXDst.Desc.Width, ElementSize))};
-    const CD3DX12_TEXTURE_COPY_LOCATION DstLoc(DXDst.Resource.Get(), 0);
-    const CD3DX12_TEXTURE_COPY_LOCATION SrcLoc(DXSrc.Buffer.Get(), Footprint);
-    CB.CmdList->CopyTextureRegion(&DstLoc, 0, 0, 0, &SrcLoc, nullptr);
+    const D3D12_RESOURCE_DESC TexDesc = DXDst.Resource->GetDesc();
+    const uint32_t NumSubresources = TexDesc.MipLevels;
+    llvm::SmallVector<D3D12_PLACED_SUBRESOURCE_FOOTPRINT> Layouts(
+        NumSubresources);
+    ComPtr<ID3D12DeviceX> Device;
+    DXDst.Resource->GetDevice(IID_PPV_ARGS(&Device));
+    Device->GetCopyableFootprints(&TexDesc, 0, NumSubresources, 0,
+                                  Layouts.data(), nullptr, nullptr, nullptr);
+    for (uint32_t Sub = 0; Sub < NumSubresources; ++Sub) {
+      const CD3DX12_TEXTURE_COPY_LOCATION DstLoc(DXDst.Resource.Get(), Sub);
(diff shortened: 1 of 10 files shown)


test/Feature/Textures/Texture2D.GetDimensions.test.yaml
```diff
@@ -114,8 +114,8 @@ Results:
 ...
 #--- end
 
-# Unimplemented: Clang + DX: https://github.com/llvm/llvm-project/issues/101558
-# XFAIL: DirectX || Metal
+# Unimplemented: Metal: https://github.com/llvm/llvm-project/issues/101558
+# XFAIL: Metal
```
MarijnS95, line 118: This re-points the same issue (llvm-project#101558) from "Clang + DX" to "Metal". If it's a clang frontend issue it should hit DX and Metal equally, so DX presumably now passes for a different reason (the `MipLevels` propagation) and the link was mis-attributed. Can you confirm DX CI is green here and fix the comment's rationale rather than re-aiming the same link at Metal?

lib/API/DX/Device.cpp
```diff
@@ -2049,6 +2054,43 @@ class DXDevice : public offloadtest::Device {
     return getAlignedTexturePitch(Desc.Width, getFormatSizeInBytes(Desc.Fmt));
   }
 
+  TextureUploadLayout
+  getTextureUploadLayout(const TextureCreateDesc &Desc) const override {
+    // Only the fields GetCopyableFootprints consults are needed here; layout,
+    // flags, and clear value do not affect the copyable footprint.
+    D3D12_RESOURCE_DESC TexDesc = {};
```
MarijnS95, line 2061: Retracting this — it's not a real concern, and the comment right above already says why: the copy buffer is row-major, so the texture's `Layout`/swizzle doesn't affect its copyable footprint. That holds whether the real resource is `UNKNOWN` or `64KB_UNDEFINED_SWIZZLE` (single-plane color, undefined swizzle), so the two `GetCopyableFootprints` calls can't diverge. The docs' "uses texture layouts" caveat is about cases that don't apply here (ROW_MAJOR textures, planar depth/stencil). Sorry for the noise.

lib/API/Texture.cpp
```diff
@@ -10,3 +12,24 @@ offloadtest::Texture::calculateLinearSizeInBytes(const Device &Dev) const {
...
+  const uint32_t ElementSize = getFormatSizeInBytes(Desc.Fmt);
+  TextureUploadLayout Layout;
+  Layout.Subresources.reserve(Desc.MipLevels);
+  uint64_t Offset = 0;
+  for (uint32_t I = 0; I < Desc.MipLevels; ++I) {
+    const uint32_t MipWidth = std::max(1u, Desc.Width >> I);
+    const uint32_t MipHeight = std::max(1u, Desc.Height >> I);
+    SubresourceFootprint Sub;
+    Sub.Offset = Offset;
+    Sub.RowSizeInBytes = MipWidth * ElementSize;
```
MarijnS95, line 27: This assumes uncompressed formats: `MipWidth * ElementSize` and mips shrinking to 1x1 texel are wrong for block-compressed formats (min 4x4 block). DX is fine via `GetCopyableFootprints`; VK/Metal would not be. No BC formats in the suite today, so just worth a note:

```suggestion
    // Assumes uncompressed formats; block-compressed formats would need
    // block-based row sizing rather than MipWidth * ElementSize.
    Sub.RowSizeInBytes = MipWidth * ElementSize;
```

MarijnS95 (approved):
LGTM with a few nits inline. Mip math is consistent across backends and the new seam is clean.

MarijnS95:
Is this now a duplicate of https://github.com/llvm/offload-test-suite/pull/1230, which we can subsequently close as fixed?

EmilioLaiso (approved):
Amazing! 😄

bogner (approved):

PR #1335, opened 2026-06-25
Added WaveOps tests for QuadReadLaneAt

### LINKED ISSUE #884: Add test for `QuadReadLaneAt`
- [ ] Test for 16 bit int types (`QuadReadLaneAt.int16.test`)
- [ ] Test for Half type (`QuadReadLaneAt.fp16.test`)
- [ ] Test for 32 bit types (`QuadReadLaneAt.32.test`)
- [ ] Test for 64 bit int types (`QuadReadLaneAt.int64.test`)
- [ ] Test for Double type (`QuadReadLaneAt.fp64.test`)

A .test file should contain 
```
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# RUN: split-file %s %t
```

kcloudy0717:
Resolves https://github.com/llvm/offload-test-suite/issues/884.

Note there is a separate testing path between DXIL and SPIRV due to the lane index restriction between the two IRs. DXIL allows lane index to be non-uniform as described in https://github.com/microsoft/DirectXShaderCompiler/wiki/Wave-Intrinsics, whereas SPIRV only allows constant (dynamically uniform after version 1.5, https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpGroupNonUniformQuadBroadcast).

### DIFF
--- a/test/WaveOps/QuadReadAcrossDiagonal.int64.test
+++ b/test/WaveOps/QuadReadAcrossDiagonal.int64.test
@@ -232,7 +232,7 @@ DescriptorSets:
 
 # REQUIRES: Int64
 
-# Bug: https://github.com/llvm/offload-test-suite/issues/988
+# Bug: https://github.com/llvm/offload-test-suite/issues/959
 # XFAIL: Metal
 
 # RUN: split-file %s %t
--- a/test/WaveOps/QuadReadAcrossX.int64.test
+++ b/test/WaveOps/QuadReadAcrossX.int64.test
@@ -232,8 +232,7 @@ DescriptorSets:
 
 # REQUIRES: Int64
 
-# Bug: https://github.com/llvm/offload-test-suite/issues/988
-# Bug: https://github.com/llvm/offload-test-suite/issues/989
+# Bug: https://github.com/llvm/offload-test-suite/issues/959
 # XFAIL: Metal
 
 # RUN: split-file %s %t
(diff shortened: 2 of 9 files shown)


bob80905 (approved):
LGTM

inbelic (approved):
LGTM, are we able to close the issues that are no longer linked as duplicate now?

kcloudy0717:
> LGTM, are we able to close the issues that are no longer linked as duplicate now?

We should be able to, I'll go through the list here and close them now that we have a mega-issue for all of the unsupported 64bit ops.

PR #1336, opened 2026-06-25
[SplitBuild] Remove persistent install / dxc-dist directories

### LINKED ISSUE #1338: [Split Build] Stale / non-existent tests are being run on PR's based on irrelevant branches
From @bogner's comment:

The offload test suite seems to be failing to clean up the working directory of the builders and leaving tests from other PRs around. 
For example, [Added WaveOps tests for QuadReadLaneAt by kcloudy0717 · Pull Request #1335 · llvm/offload-test-suite](https://github.com/llvm/offload-test-suite/pull/1335) is failing on InlineRT/geometry-transform ([Added WaveOps tests for QuadReadLaneAt · llvm/offload-test-suite@0a21459](https://github.com/llvm/offload-test-suite/actions/runs/28151333936/job/83370715079?pr=1335)), but that's introduced in [Add per-geometry Transform on TriangleGeometryDesc by MarijnS95 · Pull Request #1288 · llvm/offload-…](https://github.com/llvm/offload-test-suite/pull/1288) which is not merged.

We need to fix the infrastructure to make sure to start fresh and delete any pre-existing tests that were extracted from any previous artifacts.

bob80905:
An issue is occurring with tests from other branches being run on PRs with branches that don't have those tests defined.
The cause is that the split build feature does not reset its test directories, just overlaying what was extracted instead.
This PR makes sure to delete any existing extraction sites from previous extractions, making sure what will be extracted is strictly from the current branch

Assisted by: Github Copilot
Fixes https://github.com/llvm/offload-test-suite/issues/1338

### DIFF
--- a/.github/workflows/build-and-test-callable.yaml
+++ b/.github/workflows/build-and-test-callable.yaml
@@ -360,11 +360,6 @@ jobs:
         run: |
             set -euxo pipefail
             cd $GITHUB_WORKSPACE
-            # Self-hosted test runners persist $GITHUB_WORKSPACE between runs.
-            # `tar xf` overlays onto whatever the previous run left behind and
-            # never removes orphan files, so stale tests from an earlier PR
-            # extracted here would survive and be picked up by lit. Wipe the
-            # prefixes first so we run exactly the tests in this PR's tarball.
             rm -rf install dxc-dist
             mkdir -p install dxc-dist
             # Pipe the archive in via stdin and use a relative -C path so
--- a/.github/workflows/build-and-test-callable.yaml
+++ b/.github/workflows/build-and-test-callable.yaml
@@ -266,6 +266,15 @@ jobs:
         shell: bash
         run: |
             set -euxo pipefail
+            # Self-hosted runners persist $GITHUB_WORKSPACE between runs, and
+            # neither `cmake --install` nor `install(DIRECTORY ...)` deletes
+            # files that no longer exist in the source tree -- they only
+            # overlay. Without wiping the prefix first, stale test sources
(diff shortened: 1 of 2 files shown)


bogner (approved):

.github/workflows/build-and-test-callable.yaml
```diff
@@ -266,6 +266,15 @@ jobs:
         shell: bash
         run: |
             set -euxo pipefail
+            # Self-hosted runners persist $GITHUB_WORKSPACE between runs, and
```
damyanp: This comment seems excessively verbose to me, and only applies to the `rm -rf` line.  To me this would be easier to read and maintain if the comment were removed.  I also suggest you instruct copilot not to add verbose comments.  For example, I have stored in the ghcp memory "Keep code comments terse and human-style; avoid verbose, over-explaining "AI-sounding" comments."

.github/workflows/build-and-test-callable.yaml
```diff
@@ -358,6 +360,12 @@ jobs:
         run: |
             set -euxo pipefail
             cd $GITHUB_WORKSPACE
+            # Self-hosted test runners persist $GITHUB_WORKSPACE between runs.
```
damyanp: and here

I maintain though that the comment is just noise and should be removed.

bob80905: Yep, I'll remove this one entirely.
I think it would be helpful to have at least one terse comment in the source to explain what's going on with the removal though, so I think I should leave the short comment at the other site.

PR #1346, opened 2026-06-30
[Metal] Make dispatchRays a self-contained encoder operation

MarijnS95:
The DX and Vulkan backends record a ray dispatch by creating an encoder and calling `Encoder.dispatchRays()` directly. Metal instead routed through `createRayTracingCommands`, an `MTLDevice` method that bound the descriptor heap and top-level argument buffer, synthesized and bound the `IRDispatchRaysArgument`, marked the RT resources resident, and only then called a thin `dispatchRays()`.

Since Metal lowers raygen to a compute kernel, ray tracing now records through the shared `createComputeCommands` path — which already binds the heap and argument buffer and marks the common resources resident — by branching on `P.isRayTracing()`. The `IRDispatchRaysArgument` construction, upload, binding, and RT-specific residency move into `MTLComputeEncoder::dispatchRays`, defined out-of-line so it can allocate via `CB->Dev` and keep the buffer alive in `CB->KeepAliveOwned`. `IRDispatchRaysArgument::ResDescHeap` needs the resource heap address, which isn't in the cross-backend `dispatchRays` signature, so the encoder remembers the bound heap via `bindResourceHeap()`. `createRayTracingCommands` is removed.

No behavioral change. `check-hlsl-mtl-feature-rt` and `check-hlsl-mtl-basic` pass.

### DIFF
--- a/lib/API/MTL/MTLDevice.cpp
+++ b/lib/API/MTL/MTLDevice.cpp
@@ -609,6 +609,9 @@ class MTLComputeEncoder : public offloadtest::ComputeEncoder {
   /// the next encoder transition (via endEncodingImpl).
   MTL::AccelerationStructureCommandEncoder *ASEnc = nullptr;
 
+  /// Bound resource heap, read back for IRDispatchRaysArgument::ResDescHeap.
+  MTLDescriptorHeap *BoundResourceHeap = nullptr;
+
   /// Accumulated barrier scope from commands recorded since the last barrier.
   MTL::BarrierScope PendingScope = MTL::BarrierScope(0);
 
@@ -664,6 +667,14 @@ public:
   }
 
   MTL::ComputeCommandEncoder *getNative() const { return ComputeEnc; }
+
+  llvm::Error bindResourceHeap(MTLDescriptorHeap &Heap) {
+    if (auto Err = ensureComputeEncoder())
+      return Err;
+    Heap.bind(ComputeEnc);
+    BoundResourceHeap = &Heap;
+    return llvm::Error::success();
+  }
 
   MTL::CommandEncoder *getActiveEncoder() const {
     if (ComputeEnc)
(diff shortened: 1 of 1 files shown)


lib/API/MTL/MTLDevice.cpp
```diff
@@ -787,40 +801,17 @@ class MTLComputeEncoder : public offloadtest::ComputeEncoder {
...
+  // compute kernel. The caller must have already bound the descriptor heap and
+  // top-level argument buffer (the compute-dispatch path does the same).
+  // Defined out-of-line below — allocating the argument buffer needs
+  // MTLDevice's full type.
+  llvm::Error dispatchRays(const PipelineState &PSO,
+                           const ShaderBindingTable &SBT, uint32_t Width,
+                           uint32_t Height, uint32_t Depth) override;
 
   /// Lazily transition into an AccelerationStructureCommandEncoder; mirrors
   /// the existing compute↔blit lazy switch.
```
manon-traverse, line 806: Do we need this claude comment?

lib/API/MTL/MTLDevice.cpp
```diff
@@ -1772,7 +1764,18 @@ class MTLDevice : public offloadtest::Device {
       NativeEncoder->useResource(llvm::cast<MTLBuffer>(B.get())->Buf,
                                  MTL::ResourceUsageRead);
 
-    if (auto Err = Encoder.dispatch(*IS.Pipeline.get(),
+    // Metal compiles raygen as a compute kernel, so a ray-tracing pipeline
+    // dispatches through the same compute encoder as a plain compute shader;
+    // dispatchRays adds the RT-specific argument buffer and residency.
```
manon-traverse: pls remove useless comments

lib/API/MTL/MTLDevice.cpp
```diff
@@ -3094,7 +2991,9 @@ class MTLDevice : public offloadtest::Device {
       IS.SBT = std::move(*SBTOrErr);
       llvm::outs() << "Shader Binding Table created.\n";
 
-      if (auto Err = createRayTracingCommands(P, IS))
+      // Metal lowers raygen to a compute kernel, so ray tracing records its
+      // dispatch through the shared compute-command path.
```
manon-traverse: pls remove this useless comment.

lib/API/MTL/MTLDevice.cpp
```diff
@@ -787,40 +801,17 @@ class MTLComputeEncoder : public offloadtest::ComputeEncoder {
...
-    return llvm::Error::success();
-  }
+  // Issue a ray dispatch: synthesizes the per-dispatch IRDispatchRaysArgument
+  // (SBT region addresses, GRS + resource-heap pointers, visible/intersection
+  // function-table IDs), binds it at kIRRayDispatchArgumentsBindPoint, marks
+  // the RT-specific resources resident, and dispatches the irconverter raygen
+  // compute kernel. The caller must have already bound the descriptor heap and
+  // top-level argument buffer (the compute-dispatch path does the same).
+  // Defined out-of-line below — allocating the argument buffer needs
+  // MTLDevice's full type.
```
manon-traverse: We shouldn't be needing any of these comments

manon-traverse (approved):
Please just remove the useless claude comments before merning

MarijnS95:
In hindsight I do like the terseness of the original dependent un-rebased commit https://github.com/llvm/offload-test-suite/commit/ae763a1b531f08032c6d1303ddf4b936bf4bbc1b a lot more than what ended up here; will align the two a bit better.

MarijnS95:
Stripped the chatty comments and condensed the remaining ones to terse one-liners (force-pushed). I also double-checked the `dispatchThreads` claim against Apple's docs: since macOS 10.13 it uses non-uniform threadgroup sizes, dispatching exactly Width×Height×Depth threads and sizing edge threadgroups down, so the raygen kernel's per-ray index never exceeds the grid — hence the bounds-checking note. Thanks for the review!

manon-traverse (approved):

EmilioLaiso (approved):

PR #1347, opened 2026-06-30
Add InterlockedMin tests

### LINKED ISSUE #104: Add test for InterlockedMin
In Test\Feature\Intrinsics create, if applicable: (reference the HLSL headers for the types supported by the intrinsic)

A .test file should contain 
```
@@ -0,0 +1,54 @@
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# UNSUPPORTED: Clang-Vulkan
# RUN: split-file %s %t
```

bob80905:
This PR adds tests for InterlockedMin
Adds 32 bit and 64 bit signed/unsigned integer tests, along with some resource member function tests.
Fixes https://github.com/llvm/offload-test-suite/issues/104
Assisted by: Github Copilot

### DIFF
--- a/test/Feature/HLSLLib/InterlockedMin.32.test
+++ b/test/Feature/HLSLLib/InterlockedMin.32.test
@@ -39,7 +39,7 @@ groupshared int  MinIntNoOrig;    // 2-arg form, signed
 groupshared uint MinUIntNoOrig;   // 2-arg form, unsigned
 groupshared int  SignedCounter;   // 2-arg form reaching a negative minimum
 
-[numthreads(64, 1, 1)]
+[numthreads(256, 1, 1)]
 void main(uint3 GTID : SV_GroupThreadID) {
   if (GTID.x == 0) {
     MinInt = 1000;
@@ -63,7 +63,7 @@ void main(uint3 GTID : SV_GroupThreadID) {
   uint afterU = MinUInt;
   OutMonoUInt[GTID.x] = (afterU <= origU) ? 1u : 0u;
 
-  // 2-argument form. Proposals span 0..-63, min = -63.
+  // 2-argument form. Proposals span 0..-255, min = -255.
   InterlockedMin(MinIntNoOrig, -(int)GTID.x);
   // 2-argument form, unsigned. Proposals span 100..163, min = 100.
   InterlockedMin(MinUIntNoOrig, GTID.x + 100u);
@@ -75,7 +75,7 @@ void main(uint3 GTID : SV_GroupThreadID) {
   if (GTID.x == 0) {
     OutFinal[0] = MinInt;             // -32
     OutFinal[1] = (int)MinUInt;       // 0
-    OutFinal[2] = MinIntNoOrig;       // -63
+    OutFinal[2] = MinIntNoOrig;       // -255
     OutFinal[3] = (int)MinUIntNoOrig; // 100
     OutFinal[4] = SignedCounter;      // -1000
   }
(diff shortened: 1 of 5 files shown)


Icohedron (approved):

test/Feature/HLSLLib/InterlockedMin.32.test
```diff
@@ -0,0 +1,165 @@
+#--- source.hlsl
+
+// This test exercises InterlockedMin against non-resource (groupshared)
+// destinations. A single threadgroup of 64 threads concurrently updates
```
damyanp, line 4: Is 64 really enough?  Presumably we really a threadgroup size that's larger than the hardware's lane count - so I think at least 256?

test/Feature/HLSLLib/InterlockedMin.32.test
```diff
@@ -0,0 +1,165 @@
...
+//
+// Both the 2-argument and 3-argument overloads are covered for int and uint.
+//
+// Atomicity is verified for these interlockedmin intrinsics
+// two independent ways:
+//
+//   1. Final value. Each counter's deterministic final value is the minimum
+//      over all participating threads (and the initial value). A non-atomic
+//      read-modify-write can lose the lowest write: a thread reads the
+//      pre-minimum value, then stores its larger proposal *after* the
```
damyanp, line 15: Where did the term "proposal" come from in this context?  I can't find any other references calling this a proposal (checks HLSL docs, the spec, slang docs).

test/Feature/HLSLLib/InterlockedMin.32.test
bob80905: Is 64 really enough to create the environment for a race condition? Is that your question? I would say yes, and I am unsure what the benefit would be for more threads i.e., 256.

The expected buffers would quadruple in size, and we have somewhat of a precedent for previous interlocked functions to use 64 threads.

test/Feature/HLSLLib/InterlockedMin.32.test
bob80905: The term did come from github copilot, but I think it makes sense in this context. A new minimum is constantly being proposed for each call.

test/Feature/HLSLLib/InterlockedMin.32.test
damyanp: You really want to be testing a race condition across waves.  On many GPUs, 64 is two waves.  But on some, 64 is only half a wave.  So 256 would guarantee you have at least 2 waves involved in the test.

Icohedron: I'm curious what commerically-available GPUs have a wave size of 128 🤔

damyanp: From https://d3d12infodb.boolka.dev/FeatureTable.html:

image

test/Feature/HLSLLib/InterlockedMin.32.test
```diff
@@ -0,0 +1,189 @@
+#--- source.hlsl
+
+// This test exercises InterlockedMin against non-resource (groupshared)
+// destinations. A single threadgroup of 64 threads concurrently updates
```
inbelic, line 4:
```suggestion
// destinations. A single threadgroup of 256 threads concurrently updates
```

nit: forgot to update this, maybe we can ammend in the max pr

PR #1348, opened 2026-06-30
Add a few tests for ConstantBuffer<T>

### LINKED ISSUE #200: [HLSL] Survey current `ConstantBuffer<T>` support in DXC and add tests to offload-test-suite
Document what can be done with `ConstantBuffer<T>` in DXC (local variables, arrays, ...).
Add a set of test cases for this to offload test suite.

```
struct S {
  int a;
  float4 b;
};

RWBuffer<float> Buf;

ConstantBuffer<S> CBS[2];

[numthreads(4,1,1)]
void main(int i : SV_GROUPID) {
    ConstantBuffer<S> cb = CBS[i % 2];
    Buf[0] = cb.b.x;
}
```
https://godbolt.org/z/M8TjExTnq

bogner:
A lot of these are fairly similar to the CBuffers tests, but use the resource type `ConstantBuffer<>` instead of the `cbuffer` construct. I also updated a couple of XFAILs that are failing in the same way between the two constructs, as they were out of date.

Fixes llvm/wg-hlsl#200

### DIFF
--- a/test/Feature/CBuffer/structs.test
+++ b/test/Feature/CBuffer/structs.test
@@ -97,7 +97,7 @@ DescriptorSets:
 ...
 #--- end
 
-# Unimplemented https://github.com/llvm/llvm-project/issues/186465
+# Bug: https://github.com/llvm/llvm-project/issues/180600
 # XFAIL: Clang && Vulkan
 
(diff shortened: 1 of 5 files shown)


hekota (approved):
LGTM!

damyanp (approved):

PR #1352, opened 2026-07-01
OffloadMigration: Fix `-Wchanges-meaning` errors from newer GCC

MarijnS95:
Simply shorten the field names from `Buffer` -> `Buf` and `Texture` -> `Tex` so that they no longer conflict with the same type name, allowing the framework to be compiled once again.  A similar issue was fixed earlier in #1095.

### DIFF
--- a/lib/Support/OffloadMigration.cpp
+++ b/lib/Support/OffloadMigration.cpp
@@ -3,6 +3,8 @@
 #include "API/Device.h"
 #include "API/FormatConversion.h"
 
+#include "llvm/ADT/STLForwardCompat.h"
+
 namespace offloadtest {
 
 static BufferUsage bufferUsageFromResourceKind(ResourceKind Kind) {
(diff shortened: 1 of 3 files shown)


lib/Support/OffloadMigration.cpp
```diff
@@ -37,7 +37,8 @@ static BufferShaderAccessType bufferShaderAccessTypeFromResourceKind(
         toFormat(Resource.BufferPtr->Format, Resource.BufferPtr->Channels);
     if (!FmtOrErr) {
       printf("Invalid format! FMT: %d, CHANNELS: %d\n",
-             Resource.BufferPtr->Format, Resource.BufferPtr->Channels);
+             static_cast<int>(Resource.BufferPtr->Format),
```
bogner: `llvm::to_underlying(Resource.BufferPtr->Format)` is probably better here (similar to `std::to_underlying`, but we aren't on a new enough C++ version for that quite yet)

bogner (approved):

bob80905 (approved):

PR #1359, opened 2026-07-07
Adding lavapipe support

joaosaffran:
This PR adds lavapipe support as a testing driver. Such includes a few changes: adding new pipelines to run lavapipe on amd and qc runners; adding `check-hlsl-vk-lavapipe` and `check-hlsl-clang-vk-lavapipe` to force lavapipe as test driver; updating docs to show lavapipe badges


utils/configure-test-suite.py
```diff
@@ -29,27 +29,53 @@
 import shutil
 import sys
 
-# Suite name -> (d3d12, vk, mtl, clang, warp, description).
+# Suite name -> (d3d12, vk, mtl, clang, warp, lavapipe, description).
 SUITES = {
-    "d3d12": (True, False, False, False, False, "DXC-compiled shaders on DirectX 12"),
-    "vk": (False, True, False, False, False, "DXC-compiled shaders on Vulkan"),
-    "mtl": (False, False, True, False, False, "DXC-compiled shaders on Metal"),
+    "d3d12": (
```
bob80905: If you can undo this change and clang-format doesn't complain, that would be a win

.github/workflows/build-and-test-callable.yaml
```diff
@@ -221,6 +221,53 @@ jobs:
             cd llvm-project
             cd build
             ./bin/api-query
+      - name: Setup Lavapipe
+        if: inputs.SplitBuild != true && contains(inputs.TestTarget, 'lavapipe')
```
bob80905, line 225: I am slightly curious about this, and how lavapipe works:
SplitBuild allows certain machines to build compiler binaries, and uploads the artifacts to a server, so that other machines can pull them down and run tests.
Lavapipe, as far as I understand, is for replicating linux environments / testing on linux.
Do you expect the binaries to be compatible / executable within the lavapipe context? (If built by another machine)

joaosaffran: lavapipe is warp for vulkan. It allows us to check spirv generated code without running in the GPU, we use a CPU to emulate its behavior. This shouldn't be an issue, since all our builds are windows, so we are building and testing on windows, so the binaries should be fine.

utils/configure-test-suite.py
bob80905: Unfortunate, the formatter does complain for some reason :(

.github/workflows/build-and-test-callable.yaml
```diff
@@ -456,6 +503,53 @@ jobs:
               --suite "${{ steps.suite.outputs.suite }}" \
               --dxc-path "$dxc_path" \
               "${warp_arch_args[@]}"
+      - name: Setup Lavapipe
+        if: contains(inputs.TestTarget, 'lavapipe')
+        shell: bash
+        run: |
+            set -euxo pipefail
+            # Lavapipe (Mesa's software Vulkan rasterizer) provides a
```
bob80905: Maybe refer back to the previous copy of this comment instead of verbosely repeating it?

test/lit.cfg.py
```diff
@@ -340,18 +372,62 @@ def setDeviceFeatures(config, device, compiler):
...
+# be llvmpipe. Pin it to the same hardware device lit selected above. This
+# only kicks in when llvmpipe is actually present and no explicit GPU name
+# filter is already in effect, so single-Vulkan-device runners are unaffected.
+vulkan_has_lavapipe = any(
+    d.get("API") == "Vulkan" and "llvmpipe" in d.get("Description", "").lower()
+    for d in devices.get("Devices", [])
+)
+if (
+    not config.offloadtest_test_lavapipe
+    and not ShouldSearchByGPuName
```
bob80905: A nit, but while you're here, could you fix this var name and go with ShouldSearchByGPUName instead?

bob80905 (approved):
I just have nits, take em or leave em, LGTM

farzonl:
hmm I have a feeling this could have been easier if we took advantage of `OFFLOADTEST_GPU_NAME` since that is how I am running lavapipe changes today.

utils/configure-test-suite.py
```diff
@@ -232,7 +277,7 @@ def main():
         golden_dir = str(candidate.resolve()) if candidate.is_dir() else ""
 
     for suite in args.suites:
-        d3d12, vk, mtl, clang, warp, _ = SUITES[suite]
+        d3d12, vk, mtl, clang, warp, lavapipe, _ = SUITES[suite]
```
farzonl: warp I feel should be the one exception for cpu renderers because vulkan has so many like SwiftShader and kazan. I think this would be far less code if  you used the `OFFLOADTEST_GPU_NAME`.  You can still add the cmake just lets not do the python this way.

test/CMakeLists.txt
```diff
@@ -193,6 +220,24 @@ if (OFFLOADTEST_ENABLE_D3D12 AND WIN32)
...
+  # Lavapipe reuses the plain Vulkan suite config; device selection happens via
+  # OFFLOADTEST_GPU_NAME=llvmpipe, which add_offloadtest_lavapipe_suite exports
+  # into lit's environment. No lavapipe-specific logic lives in the Python
+  # test config.
+  add_offloadtest_lavapipe_suite(vk-lavapipe)
+
+  if (OFFLOADTEST_TEST_CLANG)
+    set(FORCE_CLANG True)
+    add_offloadtest_lavapipe_suite(clang-vk-lavapipe)
+    set(FORCE_CLANG False)
```
farzonl: why do we toggle `FORCE_CLANG` on before calling `add_offloadtest_lavapipe_suite` and then turn it off? It doesn't looking like we use FORCE_CLANG anywhere in this pr other than to set it and unset it.

test/CMakeLists.txt
```diff
@@ -193,6 +220,24 @@ if (OFFLOADTEST_ENABLE_D3D12 AND WIN32)
...
+  set(FORCE_WARP False)
+endif()
+
+if (SUPPORTS_SPIRV AND OFFLOADTEST_ENABLE_VULKAN AND WIN32)
+  set(TEST_vk True)
+  # Lavapipe reuses the plain Vulkan suite config; device selection happens via
+  # OFFLOADTEST_GPU_NAME=llvmpipe, which add_offloadtest_lavapipe_suite exports
+  # into lit's environment. No lavapipe-specific logic lives in the Python
+  # test config.
+  add_offloadtest_lavapipe_suite(vk-lavapipe)
```
farzonl: Is it possible to use `api-query` to know if `lavapipe` is available? feels like we are setting it  always for vulkan which I think is true for linux, but has to be installed for Windows. Feels like it would be safer if we could check if the target is available before making it something cmake adds.

test/CMakeLists.txt
```diff
@@ -146,6 +146,33 @@ function(add_offloadtest_lit_suite suite)
   )
 endfunction()
 
+
+function(add_offloadtest_lavapipe_suite suite)
```
farzonl: this function has given me an idea on how we can make `OFFLOADTEST_GPU_NAME` more useful.  If this function were  more generic  then we can support `-E env OFFLOADTEST_GPU_NAME=${gpu_name}` that would let us support the creation of a `clang-vk-nvidia`, `clang-vk-amd`, and `clang-vk-intel`. if we combine this with a call to api-query then we can generate all the supported vulkan targets in cmake via their description or vendor name. I dunno just an idea maybe we defer it for a future change.

test/lit.cfg.py
```diff
@@ -352,6 +366,14 @@ def setDeviceFeatures(config, device, compiler):
     config.fatal("No target device found!")
 setDeviceFeatures(config, target_device, HLSLCompiler)
 
+# Lavapipe (Mesa's llvmpipe software Vulkan device) has known limitations
+# compared to hardware drivers -- notably its LLVM JIT can hit access
+# violations on some shaders. Expose a "Lavapipe" feature so affected tests can
+# be scoped with "# XFAIL: Lavapipe" / "# UNSUPPORTED: Lavapipe" without
+# affecting other backends.
+if "llvmpipe" in target_device.get("Description", "").lower():
+    config.available_features.add("Lavapipe")
```
farzonl, line 370: you added this but there are no tests that use it.

joaosaffran: I have a follow-up PR that adds that to some tests, since this PR is introducing the lavapipe infrastructure, I think it makes sense to put it in here

farzonl (approved):

docs/CI.md
```diff
@@ -65,6 +67,7 @@ Configurations:
 * Configurations:
   * Windows AMD GPU DirectX
   * Windows AMD GPU Vulkan
+  * Windows Lavapipe Vulkan (x64)
```
bob80905, line 70: Why is this only added to amd and not intel or nvidia? Is there a rationale for selecting only amd?

farzonl: Because we only need one x86 machine to test lavapipe. I assume the AMD gpu machine is the fastest or most available.

PR #1362, opened 2026-07-10
XFAIL lavapipe tests

joaosaffran:
This patch is XFAIL and UNSUPPORTED for some tests that have limited coverage or are currently failing on lavapipe. All of those were triaged, and bugs were filed to keep track of them.


test/Feature/HLSLLib/fma.test
```diff
@@ -56,8 +56,8 @@ Buffers:
     # last two rows: 2^-26, -(2^-26), 2^-104, -3
 Results:
   - Result: Result
-    Rule: BufferFloatULP
-    ULPT: 1
+    Rule: BufferFloatEpsilon 
+    Epsilon: 0.0001
```
llvm-beanz, line 60: Do you happen to know if the values impacted here were very close to 0?

lib/API/VK/Device.cpp
```diff
@@ -5115,9 +5115,9 @@ VulkanDevice::createShaderBindingTable(const PipelineState &PSO,
   const auto &VKPSO = llvm::cast<VKRayTracingPipelineState>(PSO);
 
   const uint32_t HandleSize = RTPipelineProps.shaderGroupHandleSize;
-  const SBTLayout Layout =
-      computeSBTLayout(HandleSize, RTPipelineProps.shaderGroupHandleAlignment,
-                       RTPipelineProps.shaderGroupBaseAlignment, Desc);
+  const uint32_t BaseAlign = RTPipelineProps.shaderGroupBaseAlignment;
+  const SBTLayout Layout = computeSBTLayout(
+      HandleSize, RTPipelineProps.shaderGroupHandleAlignment, BaseAlign, Desc);
```
llvm-beanz: Please put the functional change into its own PR separate from the test updates.

That allows us to have a clear history and not lose context that many of these test updates aren't actually associated with this change at all.

test/Feature/HLSLLib/fma.test
joaosaffran: Yeah, this is the error messages I was seeing before this fix:

```
Mismatch at element 10: expected=4.93038e-32 (0x1.0000000000000p-104), actual=0 (0x0.0000000000000p+0)
```

Not sure how precise we want this to be, though. I am open to be educated as the best approach for such precise tests

llvm-beanz: Thanks! I think this is fine. I have a local patch working around similar issues, so I may propose an alternate solution in the future.

llvm-beanz (approved):

alsepkow (approved):

PR #1371, opened 2026-07-14
Move Lavapipe targets to experimental tier

kmpeng:
Lavapipe targets were added as tier 1 targets in #1359, but since they're not completely passing yet, we should move them to the experimental tier.


.github/workflows/pr-matrix.yaml
```diff
@@ -129,7 +112,10 @@ jobs:
       fail-fast: false
       matrix:
         SKU: [windows-amd, windows-nvidia, windows-qc]
-        TestTarget: [check-hlsl-vk, check-hlsl-clang-vk]
+        TestTarget: [check-hlsl-vk, check-hlsl-clang-vk, check-hlsl-vk-lavapipe, check-hlsl-clang-vk-lavapipe]
```
kmpeng, line 115: I added Lavapipe to `test-vk-all` in addition to `test-all`, not sure if that's right

Icohedron (approved):

.github/workflows/pr-matrix.yaml
```diff
@@ -82,10 +63,12 @@ jobs:
       fail-fast: false
       matrix:
         SKU: [windows-amd, windows-nvidia, windows-qc]
-        TestTarget: [check-hlsl-d3d12, check-hlsl-vk, check-hlsl-clang-d3d12, check-hlsl-clang-vk]
+        TestTarget: [check-hlsl-d3d12, check-hlsl-vk, check-hlsl-clang-d3d12, check-hlsl-clang-vk, check-hlsl-vk-lavapipe, check-hlsl-clang-vk-lavapipe]
         exclude:
           - { SKU: windows-nvidia, TestTarget: check-hlsl-d3d12 }
           - { SKU: windows-nvidia, TestTarget: check-hlsl-clang-d3d12 }
+          - { SKU: windows-nvidia, TestTarget: check-hlsl-vk-lavapipe }
```
bob80905, line 70: To be clear, I think this means we're excluding NVIDIA lavapipe targets from the extra / experimental tier.
Which means they work well. Is that right?

kmpeng: We are "excluding" them from the experimental tier because there's no nvidia lavapipe config. Since the matrix is a SKU × TestTarget cross-product and windows-nvidia is needed for other targets, the excludes are just to drop the generated nvidia × lavapipe combos

bob80905 (approved):

.github/workflows/pr-matrix.yaml
farzonl: Ah that wasn't clear I thought we were excluding `windows-nvidia` because we only needed one x86 SKU covered by `windows-amd`  and one aarch64 sku covered by `windows-qc`.

PR #1372, opened 2026-07-14
Add `test-lavapipe`, `test-warp` and `disable-default-tests`

joaosaffran:
Add `test-lavapipe` and `test-warp` to only run lavapipe and warp tests


.github/workflows/pr-matrix.yaml
```diff
@@ -39,6 +39,12 @@ jobs:
     permissions:
       contents: read
       checks: write
+    if: >-
+      contains(github.event.pull_request.labels.*.name, 'test-all')
+      || contains(github.event.pull_request.labels.*.name, 'test-warp')
+      || (!contains(github.event.pull_request.labels.*.name, 'test-lavapipe')
+      && !contains(github.event.pull_request.labels.*.name, 'test-warp'))
```
Icohedron:
```suggestion
      && !contains(github.event.pull_request.labels.*.name, 'test-warp')
```

.github/workflows/pr-matrix.yaml
```diff
@@ -17,6 +17,10 @@ jobs:
     permissions:
       contents: read
       checks: write
+    if: >-
+      contains(github.event.pull_request.labels.*.name, 'test-all')
```
kmpeng: Is the condition `contains(github.event.pull_request.labels.*.name, 'test-all')` needed for the tier 1 targets + macos?

.github/workflows/pr-matrix.yaml
```diff
@@ -17,6 +17,10 @@ jobs:
...
+      || (!contains(github.event.pull_request.labels.*.name, 'test-lavapipe')
+      && !contains(github.event.pull_request.labels.*.name, 'test-warp'))
```
Icohedron: I find it odd that `test-warp` and `test-lavapipe` behave very differently from `test-dx-all` and `test-vk-all`.
`test-dx-all` and `test-vk-all` do not disable the default set of tests, while here you are configuring `test-warp` and `test-lavapipe` to disable them.

Furthermore, the way `test-warp` and `test-lavapipe` are handled will lead to even more difficult-to-read if-condition soup if we decide to add even more labels like it (e.g., `test-amd`, `test-nv`, etc.). 

I would prefer to have a label such as `disable-default-tests` to disable the default tests. Then have `test-warp` and `test-lavapipe` to selectively enable warp and lavapipe tests.

.github/workflows/pr-matrix.yaml
```diff
@@ -39,6 +43,11 @@ jobs:
     permissions:
       contents: read
       checks: write
+    if: >-
+      contains(github.event.pull_request.labels.*.name, 'test-all')
+      || contains(github.event.pull_request.labels.*.name, 'test-warp')
+      || (!contains(github.event.pull_request.labels.*.name, 'test-lavapipe')
+      && !contains(github.event.pull_request.labels.*.name, 'test-warp'))
```
bob80905: I think you can drop && !contains(github.event.pull_request.labels.*.name, 'test-warp')
If this is true, then the previous condition would be true, and the or would resolve to true.

.github/workflows/pr-matrix.yaml
```diff
@@ -79,13 +88,37 @@ jobs:
       SplitBuild: true
       LLVM-ExtraCMakeArgs: -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DOFFLOADTEST_USE_CLANG_TIDY=On
 
+  Exec-Tests-Extra-Lavapipe:
+    permissions:
+      contents: read
+      checks: write
+    if: >-
+      contains(github.event.pull_request.labels.*.name, 'test-all')
+      || contains(github.event.pull_request.labels.*.name, 'test-lavapipe')
```
kmpeng: `test-all` already runs Lavapipe - we don't want `Exec-Tests-Extra-Lavapipe` to run a duplicate
```suggestion
    if: >-
      contains(github.event.pull_request.labels.*.name, 'test-lavapipe')
      && !contains(github.event.pull_request.labels.*.name, 'test-all')
```

Icohedron (changes requested):

.github/workflows/pr-matrix.yaml
bob80905: Symmetrically referencing the WARP conditions, I think something is missing here.
Do you want lavapipe to run by default? If there are no labels, it looks like this would not run, and so by default lavapipe is turned off.
I expect the conditional structure for this set of tests to reflect or closely mirror that of the WARP tests above.

kmpeng (commented):
Did we decide these labels are just temporary? If they are we should probably write that in the title and/or description

.github/workflows/pr-matrix.yaml
Icohedron: No, the condition here is that the tests run if either
1. The `test-all` label is there, or
2. The labels `test-warp` and `test-lavapipe` are both not present.

This condition will get even bigger if we decide to make other `test-<gpu>` labels that disable default tests.
Thus, I [proposed](https://github.com/llvm/offload-test-suite/pull/1372#discussion_r3582725717) having a `disable-default-tests` label instead of this more difficult-to-read if condition.

.github/workflows/pr-matrix.yaml
kmpeng: We don't want any Lavapipe tests to run by default since they are experimental. Regarding `Exec-Tests-Extra-Lavapipe`, those should only run when the `test-lavapipe` label is present because they're already run in `Exec-Tests-Extra` (hence my suggestion [here](https://github.com/llvm/offload-test-suite/pull/1372#discussion_r3582737858))

.github/workflows/pr-matrix.yaml
kmpeng: Pretty sure that parentheses goes with the `(` that's before `!contains` in the previous line
```
|| (!contains(github.event.pull_request.labels.*.name, 'test-lavapipe')
      && !contains(github.event.pull_request.labels.*.name, 'test-warp'))
```

.github/workflows/pr-matrix.yaml
Icohedron: I guess we could also just change `test-dx-all` and `test-vk-all` to also disable the default set of tests.
Just as long as it's consistent and we can reduce the amount of complexity when it comes to managing all the if conditions so there isn't conflicting/differing logic between the `test-` labels.

Icohedron (approved):

Icohedron:
Change the PR title to mention the new `disable-default-tests` label

bob80905 (approved):

.github/workflows/pr-matrix.yaml
```diff
@@ -79,6 +83,28 @@ jobs:
       SplitBuild: true
       LLVM-ExtraCMakeArgs: -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DOFFLOADTEST_USE_CLANG_TIDY=On
 
+  Exec-Tests-Extra-Lavapipe:
+    permissions:
+      contents: read
+      checks: write
+    if: >-
+      contains(github.event.pull_request.labels.*.name, 'test-lavapipe')
+      || contains(github.event.pull_request.labels.*.name, 'test-all')
```
kmpeng: This still should never run with `test-all` because lavapipe is already being run in `Exec-Tests-Extra`
```suggestion
    if: >-
      contains(github.event.pull_request.labels.*.name, 'test-lavapipe')
      && !contains(github.event.pull_request.labels.*.name, 'test-all')
```

kmpeng (approved):

.github/workflows/pr-matrix.yaml
bob80905: Should we sub out intel for sw-rasterizer-x86_64 here too?

PR #1373, opened 2026-07-15
Allow ARM64 builds on any Windows runner

### LINKED ISSUE #1398: Enable Arm64 cross compilation for test runners
This issue tracks enabling Arm64 cross compilation for test runners.
Right now, only the Qualcomm test runner may run arm64 build and test jobs.
This places a bottleneck on the qualcomm test runner.
The issue here is to allow x64 machines to cross compile dxc / clang binaries for the Arm64 runner.
The Arm64 runner can offload its build responsibility to x64 machines, which we have plenty of, and still maintain responsibility for the test jobs.
This also sets up the longer term goal of bringing up a dedicated frequent build machine that builds once for all configs.
This new machine will use the arm64 cross compilation feature to also handle distributing binaries to the arm64 machine for testing.

damyanp:
Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

Copilot-Session: 94ff14fb-6023-4436-8382-29a19a4c54b0

Fixes https://github.com/llvm/offload-test-suite/issues/1398


llvm-beanz (commented):
I think this PR adds an insane amount of complexity for a build flow that probably isn't the right way to move our pipelines.

I think we should probably not merge this PR and instead we should re-evaluate how to move in a direction where a separate workflow builds and uploads artifacts for x64 and arm64 and the callable workflow should grab those artifacts.

bob80905:
Closing in favor of introducing frequent builder feature first, hooking it up with the other test machines over time, then cross compilation feature, and making sure it works on the frequent builder, before finally hooking the QC machine to it.

PR #1374, opened 2026-07-15
[Resource] Add tests for resource access in control flow

inbelic:
This pr adds test coverage for resource access patterns that are potential optimizations in modern llvm. These optimizations can create illegal looking resource accesses in clang. DXC does not represent resource access through a regular store/load and will perform any such optimizations.

Optimizations considered:
- `GVN` sink of load/store
- `SimplifyCFG` sink of load/store
- `InstCombine` phi sink of load
- `LICM` loop invariant sink

These are considered for [non-]dynamic resources and resources that result in a gep of a struct member.

Assisted by: Claude Opus 4.8


test/Feature/ResourceControlFlow/gvn-redundant-load.test
```diff
@@ -0,0 +1,80 @@
...
+    Format: Float32
+    FillSize: 32
+
+  - Name: ExpectedOut
+    Format: Float32
+    Data: [ 11, 11, 13, 13, 15, 15, 17, 17 ]
+
+Results:
+  - Result: BufOut
+    Rule: BufferFloatULP
```
bob80905: Could this be an exact match instead?

test/Feature/ResourceControlFlow/licm-no-hoist-load-array.test
```diff
@@ -0,0 +1,67 @@
+#--- source.hlsl
+
+// Dynamic resource: Bufs[I] is indexed by the loop induction variable, so each
+// iteration selects a different resource. LICM cannot hoist the load, and the
```
alsepkow, line 4: Co-pilot suggested the filename should be 'licm-**no**-hoist-load-array.test'.
I agree, Thanks co-pilot!

test/Feature/ResourceControlFlow/divergent-dynamic.test
```diff
@@ -0,0 +1,89 @@
...
+cbuffer Constants : register(b0) {
+  uint Condition;
+}
+
+[numthreads(8, 1, 1)]
+void main(uint TID : SV_DispatchThreadID) {
+  RWStructuredBuffer<float> LocalBuf = Bufs[0];
+
+  if ((TID & 1u) == Condition)
+    LocalBuf = Bufs[1];
```
bob80905: Might this be a scenario where we use `NonUniformResourceIndex` ?

test/Feature/ResourceControlFlow/gvn-redundant-load.test
inbelic: Good idea, changed the type to `uint` to make it easier and to match exactly

bob80905 (approved):

test/Feature/ResourceControlFlow/simplifycfg-hoist-load.test
```diff
@@ -0,0 +1,96 @@
+#--- source.hlsl
```
kmpeng: This test looks the exact same as `instcombine-sink-load.test`. Are they testing different things? If they are, how?

test/Feature/ResourceControlFlow/simplifycfg-hoist-load.test
```diff
@@ -0,0 +1,96 @@
...
+}
+
+[numthreads(8, 1, 1)]
+void main(uint TID : SV_DispatchThreadID) {
+  uint Val;
+  if ((TID & 1u) == Condition) {
+    Val = BufA[TID];
+  } else {
+    Val = BufB[TID];
+  }
```
kmpeng: nit
```suggestion
  if ((TID & 1u) == Condition)
    Val = BufA[TID];
  else
    Val = BufB[TID];
```

test/Feature/ResourceControlFlow/simplifycfg-sink-load-with-gep.test
```diff
@@ -0,0 +1,102 @@
...
+// the common successor, creating a phi of pointers into a resource.
+
+struct Foo {
+  uint A;
+  uint B;
+};
+
+[[vk::binding(1, 0)]] RWStructuredBuffer<Foo>   BufA : register(u0);
+[[vk::binding(2, 0)]] RWStructuredBuffer<Foo>   BufB : register(u1);
+[[vk::binding(3, 0)]] RWStructuredBuffer<uint> Out  : register(u2);
```
kmpeng: nit

```suggestion
[[vk::binding(1, 0)]] RWStructuredBuffer<Foo> BufA : register(u0);
[[vk::binding(2, 0)]] RWStructuredBuffer<Foo> BufB : register(u1);
[[vk::binding(3, 0)]] RWStructuredBuffer<uint> Out : register(u2);
```

kmpeng (commented):
I'm not too well-versed in this area but it LGTM overall. One thing I don't understand though is the difference between `instcombine-sink-load.test` and `simplifycfg-hoist-load.test`

test/Feature/ResourceControlFlow/simplifycfg-hoist-load.test
inbelic: Good catch thanks. I had been working on the ir level and so this must have ignored by `InstCombine` and then when I just added the hlsl source that would create the ir I was working with it was the same.

kmpeng (approved):

PR #1378, opened 2026-07-17
[NFC] Fix clang-tidy issues

Icohedron:
- Modifies the `.clang-tidy` to separate the Checks into multiple lines for better readability, adds -llvm-header-guard to silence clang-tidy warnings/errors regarding `OFFLOADTEST_` prefixes on header guards, and adds `-misc-const-correctness` to disable const correctness checks which have false-positives that differ across clang-tidy versions

- The CMakeLists.txt has been modified to pass `--header-filter=^${CMAKE_CURRENT_SOURCE_DIR}/.*)` to clang-tidy so that it is restricted to scanning project source files. This prevents clang-tidy from scanning llvm-project headers.

- All github workflows now specify -DOFFLOADTEST_USE_CLANG_TIDY=ON (except MacOS which does not have clang-tidy).

- Code refactoring has been performed to eliminate all clang-tidy warnings/errors from the latest clang-tidy version of 22.1.8

Assisted by: Claude Opus 4.8

### DIFF
--- a/include/Support/Pipeline.h
+++ b/include/Support/Pipeline.h
@@ -1059,11 +1059,11 @@ template <> struct ScalarEnumerationTraits<offloadtest::dx::RootParamKind> {
 };
 
 template <typename T> struct SequenceTraits<SmallVector<SmallVector<T>>> {
-  static size_t size(IO &Io, SmallVector<SmallVector<T>> &Seq) {
+  static size_t size(IO & /*Io*/, SmallVector<SmallVector<T>> &Seq) {
     return Seq.size();
   }
 
-  static SmallVector<T> &element(IO &Io, SmallVector<SmallVector<T>> &Seq,
+  static SmallVector<T> &element(IO & /*Io*/, SmallVector<SmallVector<T>> &Seq,
                                  size_t Index) {
     if (Index >= Seq.size())
       Seq.resize(Index + 1);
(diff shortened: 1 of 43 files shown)


include/Support/Pipeline.h
```diff
@@ -1059,11 +1059,11 @@ template <> struct ScalarEnumerationTraits<offloadtest::dx::RootParamKind> {
 };
 
 template <typename T> struct SequenceTraits<SmallVector<SmallVector<T>>> {
-  static size_t size(IO &Io, SmallVector<SmallVector<T>> &Seq) {
+  static size_t size(IO & /*Io*/, SmallVector<SmallVector<T>> &Seq) {
```
bob80905, line 1062: Surprised, this function parameter doesn't cause compilation errors?

bob80905 (approved):
If the PR builds, then I have no other concerns, LGTM

include/Support/Pipeline.h
Icohedron: Compiles on my machine ✅

bogner (approved):
This mostly looks fine. I don't particularly like the NOLINT suppression - maybe we should just disable tests that have false positives?

Icohedron:
> This mostly looks fine. I don't particularly like the NOLINT suppression - maybe we should just disable tests that have false positives?

Can't do that because it's a build error and any machine can perform the build for any of the tests.
We either need to use NOLINT or disable clang-tidy altogether,

bogner:
> > This mostly looks fine. I don't particularly like the NOLINT suppression - maybe we should just disable tests that have false positives?
> 
> Can't do that because it's a build error and any machine can perform the build for any of the tests. We either need to use NOLINT or disable clang-tidy altogether,

Couldn’t we put `-misc-const-correctness` in the config file so it just doesn’t run that check?

Icohedron:
More clang-tidy errors on the builders... I think clang-tidy should just be disabled on the runners.

> > > This mostly looks fine. I don't particularly like the NOLINT suppression - maybe we should just disable tests that have false positives?
> > 
> > 
> > Can't do that because it's a build error and any machine can perform the build for any of the tests. We either need to use NOLINT or disable clang-tidy altogether,
> 
> Couldn’t we put `-misc-const-correctness` in the config file so it just doesn’t run that check?

Can try that. If there are more clang-tidy errors though, I think it'll be better to have clang-tidy off by default.

Icohedron:
Merging now since the PR checks have finished and I don't see any more build or clang-tidy issues.

PR #1382, opened 2026-07-20
Adding lavapipe to FMA XFAIL

joaosaffran:
This patch, rolls back a previous relaxation of FMA, so it keeps the ULP check and XFAIL due to an existing DXC issue.

### DIFF
--- a/test/Feature/HLSLLib/fma.test
+++ b/test/Feature/HLSLLib/fma.test
@@ -92,7 +92,7 @@ DescriptorSets:
         Binding: 3
 #--- end
 
-# Bug https://github.com/llvm/offload-test-suite/issues/1026
+# Bug https://github.com/microsoft/DirectXShaderCompiler/issues/8330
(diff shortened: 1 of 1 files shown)


kmpeng:
Could you change the bug links to the DXC one? https://github.com/microsoft/DirectXShaderCompiler/issues/8330

kmpeng (approved):

llvm-beanz (changes requested):
Please hold off on this pending a larger discussion of floating tolerances.

kmpeng:
You might want to set the label to `test-all` and not disable default tests since the ULP change affects all configurations. Or we need to at least test Intel DXC Vulkan with Lavapipe since that was the one affected by the previous change

bogner:
Is this needed anymore? IIUC the underlying issues were resolved with https://github.com/llvm/offload-test-suite/pull/1394

PR #1386, opened 2026-07-20
XFAIL Lavapipe WaveActiveMax

joaosaffran:
This patch XFail lavapipe on wave active max.


kmpeng (approved):
Can we do `test-vk-all` so we can see all the affected tests?

joaosaffran:
> Can we do `test-vk-all` so we can see all the affected tests?

Testing my fix first for lavapipe since that was the original intent. Once done I do the remaining tests.

.github/workflows/pr-matrix.yaml
```diff
@@ -7,9 +7,21 @@ on:
   pull_request:
     branches:
       - main
+  workflow_dispatch:
```
kmpeng: I don't think you meant to push these `pr-matrix.yaml` changes

PR #1392, opened 2026-07-22
[SPIRV] Enable particle-life.test

farzonl:
This PR: https://github.com/llvm/llvm-project/pull/209912 did some vector widdening for cbuffers. It seems to have been the last piece we needed to enable SPIR-V cbuffer support and turn on the particle life use case.

### DIFF
--- a/test/UseCase/particle-life.test
+++ b/test/UseCase/particle-life.test
@@ -351,10 +351,6 @@ DescriptorSets:
 ...
 #--- end
 
-# Unimplemented: CBuffer not implemented in Clang's Vulkan implementation.
-# https://github.com/llvm/llvm-project/issues/124597
-# XFAIL: Clang && Vulkan
-
 # CBuffer bindings seem to be broken under metal
 # Bug https://github.com/llvm/offload-test-suite/issues/55
 # XFAIL: Metal


damyanp (approved):

kmpeng (commented):
Seems like this test will still fail on [Windows-VK-QC-Clang](https://github.com/llvm/offload-test-suite/actions/runs/29920074583/job/88928897064#logs) btw

farzonl:
> Seems like this test will still fail on [Windows-VK-QC-Clang](https://github.com/llvm/offload-test-suite/actions/runs/29920074583/job/88928897064#logs) btw

Since I can't test these locally, lets let it fail for a bit so I can see why it is failing before we xfail it.

kmpeng:
> Since I can't test these locally, lets let it fail for a bit so I can see why it is failing before we xfail it.

You can use the `test-vk-all` label to see why it's failing now (and maybe add `disable-default-tests`)

farzonl:
> > Since I can't test these locally, lets let it fail for a bit so I can see why it is failing before we xfail it.
> 
> You can use the `test-vk-all` label to see why it's failing now (and maybe add `disable-default-tests`)

Thanks for the suggestion. I'm aware I just think that investigation can wait since QC is still experimental.

kmpeng (approved):

PR #1393, opened 2026-07-22
[SPIRV] Enable switch statement tests

### LINKED ISSUE #213802: [SPIR-V] Failure to legalize 16-lane PHI generated by HLSL matrix swizzles
## Issue Discovered
 In trying to enable the `matrix_swizzle_one_based.test` and `matrix_swizzle_zero_based.test`  In the offload test suite: https://github.com/llvm/offload-test-suite/pull/1393 We discovered that the issue blocking the tests was broader than just lack of switch statement support in the spirv backend. the following legalization rule was failing

```
| fatal error: error in backend: unable to legalize instruction: %1012:_(<4 x s32>), %1013:_(<4 x s32>), %1014:_(<4 x s32>), %1015:_(<4 x s32>) = G_UNMERGE_VALUES %34:viid(<16 x s32>) (in function: main)
```

## Analysis of issue
Compiling the HLSL matrix swizzle tests for Vulkan through the SPIR-V backend fails during GlobalISel legalization.

A `G_PHI` with type <16 x s32> is treated as legal, even though shader execution models only support SPIR-V vectors with up to four components. Later legalization attempts to split that value using `G_UNMERGE_VALUES`, which the backend cannot legalize. It appears the testing for this expected this to work becaause a <32 x s32> test exists and is properly splitting.

## Cause
The SPIR-V legalizer explicitly includes `v16s32` in `allPtrsScalarsAndVectors`. The `G_PHI` rules check this explicit legal type list before applying the rule that limits vectors to `MaxVectorSize`.

In shader mode,`MaxVectorSize` is four. Because legalizer rules use first-match semantics, the `<16 x s32>` `PHI` is accepted before the wide-vector splitting rule is reached.

farzonl:
This change was not really a matrix spir-v bug, but it did disable our swizzle testing. Now that https://github.com/llvm/llvm-project/pull/209859 has merged issue https://github.com/llvm/llvm-project/issues/209310 is fixed and we can re-enable these matrix swizzle tests for spir-v via llvm\clang.

This secondary issue also had to be fixed: https://github.com/llvm/llvm-project/issues/213802


Icohedron (approved):
LGTM assuming the tests do pass

kmpeng (approved):

farzonl:
I need one more pr to merge before we re enable these tests. Moving to draft

PR #1394, opened 2026-07-23
Add ZeroTolerance for ULP compares near 0

llvm-beanz:
ULP comparisons get really gnarly for values very close to zero. In those cases extremely small value differences caused by rounding errors can result in very large ULP differences.

To normalize for this I've added an optional "ZeroTolerance" value that allows specifying an epsilon value for matching against values when either the result or expected value are within the epsilon of zero.


test/Feature/HLSLLib/fma.test
kmpeng: Remove this - it no longer XFAILs with this change

lib/Support/Check.cpp
kmpeng, line 146: Should `ZeroTolerance` be added to `compareFloat16ULP` too?

lib/Support/Pipeline.cpp
```diff
@@ -609,6 +609,7 @@ void MappingTraits<offloadtest::Result>::mapping(IO &I,
   switch (R.ComparisonRule) {
   case Rule::BufferFloatULP: {
     I.mapRequired("ULPT", R.ULPT);
+    I.mapOptional("ZeroTolerance", R.ZeroTolerance, 0.0);
     I.mapOptional("DenormMode", R.DM);
```
kmpeng: nit: `ZeroTolerance` comes after `DenormMode` everywhere else

test/Feature/HLSLLib/fma.test
```diff
@@ -51,13 +51,14 @@ Buffers:
     Stride: 32
     Data: [ 3.5, -7.75, -3.5, 6.0,
             -3.5, 6.5, 9.25, -6.5,
-            0x1.0000000000000p-26, -0x1.0000000000000p-26, 0x1.0000000000000p-104, -3.0,
-            0x1.0000000000000p-26, -0x1.0000000000000p-26, 0x1.0000000000000p-104, -3.0 ]
+            0, 0, 0, -3.0,
+            0, 0, 0, -3.0 ]
```
kmpeng: What's the reason for changing these to 0? I think they should still stay as the exact expected values

kmpeng (commented):
Overall looks good, had a couple of comments. We should also add this `ZeroTolerance` field to the readme

kmpeng (approved):

V-FEXrt (approved):

Issue #1388, opened 2026-07-20
[Infra] Restructure CI pipelines so one machine distributes builds to all test runner machines


bob80905:
Previously, in https://github.com/llvm/offload-test-suite/pull/1328, we enabled the split build feature.
This feature allows a specific build and test / vendor tuple to be built by a non-matching vendor, as long as the platform architecture matches.
In this way, if AMD is idle and we need to build and test something on NVIDIA, AMD can build the binaries and distribute it to the NVIDIA machine to download the binaries in the artifact and test with them, saving some time and improving parallelization / performance.
This did make a marked improvement, but after adding the lavapipe configuration to the test matrix, the testing load demands that we carry the original idea through:
We need a dedicated build machine that frequently builds the relevant binaries, and distributes them to all machines. We don't need to duplicate builds of llvm / dxc, we only need the offload test suite to be the latest build.
So, this issue depends on https://github.com/llvm/offload-test-suite/pull/1376 getting merged in first, so that the offload test suite can be built independently.
The pipeline / yaml should be changed so that one distinguished label gets assigned the responsibility of frequently rebuilding dxc / clang, and uploading these binaries as artifacts.
Other machines will pull the appropriate artifacts (based on an expected commit hash and other config stuff like architecture), and build latest offload test suite binaries, and run the tests.
This cuts the need for redundant builds from N (where N is the number of different test configurations for a specific architecture (e.g., clang vs dxc, Warp vs lavapipe vs intel vs nvidia vs AMD, d3d12 vs vulkan, ~2 * 5 * 2, or about 20), to 1.
This should also improve the qualcomm performance, since qualcomm has not taken advantage of the split build feature.
It is the only vendor with its own unique (arm64) architecture, and doesn't benefit from another machine building the relevant binaries. 
If the dedicated builder can build the binaries targeting arm64, then qualcomm can dramatically reduce the time spent building for tests.
This issue tracks setting up a dedicated frequent builder that builds dxc/clang targeting x64 and arm64, and distributes these builds to all test runners via an artifact store. The test runners ought to be capable of pulling down the right artifacts, and test jobs depend on the Build job; a failed build automatically skips dependent test jobs.

PR #1406, opened 2026-07-29
Set up frequent builder machine that builds dxc and clang on x64

bob80905:
The new frequent builder machine has been registered as a github actions runner.
To improve performance, we need a machine to frequently build dxc / clang and distribute it to the test runner machines.
This PR sets up the new machine to frequently build dxc / clang.
It does not yet distribute the binaries to the test runner machines, they will stay on their own same path.
We will set this machine up to frequently build and keep an eye on whether the builds look good, then slowly hook up
the x64 machines so they take the built binaries from this machine and skip their own build step.

Assisted by: Github Copilot
Addresses https://github.com/llvm/offload-test-suite/issues/1388


.github/workflows/build-callable.yaml
```diff
@@ -0,0 +1,201 @@
+name: HLSL Frequent Builder
```
damyanp: I think we've ended up over indexing on the name "frequent builder".  Maybe this was used while discussing the design, but I don't think it's really one that we should use going forward.

This is just the....build job, isn't it?  Or the DXC and Clang build job?  Saying that out loud makes me wonder if we really need to tie the DXC and Clang builds together - we can probably build DXC at a much lower frequency, for example.

.github/workflows/build-dxc-callable.yaml
```diff
@@ -0,0 +1,201 @@
...
+
+jobs:
+  build:
+    # Dedicated frequent-builder pool. Machines in this pool carry the
+    # `hlsl-frequent-builder` label and no `hlsl-<sku>` label, so GitHub
+    # Actions can never schedule a GPU test job here.
+    runs-on: [self-hosted, hlsl-frequent-builder]
+    # One build at a time per ref and target. Queued builds are kept, not
+    # cancelled -- the builder shares a pool, so a wait can exceed the
+    # schedule interval. timeout-minutes limits a stuck build instead.
```
damyanp, line 47: This seems like an odd comment to me....took me a while to realize that it's explaining both the concurrency and the timeout settings.  At the same time, I don't entirely understand why we need this here.

.github/workflows/build-callable.yaml
```diff
@@ -0,0 +1,201 @@
...
+          fetch-depth: 1
+      - name: Compute artifact names
+        id: key
+        shell: bash
+        env:
+          OS: ${{ inputs.OS }}
+          ARCH: ${{ inputs.Arch }}
+        run: |
+          set -euxo pipefail
+          # Stable names: PRs locate these by (run id, name).
```
damyanp: I don't understand what this comment means.  It says "PRs locate these by (run id, name)" but then the names are made based on `${OS}` and `${ARCH}`.

.github/workflows/build-callable.yaml
```diff
@@ -0,0 +1,201 @@
...
+          path: ${{ runner.temp }}/build-artifacts.tar
+          # Must outlive the gap to the next successful scheduled build,
+          # since PRs consume the newest successful run's artifacts.
+          retention-days: 3
+      - name: Upload DXC artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ steps.key.outputs.dxc_artifact_name }}
+          path: ${{ runner.temp }}/dxc-artifacts.tar
+          retention-days: 3
```
damyanp: Seeing them separate like this really does make me think these should be separate jobs / actions / whatever they're called in github.

.github/workflows/frequent-build.yaml
```diff
@@ -0,0 +1,26 @@
...
+  # No pull_request trigger: this runs on a self-hosted machine, and a fork
+  # PR must never execute there. Validate builder changes by dispatching
+  # this workflow against the branch.
+  workflow_dispatch:
+  schedule:
+    - cron: '0 */2 * * *' # run every 2 hours
+
+jobs:
+  Build-Windows-X64:
+    uses: ./.github/workflows/build-callable.yaml
```
damyanp: Is there a good reason to separate the yaml files like this?  Can we not just have a single "build-dxc.yaml" and "build-clang.yaml" file?

.github/workflows/pr-matrix.yaml
```diff
@@ -13,6 +13,66 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # TODO: migration template. Uncomment the resolver and one SKU at a time,
```
damyanp: Should we really be checking in these commented out changes?

cmake/caches/StandaloneDistribution.cmake
```diff
@@ -13,6 +13,10 @@ set(LLVM_INSTALL_UTILS ON CACHE BOOL "")
 set(LLVM_INSTALL_TOOLCHAIN_ONLY OFF CACHE BOOL "")
 set(LLVM_DISTRIBUTION_COMPONENTS
     clang
+    # Ships clang-tidy so a standalone offload-test-suite build can enable
+    # OFFLOADTEST_USE_CLANG_TIDY without needing a separate Clang toolchain
+    # on the machine doing the standalone build.
+    clang-tidy
```
damyanp, line 19: Why does clang-tidy get a big description of why it belongs here but none of the others do?

docs/offload-distribution.md
```diff
@@ -214,15 +204,162 @@ Then configure and build the offload test suite with a command like so:
...
 
 In this configuration the offload-test-suite will build its tools against the
 already built LLVM libraries which dramatically reduces build time. This
 configuration does still require a checkout of the LLVM source tree to pull LIT
-and the third-party unit testing libraries. If clone/checkout time or disk space
-is a concern this could be a sparse checkout or future changes could allow this
-to use LIT from pip and a stock googletest framework.
+and the third-party unit testing libraries.
+
+## Frequent Builder
```
damyanp: I'm not going to spend time reading this.  I assume it isn't really intended for human consumption?

.github/workflows/build-callable.yaml
bob80905: I wanted to encode the information that this isn't a PR triggered workflow, rather a scheduled workflow.
Maybe `Scheduled Builder`, does that sound better than frequent builder?
But I fear calling it just the build job would lead people to believe that the build triggers when a new PR opens.
Yes, I agree we could build DXC at a slower frequency.

.github/workflows/build-dxc-callable.yaml
bob80905: It's explaining or at least attempting to explain that during the bring-up phase, the scheduled builder will not only need to build on its assigned schedule, but will also take up build jobs for other scheduled SKU jobs, and PR triggered build jobs. So this machine will possibly have a backlog of scheduled build jobs.
We need to not cancel in-progress jobs. The timeout will take care of stuck jobs, but otherwise jobs that are queued shouldn't be cancelled out.

.github/workflows/build-callable.yaml
bob80905: Understandable, it's explained more in the offload-distribution.md doc.
Basically, if a test job is looking for an artifact to download, it'll run a resolution step to determine the runid/name from which to pull the artifact from. Then it'll pull the artifacts from that runid-associated run, with OS and Arch in the artifact name.
Artifacts live inside of a run, so we don't need to embed the runid / name into the artifact name to disambiguate it. But if we have different OSs or Arch's in the SAME run then we need to embed that into the artifact name to disambiguate within that run.
This comment could cause confusion without including more details and getting long, so I'll just remove it.

cmake/caches/StandaloneDistribution.cmake
bob80905: I think it is self-evident why the others are included in the distribution components, but clang-tidy is not so self-explanatory.
Without including clang-tidy, IIRC, there were build errors when building the offload-test-suite standalone.

docs/offload-distribution.md
bob80905: It's documentation intended for human consumption. For those that are interested / need to understand how the feature works, the text below describes how the infrastructure and the pieces fit together.
However, I do think it should be moved to its own .md file, rather than tacked onto this file.

.github/workflows/frequent-build.yaml
bob80905: This is the repo convention, many of the trigger files already call build-and-test-callable from a separate yaml file, and it's clear what the trigger policy is.
Adding arm64 later on should be straightforward too, we can define it clearly right here.

.github/workflows/pr-matrix.yaml
bob80905: I guess we can do without them.
My thought process was it shows what the intent is and how the code will be used. Fine either way, I can remove it.

.github/workflows/build-dxc-callable.yaml
```diff
@@ -0,0 +1,147 @@
...
+    inputs:
+      OS:
+        description: 'Target OS. Accepted values: `windows`. macOS is not supported -- the frequent-builder pool has no macOS machines. Enforced by the "Validate inputs" step.'
+        required: true
+        type: string
+      Arch:
+        description: 'Target arch. Accepted values: `x64`. arm64 cross-compile lands in Phase 2. Enforced by the "Validate inputs" step.'
+        required: true
+        type: string
+      OffloadTest-branch:
```
llvm-beanz, line 22: If this is just building DXC, it shouldn't need to checkout the offload test suite right?

It looks to me like the only thing we're actually pulling from there is the cmake cache file to configure scccache. We can probably just set those variables manually.

.github/workflows/build-llvm-callable.yaml
```diff
@@ -0,0 +1,131 @@
...
+    inputs:
+      OS:
+        description: 'Target OS. Accepted values: `windows`. macOS is not supported -- the frequent-builder pool has no macOS machines. Enforced by the "Validate inputs" step.'
+        required: true
+        type: string
+      Arch:
+        description: 'Target arch. Accepted values: `x64`. arm64 cross-compile lands in Phase 2. Enforced by the "Validate inputs" step.'
+        required: true
+        type: string
+      OffloadTest-branch:
```
llvm-beanz, line 22: Similar concern here to the DXC build. I think checking out the offload test suite just for a few cmake caches is a bit overkill. This one is a bit more complicated than the DXC case, so it might be warranted, but we should checkout only the cmake/caches subdirectory.

cmake/caches/StandaloneDistribution.cmake
llvm-beanz: There are tradeoffs to doing this... using an unreleased in-development clang-tidy will allow us to align all the builders on roughly the same clang-tidy, but it's not necessarily a stable clang-tidy. This workflow also means we will be mismatching clang-tidy and the compiler we use to build (which I guess is maybe normal-ish on Windows), which can cause some odd issues too.

I'm okay with trying this out, but it may not be something we want to stick to.

llvm-beanz (approved):

PR #1408, opened 2026-07-30
Add `step` matrix tests

### LINKED ISSUE #795: [Matrix] Add matrix test for `step`
> Note: This issue tracks matrix testing completion for `step`. The original test issue is [here](https://github.com/llvm/offload-test-suite/issues/154).  

Matrix requirements:

kmpeng:
Closes #795.

Adds matrix tests for `step` testing half and float. Will remove the XFAILs once the Clang implementation is in.

Assisted-by: Claude Opus 4.8

### DIFF
--- a/test/Feature/HLSLLib/step_mat.16.test
+++ b/test/Feature/HLSLLib/step_mat.16.test
@@ -1,0 +1,131 @@
+#--- source.hlsl
+StructuredBuffer<half4> Y : register(t0);
+StructuredBuffer<half4> X : register(t1);
+
+RWStructuredBuffer<half4> Out : register(u2);
+
+[numthreads(1,1,1)]
+void main() {
+  half3x2 F32 = step(half3x2(Y[0].xyz, Y[1].xyz),
+                     half3x2(X[0].xyz, X[1].xyz));
+  half2x4 F24 = step(half2x4(Y[2], Y[3]),
+                     half2x4(X[2], X[3]));
+  half3x1 F31 = step(half3x1(Y[4].xyz),
+                     half3x1(X[4].xyz));
+  half4x4 F44 = step(half4x4(Y[5], Y[6], Y[7], Y[8]),
+                     half4x4(X[5], X[6], X[7], X[8]));
+  half2x2 F22Const = step(half2x2(0.0, 1.0, -1.0, 2.0),
+                          half2x2(0.0, 0.5, -1.0, 3.0));
+
+  Out[0] = half4(F32[0], F32[1]);
+  Out[1] = half4(F32[2], 0.0, 0.0);
+  Out[2] = F24[0];
+  Out[3] = F24[1];
+  Out[4] = half4(F31[0], F31[1], F31[2], 0.0);
+  Out[5] = F44[0];
+  Out[6] = F44[1];
+  Out[7] = F44[2];
+  Out[8] = F44[3];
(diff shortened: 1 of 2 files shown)


test/Feature/HLSLLib/step_mat.16.test
```diff
@@ -0,0 +1,131 @@
...
+            0x3C00, 0, 0, 0,
+            0x3C00, 0x3C00, 0, 0x3C00,
+            0x3C00, 0, 0x3C00, 0,
+            0x3C00, 0, 0x3C00, 0,
+            0x3C00, 0, 0x3C00, 0x3C00,
+            0x3C00, 0, 0x3C00, 0x3C00,
+            0x3C00, 0, 0x3C00, 0x3C00,
+            0x3C00, 0, 0x3C00, 0,
+            0x3C00, 0, 0x3C00, 0x3C00 ]
+    # 3x2: 1, 1, 0, 1, 1, 0, _, _
```
bob80905, line 89: I think there should be an underscore in position 4

test/Feature/HLSLLib/step_mat.32.test
```diff
@@ -0,0 +1,114 @@
...
+  float2x4 F24 = step(float2x4(Y[2], Y[3]),
+                      float2x4(X[2], X[3]));
+  float3x1 F31 = step(float3x1(Y[4].xyz),
+                      float3x1(X[4].xyz));
+  float4x4 F44 = step(float4x4(Y[5], Y[6], Y[7], Y[8]),
+                      float4x4(X[5], X[6], X[7], X[8]));
+  float2x2 F22Const = step(float2x2(0.0, 1.0, -1.0, 2.0),
+                           float2x2(0.0, 0.5, -1.0, 3.0));
+
+  Out[0] = float4(F32[0], F32[1]);
```
bob80905, line 20: I am not sure this is what you intended?
F32[0/1] should return a column of 3 floats, so you are making one float4 out of 2 float3s, if I understand correctly.
Regardless, could you change this to use the .xy subscript so it's absolutely clear what is intended? Or maybe use double subscripts?

test/Feature/HLSLLib/step_mat.16.test
kmpeng: Position 4 is a real element. Out isn't packed like the In buffers - the matrix elements are packed contiguously, so the padding is on positions 7-8

test/Feature/HLSLLib/step_mat.32.test
kmpeng: F32[0/1/2] are rows, not columns. F32[i] returns a float2, so this is two float2s -> one float4, same pattern as `pow_mat`/`atan2_mat`. But yes I can add `.xy` for clarity

test/Feature/HLSLLib/step_mat.16.test
bob80905: Ah I see, the 3x2 confused me, but you're referring to the 3x2 matrix set of outputs rather than that the output is a 3x2.

test/Feature/HLSLLib/step_mat.32.test
bob80905: Understood, apologies I am rusty on the matrix stuff. No need if a convention was established already.

bob80905 (approved):

Icohedron (approved):

PR #1415, opened 2026-08-03
Dynamic resources: implement support for DirectX and add tests

### LINKED ISSUE #203: Document DXC handling of dynamic resources and add tests(s) to offload-test-suite
Document how DXC handles dynamic resources.
Create a set of test cases and add them to the offload-test-suite. 

https://microsoft.github.io/DirectX-Specs/d3d/HLSL_SM_6_6_DynamicResources.html

### LINKED ISSUE #1360: [DirectX] Add support for dynamic (directly indexed) resources to offload test suite
The test yaml format should be updated to make `DXBinding` optional (under `DescriptorSets -> Resources -> Resource`). Instead, it should allow the resource to have an explicit `HeapIndex` which determines where the resource can be found on the CVB/SRV/UAV or Sampler heap. The `DXBinding` and `HeapIndex` should be mutually exclusive. The DirectX's `Device.cpp` code must be updated to support this explicit `HeapIndex`.

Additionally, when a shader uses directly indexed resources, the root signature needs to have `D3D12_ROOT_SIGNATURE_FLAG_CBV_SRV_UAV_HEAP_DIRECTLY_INDEXED` and/or `D3D12_ROOT_SIGNATURE_FLAG_SAMPLER_HEAP_DIRECTLY_INDEXED` flags set.

hekota:
Adds an optional `HeapIndex` field to the `Resource` YAML config, which places a resource at an explicit slot in a directly-indexed descriptor heap so shaders can reach it through `ResourceDescriptorHeap[i]` / `SamplerDescriptorHeap[i]`. The heap is selected from the resource kind - samplers go to the sampler heap, everything else to the CBV/SRV/UAV heap.

```
DescriptorSets:
  - Resources:
    - Name: DynResIndices     # bound through a descriptor table
      Kind: ConstantBuffer
      DirectXBinding: { Register: 0, Space: 0 }
    - Name: SrvBuf            # reached via ResourceDescriptorHeap[5]
      Kind: StructuredBuffer
      HeapIndex: 5
```

`DirectXBinding`  is now optional, since a directly-indexed resource has no register binding. A resource is reached either through a descriptor table or by indexing the heap, so `DirectXBinding` and `HeapIndex` are mutually exclusive. Specifying both, or neither, is an error.

Changes to DirectX backend:
- Heap-indexed resources are skipped when building descriptor ranges, and the root signature gains `D3D12_ROOT_SIGNATURE_FLAG_CBV_SRV_UAV_HEAP_DIRECTLY_INDEXED` and/or `D3D12_ROOT_SIGNATURE_FLAG_SAMPLER_HEAP_DIRECTLY_INDEXED` depending on which heaps are indexed directly.
- Each heap is sized to fit both the table-bound descriptors and the highest explicit `HeapIndex`. Explicit indices may be sparse.
- A resource with `HeapIndex` is copied to that exact slot. Resources without one keep their existing sequential placement.

Tests added:
- `Feature/DynamicResources/dyn-res-cvb-srv-uav.test` uses constant buffer, SRV and UAV buffers loaded from `ResourceDescriptorHeap` at sparse indices
- `Feature/DynamicResources/dyn-res-texture-sampler.test` uses a texture plus two samplers loaded dynamically, mirroring `Feature/Textures/Texture2D.Sampler.mips.test`

Fixes #1360
Part of [llvm-project\wg-hlsl#203](https://github.com/llvm/wg-hlsl/issues/203)


include/API/Device.h
```diff
@@ -61,8 +61,9 @@ struct InputLayoutDesc {
 
 struct ResourceBindingDesc {
   ResourceKind Kind;
-  DirectXBinding DXBinding;
+  std::optional<DirectXBinding> DXBinding;
   std::optional<VulkanBinding> VKBinding;
+  std::optional<uint32_t> HeapIndex;
```
inbelic, line 66: Should these be a union of a binding/heapindex? Or is it the case that unions don't play nice with yaml config?

Then we can guarantee one is always set in a nice way

lib/API/VK/Device.cpp
```diff
@@ -4491,16 +4491,20 @@ class VulkanDevice : public offloadtest::Device {
     for (auto &S : P.Sets) {
       DescriptorSetLayoutDesc Layout;
       for (auto &R : S.Resources) {
+        assert(!R.HeapIndex && "Direct heap indexing is not yet supported.");
```
inbelic, line 4495: nit: Link to https://github.com/llvm/offload-test-suite/issues/1413

inbelic (commented):
It would be nice to represent this as a (tagged) union if possible, otherwise it looks reasonable to me

include/API/Device.h
hekota: LLVM's yaml framework doesn't have built-in generic support for unions or std::variant. Having both values as optional and then validate that only one is set is the simplest way to implement this.

lib/API/VK/Device.cpp
```diff
@@ -4491,16 +4491,20 @@ class VulkanDevice : public offloadtest::Device {
...
 
         ResourceBindingDesc ResourceBinding = {};
         ResourceBinding.Kind = R.Kind;
-        ResourceBinding.DXBinding.Register = R.DXBinding.Register;
-        ResourceBinding.DXBinding.Space = R.DXBinding.Space;
+        assert(R.DXBinding &&
+               "DXBinding must not be null when HeapIndex is not set.");
+        ResourceBinding.DXBinding = R.DXBinding;
         ResourceBinding.VKBinding = R.VKBinding;
+        ResourceBinding.HeapIndex = R.HeapIndex;
```
bogner, line 4507: This should probably assert on VKBinding/HeapIndex for Vulkan - we don't even use the DXBinding here...

lib/Support/Pipeline.cpp
```diff
@@ -451,8 +451,16 @@ void MappingTraits<offloadtest::Resource>::mapping(IO &I,
...
   I.mapOptional("VulkanBinding", R.VKBinding);
+  I.mapOptional("HeapIndex", R.HeapIndex);
+
+  // A resource is reached either through a descriptor table (DirectXBinding)
+  // or by directly indexing the descriptor heap (HeapIndex), never both.
+  if (R.DXBinding.has_value() == R.HeapIndex.has_value())
+    I.setError(Twine("Resource '") + R.Name +
+               "' must specify exactly one of 'DirectXBinding' or "
+               "'HeapIndex'.");
 }
```
bogner, line 464: I think this should probably also check VKBinding - doing heap indexing on one API and not on the other would be pretty confusing and I can't see it being useful.

bogner (approved):
Generally looks good, but I think we need to be more careful about validating inconsistent bindings.

Issue #1418, opened 2026-08-04
[WARP][Validation Layer][DXC] Hang when DeviceMemoryBarrierWithGroupSync() is followed by a live UAV read


bob80905:
When executing a shader with WARP, with validation layer enabled (GBV), and compiling the shader with DXC, there is a runtime hang.
This is the minimal offload test repro in HLSL:
```hlsl
RWStructuredBuffer<uint> Out : register(u0);
RWStructuredBuffer<uint> OutMono : register(u1);
RWByteAddressBuffer      BABuf : register(u2);
RWStructuredBuffer<uint> OutMono2 : register(u3);

[numthreads(256, 1, 1)]
void main(uint3 GTID : SV_GroupThreadID) {
  if (GTID.x == 0) {
    Out[0] = 0u;
    BABuf.Store(0, 0u);
  }
  DeviceMemoryBarrierWithGroupSync();

  uint Acc = 0u;
  for (uint I = 0; I < 2; ++I) {
    InterlockedAdd(Out[0], 1u);
    DeviceMemoryBarrierWithGroupSync();
    Acc += Out[0];
  }
  OutMono[GTID.x] = Acc;
  OutMono2[GTID.x] = 1u;
}
```

Targeting SM 6.5 also produces a hang.
Here is a table of various results:
| Compiler | Target | Offloader flags | Result |
| --- | --- | --- | --- |
| DXC | `cs_6_5` | `-warp -debug-layer -validation-layer` | **hang** |
| DXC | `cs_6_6` | `-warp -debug-layer -validation-layer` | **hang** |
| DXC | `cs_6_5` | `-warp` | completes |
| clang-dxc | `cs_6_5` | `-warp -debug-layer -validation-layer` | completes |
| clang-dxc | `cs_6_6` | `-warp -debug-layer -validation-layer` | completes |
| clang-dxc | `cs_6_5` | `-warp` | completes |

However, removing the validation layer option for runtime execution resolves the hang issue for both SM targets.

DXIL for SM 6.6:
```
;
; Input signature:
;
; Name                 Index   Mask Register SysValue  Format   Used
; -------------------- ----- ------ -------- -------- ------- ------
; no parameters
;
; Output signature:
;
; Name                 Index   Mask Register SysValue  Format   Used
; -------------------- ----- ------ -------- -------- ------- ------
; no parameters
; shader hash: 7dc7c169d0b5d2d8a75f41fcc3c286f2
;
[...]

PR #1419, opened 2026-08-05
Introduce global timeout for individual lit tests.

### LINKED ISSUE #1420: Bound individual tests by timeout
This issue: https://github.com/llvm/offload-test-suite/issues/1418
Created the need to introduce a timeout for tests.
Hung tests cause the whole HLSL Tests step on test runners to take upwards of 170 minutes.
We need to make sure individual tests do not run too long, so other PRs can get access to valuable test machine resources.

bob80905:
This issue: https://github.com/llvm/offload-test-suite/issues/1418
Created the need to introduce a timeout for tests.
The github actions runners will currently spin up to 6 hours on hung tests, which steals resources from other PRs.
This PR introduces a global max timeout limit of 5 minutes per individual test.
Lit should also usefully report TIMEOUT as a test result status if this time limit is hit.

Assisted by: Github Copilot
Fixes: https://github.com/llvm/offload-test-suite/issues/1420

### DIFF
--- a/.github/workflows/build-and-test-callable.yaml
+++ b/.github/workflows/build-and-test-callable.yaml
@@ -424,31 +424,11 @@ jobs:
           arch: arm64
       - name: Setup Python
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
-      - name: Install lit, pyyaml and psutil
-        # The test runner needs lit, pyyaml and psutil. We install from the
-        # requirements.txt that the build job ships inside the install
-        # prefix (test/CMakeLists.txt installs it under the
-        # offload-test-suite component). That keeps a single source of
-        # truth for pinned versions and avoids drift between this step
-        # and OffloadTest/test/requirements.txt.
-        #
+      - name: Install python dependencies
+        # The test runner needs lit, pyyaml and psutil.
         # We invoke pip explicitly here instead of relying on
         # setup-python's `pip-install:` input, which was observed to
         # silently no-op on our self-hosted Windows runners.
-        #
-        # When pip can't write to the interpreter's system site-packages
-        # (e.g. on self-hosted Windows runners running as NetworkService,
-        # which can't write under C:\Program Files\Python313; or on a
-        # POSIX host where the system Python is owned by root), pip
-        # falls back to a --user install. The console script wrappers
-        # (e.g. lit.exe on Windows, lit on POSIX) then land in the
-        # user-scheme scripts dir, which is NOT on PATH by default:
-        #   - Windows: %APPDATA%\Python\PythonXY\Scripts
-        #   - Linux:   ~/.local/bin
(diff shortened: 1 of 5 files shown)


llvm-beanz (approved):

.github/workflows/build-and-test-callable.yaml
damyanp: pre-existing, but OMG this comment!

.github/workflows/build-and-test-callable.yaml
```diff
@@ -424,8 +424,8 @@ jobs:
           arch: arm64
       - name: Setup Python
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
-      - name: Install lit and pyyaml
-        # The test runner needs only lit + pyyaml. We install from the
+      - name: Install lit, pyyaml and psutil
```
damyanp:
```suggestion
      - name: Install python dependencies
```

test/requirements.txt
```diff
@@ -1,2 +1,4 @@
 pyyaml==6.0.2
 lit==18.1.8
+# Required by lit to enforce maxIndividualTestTime in test/lit.cfg.py.
```
damyanp: Why does psutil get a special comment but pyyaml and lit don't?  If anything else takes a dependency on psutil (likely could happen by accident) are you committing to updating this comment?

test/lit.cfg.py
```diff
@@ -30,6 +30,11 @@
 # directories.
 config.excludes = ["Inputs", "CMakeLists.txt", "README.txt", "LICENSE.txt"]
 
+# Cap each test at 5 minutes so a GPU hang fails one test instead of wedging
```
damyanp:
```suggestion
# Cap each test so a GPU hang fails one test instead of wedging
```

Repeating the time in the comment is just begging for these to get out of sync.

damyanp (approved):
LGTM, but I have a bunch of nits about overly verbose / redundant comments that don't actually need to block the PR from merging.

test/requirements.txt
bob80905: I had thought the other 2 were self-explanatory and psutil could use some documentation as to why we now need it.
However, I am not committing to updating this comment and if something else starts depending on it, the comment could become incorrect, so I'll remove it.

test/lit.cfg.py
```diff
@@ -30,6 +30,11 @@
 # directories.
 config.excludes = ["Inputs", "CMakeLists.txt", "README.txt", "LICENSE.txt"]
 
+# Cap each test so a GPU hang fails one test instead of getting the suite
+# stuck. Requires psutil; --timeout=N still overrides.
+if lit_config.maxIndividualTestTime == 0:
+    lit_config.maxIndividualTestTime = 300
```
bogner, line 36: What can cause this to not be zero when we get here? The comment sort of implies `--timeout=N` wouldn't work if not, but as far as I can tell that's an unrelated value.

PR #1422, opened 2026-08-06
Fix validation workflow build errors

bob80905:
This PR fixes some errors in the validation workflow that were discovered when manually dispatching it.
Has to do with the fact that building the offloadtest suite standalone makes things slightly different than building through the legacy in-tree path.

For one, we are missing `HandleLLVMOptions`, which opts out of errors for deprecated code like `fopen`.

Another issue is that we are using upstream clang-tidy instead of whatever clang-tidy is built in to the build machines.
Latest clang-tidy will warn on declaring constructors with one argument without the `explicit` keyword, and will treat the warning as an error.
This PR resolves both issues, to turn the validation workflow back to green.

Assisted by: Github Copilot
Addresses: https://github.com/llvm/offload-test-suite/issues/1425
Link to a successful test run of the validation pipeline: https://github.com/llvm/offload-test-suite/actions/runs/31058724325/job/92481811942

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,20 @@ if (OFFLOADTEST_BUILT_STANDALONE)
   set(CMAKE_CXX_STANDARD_REQUIRED YES)
   set(CMAKE_CXX_EXTENSIONS NO)
 
+  # In-tree builds get these from LLVM's HandleLLVMOptions.cmake. Standalone
+  # builds do not include it, so repeat the CRT deprecation opt-outs here or
+  # MSVC flags calls like fopen as deprecated.
+  if (MSVC)
+    add_compile_definitions(
+      _CRT_SECURE_NO_DEPRECATE
+      _CRT_SECURE_NO_WARNINGS
+      _CRT_NONSTDC_NO_DEPRECATE
+      _CRT_NONSTDC_NO_WARNINGS
+      _SCL_SECURE_NO_DEPRECATE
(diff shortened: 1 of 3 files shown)


damyanp (commented):
By themselves these changes look reasonable.  However, the linked issue that the PR description says it "addresses" doesn't seem to mention any of these things (although it's a pretty difficult to read that issue).

In addition to this, I don't understand why the `explicit`s are required and how that fixes something to do with this being used in standalone builds?

Also, talking about "validation workflow" I think assumes some knowledge about things that I'm lacking.  What's the "validation workflow"?  If these changes are general goodness we should be able to explain and describe them on their own terms.

CMakeLists.txt
```diff
@@ -22,6 +22,20 @@ if (OFFLOADTEST_BUILT_STANDALONE)
   set(CMAKE_CXX_STANDARD_REQUIRED YES)
   set(CMAKE_CXX_EXTENSIONS NO)
 
+  # In-tree builds get these from LLVM's HandleLLVMOptions.cmake. Standalone
+  # builds do not include it, so repeat the CRT deprecation opt-outs here or
+  # MSVC flags calls like fopen as deprecated.
+  if (MSVC)
```
llvm-beanz: I think the right fix here is just to add `include(HandleLLVMOptions)` above right underneath the include of `AddLLVM`.

llvm-beanz (approved):

PR #1433, opened 2026-08-10
Add InterlockedCompareExchange tests

### LINKED ISSUE #850: Add test for `InterlockedCompareExchange`
- [ ] Test for 16 bit int types (`InterlockedCompareExchange.int16.test`)
- [ ] Test for Half type (`InterlockedCompareExchange.fp16.test`)
- [ ] Test for 32 bit types (`InterlockedCompareExchange.32.test`)
- [ ] Test for 64 bit int types (`InterlockedCompareExchange.int64.test`)
- [ ] Test for Double type (`InterlockedCompareExchange.fp64.test`)

A .test file should contain 
```
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# RUN: split-file %s %t
```

bob80905:
This PR adds tests for InterlockedCompareExchange.
The testing method mimics that of InterlockedCompareStore merged here: https://github.com/llvm/offload-test-suite/pull/1416

Fixes https://github.com/llvm/offload-test-suite/issues/850
Assisted by: Github Copilot

### DIFF
--- a/test/Feature/HLSLLib/InterlockedCompareExchange.32.test
+++ b/test/Feature/HLSLLib/InterlockedCompareExchange.32.test
@@ -1,0 +1,237 @@
+#--- source.hlsl
+
+// Tests InterlockedCompareExchange on groupshared destinations, int and
+// uint, with 256 concurrent threads. Four checks:
+//
+//   1. Per-thread private slots: a mismatching compare must not store, a
+//      following matching compare must, and both must report the original
+//      value.
+//   2. Contended locations: every thread proposes the same new value, so
+//      only one store can take effect and exactly one thread may observe
+//      the original compare value. A never-matching compare must leave the
+//      destination alone and always report it unchanged.
+//   3. Counter: thread `i` only advances `Counter` from `i` to `i+1`, so
+//      over 256 barrier-separated attempts each thread succeeds exactly
+//      once and the counter must end at 256. Monotonicity: the counter
+//      only moves up, so each fresh re-read must not be below the previous.
+
+RWStructuredBuffer<uint> OutMono : register(u0);
+RWStructuredBuffer<uint> OutSlots : register(u1);
+RWStructuredBuffer<uint> OutOrig : register(u2);
+RWStructuredBuffer<uint> OutWins : register(u3);
+RWStructuredBuffer<int> OutFinal : register(u4);
+
+groupshared uint Counter;        // stepped from 0 to 256, one thread per step
+groupshared int  SlotsInt[256];  // per-thread slots, signed
+groupshared uint SlotsUInt[256]; // per-thread slots, unsigned
+groupshared int  MatchInt;       // contended, compare matches
+groupshared int  NoMatchInt;     // contended, compare never matches
+groupshared uint MatchUInt;      // contended, compare matches
+groupshared uint NoMatchUInt;    // contended, compare never matches
+groupshared uint WinCount;       // threads that observed the compare value
+
+[numthreads(256, 1, 1)]
+void main(uint3 GTID : SV_GroupThreadID) {
+  SlotsInt[GTID.x] = (int)GTID.x - 128;
+  SlotsUInt[GTID.x] = GTID.x + 7u;
+  if (GTID.x == 0) {
+    Counter = 0u;
+    MatchInt = -5;
+    NoMatchInt = 77;
+    MatchUInt = 0xFFFFFFFFu;
+    NoMatchUInt = 5u;
+    WinCount = 0u;
+  }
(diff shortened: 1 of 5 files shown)


inbelic (approved):

test/Feature/HLSLLib/InterlockedCompareExchange.32.test
```diff
@@ -0,0 +1,237 @@
...
+  InterlockedCompareExchange(MatchInt, -5, 42, OrigMatchInt);
+  InterlockedCompareExchange(NoMatchInt, 78, -1, OrigNoMatchInt);
+  InterlockedCompareExchange(MatchUInt, 0xFFFFFFFFu, 1234u, OrigMatchUInt);
+  InterlockedCompareExchange(NoMatchUInt, 6u, 999u, OrigNoMatchUInt);
+
+  // Exactly one thread may observe the pre-existing value of MatchInt.
+  if (OrigMatchInt == -5)
+    InterlockedAdd(WinCount, 1u);
+
+  OutOrig[GTID.x] = ((OrigMatchInt == -5 || OrigMatchInt == 42) &&
```
V-FEXrt, line 82: This check would benefit from named constants. Right now you have to go back to the top of the test and make sure each value matches with the previously assigned value which is challenging to keep track of.

test/Feature/HLSLLib/InterlockedCompareExchange.32.test
```diff
@@ -0,0 +1,237 @@
...
+                     OrigNoMatchUInt == 5u) ? 1u : 0u;
+
+  // Counter, checked for exactly one success per thread and for monotonic
+  // non-decrease.
+  uint CompareValue = GTID.x;
+  uint NewValue = GTID.x + 1u;
+  uint Prev = 0u;
+  uint Mono = 1u;
+  uint Wins = 0u;
+  for (uint I = 0; I < 256; ++I) {
```
V-FEXrt, line 95: Why is its necessary to loop 256 times here when we already have 256 threads? Seems like we are checking to make sure the value only increased once on this thread but its not clear to me how looping 256 times does that or why we need a memory barrier for each loop

test/Feature/HLSLLib/InterlockedCompareExchange.32.test
V-FEXrt: this comment generally applies to all the constants in all these tests, but I didn't was to keep reposting it.

Feel free to tell me I'm being silly/we haven't done this for other offloads tests and shouldn't start now if thats the case

bob80905: Your concern is understandable. Several previous offload tests do that, and we shouldn't start now, though I would be open to doing a pass for this variable renaming after interlocked tests are all added.

test/Feature/HLSLLib/InterlockedCompareExchange.32.test
bob80905: Counter starts at 0.
This loop is designed to increment Counter from 0 to 256.
CompareValue is set to the executing thread's ID. Within the loop, that is only one thread.
And, NewValue is always the current thread id + 1.
The reason we want to loop 256 times is because we want for each thread to observe that after 256 iterations of the loop, Counter always arrives at 256 (it should because among all threads, at least one will have the correct CompareValue).
We want to allow cases of high contention, where either the comparevalue is wrong, or if there was a racy implementation of the interlocked builtin, we maximize the chances of another thread executing the interlocked builtin at the perfect time.
Sure thread 1 might execute interlocked and then 2 right after, in the case this was one iteration, and everything works, but on the next iteration, maybe the immediately adjacent threads (2 and 3) might execute at the right time and cause a clobber.

We have 256 threads, so why not increment Counter up to 256? We could even use 500 but the comparevalue variable wont match Counter at any point after 256.

PR #1434, opened 2026-08-10
Modernize `InterlockedAdd` tests

bob80905:
This PR modernizes the interlockedadd offload test suite tests.
The tests were missing the resource data types, and monotonicity verification, as this was one of the very first interlocked-type tests added.
Now that an established pattern is used to test new interlocked intrinsics, we should update the legacy tests to conform to the pattern of:
using 256 threads instead of 64
verifying monotonicity (have a counter scenario)
test on resource data types

Assisted by: Github Copilot


test/Feature/HLSLLib/InterlockedAdd.32.test
```diff
@@ -117,6 +170,10 @@ Results:
     Rule: BufferExact
     Actual: OutFinal
     Expected: ExpectedFinal
+  - Result: TestMono
+    Rule: BufferExact
+    Actual: OutMono
+    Expected: ExpectedSlotsInt
```
inbelic: `TestMono` is checked against `ExpectedSlotsInt`. It happens to be the right data (256 ones), but reusing the slots expectation for an unrelated result is confusing, and the new files in this same PR (and `InterlockedCompareExchange.32.test`) already define a single `ExpectedOnes` buffer for exactly this. Suggest renaming `ExpectedSlotsInt`/`ExpectedSlotsUInt` to one shared `ExpectedOnes` here and in `InterlockedAdd.int64.test:183` so all three all-ones results reference the same, correctly named buffer.

test/Feature/HLSLLib/InterlockedAdd.int64.test
```diff
@@ -36,6 +38,7 @@ void main(uint3 GTID : SV_GroupThreadID) {
     CounterUIntNoOrig = 0;
     SignedCounter = 1000;
     LargeCounter = 0x100000000ull; // 2^32
+    MonoCounter = 0x100000000ull;  // 2^32, so the loop crosses the upper word
```
inbelic: Is the trailing comment accurate? It seems like the loop only adds 256 threads * 256 iterations = 65536, so starting at exactly 2^32 means the counter never carries across the 32-bit boundary, IIUC

test/Feature/HLSLLib/InterlockedAdd.int64.test
```diff
@@ -60,18 +63,38 @@ void main(uint3 GTID : SV_GroupThreadID) {
...
+  uint64_t PrevOrig = 0;
+  uint Mono = 1u;
+  for (uint I = 0; I < 256; ++I) {
+    uint64_t Orig;
+    InterlockedAdd(MonoCounter, (uint64_t)1, Orig);
+    if (I > 0 && Orig <= PrevOrig)
+      Mono = 0u;
+    PrevOrig = Orig;
+    GroupMemoryBarrierWithGroupSync();
+    uint64_t Cur = MonoCounter;
```
inbelic: Flakiness risk specific to the 64-bit variants: this is a *non-atomic* 64-bit load racing against concurrent 64-bit atomic adds. Nothing guarantees it is not implemented as two 32-bit loads, so a torn read (low word wrapped, high word stale) can produce `Cur < Prev` and fail the test for a perfectly correct implementation. The existing tests do exactly one such re-read per thread; this loop does 256 per thread * 256 threads = 65536 chances per run, and the same pattern appears in `InterlockedAdd.resources.int64.test:76` and `InterlockedAdd.resources.typed.int64.test:55`. Note the tearing window is tiny but the aggregate exposure is not. Consider deriving the fresh re-read from a 3-argument atomic (e.g. `InterlockedAdd(MonoCounter, 0, Cur)`) so the observation is atomic too.

test/Feature/HLSLLib/InterlockedAdd.int64.test
```diff
@@ -149,17 +204,24 @@ DescriptorSets:
...
 # REQUIRES: Int64 && Int64GroupSharedAtomics && SM_6_6
 
-# Unimplemented: https://github.com/llvm/llvm-project/issues/99122
-# XFAIL: Clang
-
 # Bug: https://github.com/llvm/offload-test-suite/issues/1164
 # XFAIL: DXC && Metal
 
+# Bug: https://github.com/llvm/offload-test-suite/issues/1402
+# XFAIL: Clang && DirectX
```
inbelic, line 207: All three new resources tests XFAIL `Clang && Vulkan` for #1435, and this file seems to have the same pattern, why does it not need a Vulkan XFAIL?

test/Feature/HLSLLib/InterlockedAdd.resources.32.test
```diff
@@ -0,0 +1,284 @@
...
+//   * RWStructuredBuffer<int>[i]           (free function, 3-arg, signed)
+//   * RWBuffer<uint>[i]                    (free function on typed UAV)
+//   * RWTexture2D<uint>[coord]             (free function on typed UAV)
+//   * RWByteAddressBuffer::InterlockedAdd  (method, 2-arg and 3-arg)
+//
+// 256 threads each add 1 to a shared accumulator, so every accumulator must
+// end at exactly 256 and no update may be lost. The 3-argument forms are
+// checked for unique originals via a "seen" array, and the structured
+// buffer counter is additionally checked for monotonic non-decrease across
+// a barrier-separated loop, as in InterlockedAdd.32.test.
```
inbelic: Header/code mismatch: this says the *structured buffer* counter is checked for monotonic non-decrease, but the loop at lines 70-84 runs on the `RWByteAddressBuffer` counter at offset 8 (`BABuf.InterlockedAdd(8, ...)` / `BABuf.Load(8)`), and `ExpectedBABuf[2]` is what encodes it. Also, line 12-13 ("every accumulator must end at exactly 256") is not true for that third accumulator, which ends at 65536.

test/Feature/HLSLLib/InterlockedAdd.resources.32.test
```diff
@@ -0,0 +1,284 @@
...
+void main(uint3 GTID : SV_GroupThreadID) {
+  if (GTID.x == 0) {
+    SBufU[0] = 0u;
+    SBufI[0] = 0;
+    TBufU[0] = 0u;
+    Tex2D[uint2(0, 0)] = 0u;
+    BABuf.Store(0, 0u); // accumulator for the 2-arg method form
+    BABuf.Store(4, 0u); // accumulator for the 3-arg method form
+    BABuf.Store(8, 0u); // monotonicity counter
+  }
```
inbelic: Is this redundant? Don't all of these get initialized to 0 with the `FillBuffer`?

test/Feature/HLSLLib/InterlockedAdd.resources.32.test
```diff
@@ -0,0 +1,284 @@
...
+  // RWStructuredBuffer<int>: 3-argument free function (signed).
+  int OrigI;
+  InterlockedAdd(SBufI[0], 1, OrigI);
+  SlotsSBufI[OrigI] = 1u;
+
+  // RWBuffer<uint> (typed buffer UAV): 2-argument free function.
+  InterlockedAdd(TBufU[0], 1u);
+
+  // RWTexture2D<uint>: 2-argument free function.
+  InterlockedAdd(Tex2D[uint2(0, 0)], 1u);
```
inbelic: The two typed-UAV destinations only get the 2-argument form, and only the "final total is 256" check. Since the whole point of this PR is to bring these tests up to the established pattern, it would be good to also cover the 3-argument form on at least one typed UAV (`InterlockedAdd(TBufU[0], 1u, Orig)` plus a "seen" array), which is where a wrong `original_value` is most likely to show up on a typed resource path.

test/Feature/HLSLLib/InterlockedAdd.resources.32.test
```diff
@@ -0,0 +1,284 @@
...
+      DirectXBinding:
+        Register: 7
+        Space: 0
+      VulkanBinding:
+        Binding: 7
+...
+#--- end
+
+# Unimplemented texture store: https://github.com/llvm/llvm-project/issues/194930
+# UNSUPPORTED: Clang && (DirectX || Metal)
```
inbelic:
```suggestion
# XFAIL: Clang && (DirectX || Metal)
```
Shouldn't we use xfail for stuff that will be added so that we will see the xpass?

Also the linked issue is closed which means it should be passing now?

test/Feature/HLSLLib/InterlockedAdd.32.test
```diff
@@ -77,33 +102,61 @@ Buffers:
   - Name: OutSlotsInt
     Format: UInt32
     Stride: 4
-    FillSize: 256
+    FillSize: 1024
   - Name: ExpectedSlotsInt
     Format: UInt32
```
inbelic, line 107:
```suggestion
    Format: Int32
```
I suppose this doesn't matter so can we just have a single `ExpectedOnes` buffer to re-use? Same for other files

test/Feature/HLSLLib/InterlockedAdd.resources.32.test
```diff
@@ -0,0 +1,284 @@
...
+// (resource) destination types:
+//
+//   * RWStructuredBuffer<uint>[i]          (free function, 2-arg)
+//   * RWStructuredBuffer<int>[i]           (free function, 3-arg, signed)
+//   * RWBuffer<uint>[i]                    (free function on typed UAV)
+//   * RWTexture2D<uint>[coord]             (free function on typed UAV)
+//   * RWByteAddressBuffer::InterlockedAdd  (method, 2-arg and 3-arg)
+//
+// 256 threads each add 1 to a shared accumulator, so every accumulator must
+// end at exactly 256 and no update may be lost. The 3-argument forms are
```
inbelic: The last accumulator ends at 65536 now no?
```suggestion
// 256 threads each add 1 to a shared accumulator, so every accumulator must
// end at exactly 256 (- this should be reworded) and no update may be lost. The 3-argument forms are
```

test/Feature/HLSLLib/InterlockedAdd.int64.test
bob80905: Yeah, the comment could be improved or maybe just removed. The intent here was to communicate the loop "crosses" / encounters / deals with a counter that has data in the upper 32 bits.
So, if interlockedAdd was incorrectly implemented, and ignored the MSB, we would be able to catch it.
Regardless, I'll edit the comment.

test/Feature/HLSLLib/InterlockedAdd.int64.test
bob80905: This was an oversight, I do believe the XFAIL should be added here.

inbelic (approved):

test/Feature/HLSLLib/InterlockedAdd.32.test
```diff
@@ -1,37 +1,42 @@
...
 groupshared uint CounterUInt;         // 3-arg form, unsigned
 groupshared int  CounterIntNoOrig;    // 2-arg form, signed
 groupshared uint CounterUIntNoOrig;   // 2-arg form, unsigned
 groupshared int  SignedCounter;       // 2-arg form with negative addend
+groupshared uint MonoCounter;         // incremented in a barrier-separated loop
 
-groupshared uint SlotsInt[64];
-groupshared uint SlotsUInt[64];
+groupshared uint SlotsInt[256];
+groupshared uint SlotsUInt[256];
```
farzonl, line 29: why 256? The PR description says you are doing this but not why? Second maybe it would be worth moving this to a pre processor macross so if it changes again it is a one-line change instead of updating the loops and numthreads?

test/Feature/HLSLLib/InterlockedAdd.32.test
```diff
@@ -53,17 +58,37 @@ void main(uint3 GTID : SV_GroupThreadID) {
   // 2-argument form with negative addend (signedness test).
   InterlockedAdd(SignedCounter, -1);
 
+  // MonoCounter, checked for monotonic non-decrease and for strictly
+  // increasing per-thread originals.
+  uint Prev = 0u;
+  uint PrevOrig = 0u;
+  uint Mono = 1u;
+  for (uint I = 0; I < 256; ++I) {
```
farzonl, line 66: curious if we are doing 256 threads why do we do this loop for the I counter? can't you use the group thread id?

farzonl (approved):
LGTM

test/Feature/HLSLLib/InterlockedAdd.32.test
bob80905: 256 because of this comment: https://github.com/llvm/offload-test-suite/pull/1347#discussion_r3539912562
Essentially, maximize race condition chances, and test across waves.
I will consider doing a macro change for numthreads across all tests.

test/Feature/HLSLLib/InterlockedAdd.32.test
bob80905: Nothing about 256 is required, it was chosen to mirror the compare tests' shape and the loop exists to get a decent number of race condition possibilities. Otherwise it's arbitrary.

PR #1436, opened 2026-08-11
Migrate NVIDIA SKU over to frequent build path

bob80905:
The frequent build path has been validated and it is time to slowly migrate existing SKUs over to using the more efficient path.
This PR starts by moving the existing NVIDIA SKU default tests over to the frequent build path.
It sets up a shared resolve workflow that the legacy path and the new frequent build path can both use.

Assisted by: Github Copilot
Addresses: https://github.com/llvm/offload-test-suite/issues/1388

### DIFF
--- a/.github/workflows/resolve-frequent-builds.yaml
+++ b/.github/workflows/resolve-frequent-builds.yaml
@@ -30,7 +30,7 @@ on:
 
 jobs:
   resolve:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-26.04
     permissions:
       contents: read
       actions: read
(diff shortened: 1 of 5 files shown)


llvm-beanz (approved):
LGTM

.github/workflows/resolve-frequent-builds.yaml
```diff
@@ -0,0 +1,77 @@
...
+      LlvmRunId:
+        description: 'Resolved frequent-build-llvm.yaml run ID.'
+        value: ${{ jobs.resolve.outputs.LlvmRunId }}
+      DxcRunId:
+        description: 'Resolved frequent-build-dxc.yaml run ID.'
+        value: ${{ jobs.resolve.outputs.DxcRunId }}
+
+jobs:
+  resolve:
+    runs-on: ubuntu-24.04
```
farzonl: Interesting so the build picker doesn't need to be windows based or one of our self hosted machines. That's good to know.  Should we use a newer LTS like 26.04?

farzonl (approved):

PR #1437, opened 2026-08-12
Add `InterlockedExchange` tests

### LINKED ISSUE #854: Add test for `InterlockedExchange`
- [ ] Test for 16 bit int types (`InterlockedExchange.int16.test`)
- [ ] Test for Half type (`InterlockedExchange.fp16.test`)
- [ ] Test for 32 bit types (`InterlockedExchange.32.test`)
- [ ] Test for 64 bit int types (`InterlockedExchange.int64.test`)
- [ ] Test for Double type (`InterlockedExchange.fp64.test`)

A .test file should contain 
```
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# RUN: split-file %s %t
```

bob80905:
This PR adds the InterlockedExchange tests.
It also lights up a feature bit that is necessary for testing floats on groupshared memory using atomic operations.

Fixes https://github.com/llvm/offload-test-suite/issues/854
Assisted by: Github Copilot


lib/API/VK/Device.cpp
```diff
@@ -1601,6 +1601,33 @@ class VulkanDevice : public offloadtest::Device {
     }
 #endif
 
+#ifdef VK_EXT_SHADER_ATOMIC_FLOAT_EXTENSION_NAME
```
farzonl: why  do we need to do a preprocessor for this?

bob80905: Not strictly necessary.
This was added to mimic / stay consistent with other preprocessing guards, as seen on line 1590.
However, we use a newer macro elsewhere, unguarded (VK_EXT_mesh_shader).
So this ifdef will always be true assuming the build works.
I'll remove the guards.

test/Feature/HLSLLib/InterlockedExchange.float.test
```diff
@@ -0,0 +1,208 @@
...
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 ]
```
farzonl, line 116: Do these have to all be ones?

test/Feature/HLSLLib/InterlockedExchange.float.test
```diff
@@ -0,0 +1,208 @@
...
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
+            1 ]
```
farzonl, line 142: is it possible to make this more unique values?

test/Feature/HLSLLib/InterlockedExchange.float.test
bob80905: No, they can be arbitrary, line 42 shows where this resource is assigned.
All we need is a way to distinguish from 0. The 1 represents a flag that was set only when the expected original value is seen, and when the new value is provably written to the slot.
We want this flag set for all threads.

test/Feature/HLSLLib/InterlockedExchange.float.test
bob80905: The purpose isn't uniqueness, and these aren't inputs. It wouldn't add anything meaningful to the test to make these output flags differ between each other.

lib/API/VK/Device.cpp
```diff
@@ -1700,6 +1725,23 @@ class VulkanDevice : public offloadtest::Device {
     }
 #endif
 
+    if (HasShaderAtomicFloatExt) {
```
alsepkow, line 1721: nit: I see that we've used this same pattern elsewhere. But it seems overly verbose and a little counter-intuitive to me. We probably shouldn't stray from existing patterns in this PR. 

But a potential clean up could instead use one struct when we query support. And use another freshly initialized one to add to the EnabledDeviceExtensions list. Zero initialized by design and simply toggle the bits we want.


```VkPhysicalDeviceShaderAtomicFloatFeaturesEXT SupportedAtomicFloatFeatures{};
SupportedAtomicFloatFeatures.sType =
    VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_ATOMIC_FLOAT_FEATURES_EXT;

VkPhysicalDeviceFeatures2 ProbeFeatures{};
ProbeFeatures.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2;
ProbeFeatures.pNext = &SupportedAtomicFloatFeatures;
vkGetPhysicalDeviceFeatures2(PhysicalDevice, &ProbeFeatures);

const bool HasShaderAtomicFloatExt =
    SupportedAtomicFloatFeatures.shaderSharedFloat32Atomics ||
    SupportedAtomicFloatFeatures.shaderBufferFloat32Atomics;

VkPhysicalDeviceShaderAtomicFloatFeaturesEXT EnabledAtomicFloatFeatures{};
if (HasShaderAtomicFloatExt) {
  EnabledAtomicFloatFeatures.sType =
      VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_ATOMIC_FLOAT_FEATURES_EXT;
  EnabledAtomicFloatFeatures.shaderSharedFloat32Atomics =
      SupportedAtomicFloatFeatures.shaderSharedFloat32Atomics;
  EnabledAtomicFloatFeatures.shaderBufferFloat32Atomics =
      SupportedAtomicFloatFeatures.shaderBufferFloat32Atomics;
  EnabledAtomicFloatFeatures.pNext = Features.pNext;
  Features.pNext = &EnabledAtomicFloatFeatures;
}

File: test/Feature/HLSLLib/InterlockedExchange.float.test (hunk 3 above)
Comment (alsepkow) [reply test/Feature/HLSLLib/InterlockedExchange.float.test:142]: I had the same thoughts reflected in both of Farzons comments. Might be worth a simple comment in these tests for potential future readers.

File: test/Feature/HLSLLib/InterlockedExchange.32.test [hunk 5]
```diff
@@ -0,0 +1,248 @@
+#--- source.hlsl
+
+// Tests InterlockedExchange on groupshared destinations, int and uint, with
+// 256 concurrent threads. Four checks:
+//
+//   1. Private slots: the exchange reports the slot's known initial value
```
alsepkow, line 6: nit: Didn't check what you did on the last PR but I'd find these comments more helpful directly above the actual test cases

alsepkow: Same apples for other test files

alsepkow (approved):
LGTM. Some nits that you don't need to address this PR.

lib/API/VK/Device.cpp
```diff
@@ -1700,6 +1725,23 @@ class VulkanDevice : public offloadtest::Device {
...
+      FeaturesAtomicFloat.shaderBufferFloat32AtomicAdd = 0;
+      FeaturesAtomicFloat.shaderBufferFloat64Atomics = 0;
+      FeaturesAtomicFloat.shaderBufferFloat64AtomicAdd = 0;
+      FeaturesAtomicFloat.shaderSharedFloat32AtomicAdd = 0;
+      FeaturesAtomicFloat.shaderSharedFloat64Atomics = 0;
+      FeaturesAtomicFloat.shaderSharedFloat64AtomicAdd = 0;
+      FeaturesAtomicFloat.shaderImageFloat32Atomics = 0;
+      FeaturesAtomicFloat.shaderImageFloat32AtomicAdd = 0;
+      FeaturesAtomicFloat.sparseImageFloat32Atomics = 0;
+      FeaturesAtomicFloat.sparseImageFloat32AtomicAdd = 0;
```
farzonl: is there not like a default constructor that could do this for us. seems weird we have to null out  in a conditional.

test/Feature/HLSLLib/InterlockedExchange.32.test
alsepkow: Apples was supposed to be applies. But it kind of still works! 🍎

lib/API/VK/Device.cpp
alsepkow: Thought the same, but I think the problem is that the query can potentially populate these fields. Hence my earlier comment here: https://github.com/llvm/offload-test-suite/pull/1437#discussion_r3917808172

farzonl (approved):
I think Alex made some good points about the code in VK Device.cpp.

Tests look good. VK Device implementation is fine too. More questions than blockers.

PR #1439, opened 2026-08-12
Repair failing Interlocked Xor/Or/Add tests

bob80905:
This PR repairs a few failing InterlockedXor/Or/Add tests.
An issue has been made to track the typed resource tests: https://github.com/llvm/llvm-project/issues/217823
It was further discovered that the https://github.com/llvm/llvm-project/issues/217831 issue, and the new 1438 issue above, should be GPU agnostic, since this is a compile time error.
Another prevalent issue was new implementations being added to llvm main, which caused .32 tests to XPASS. We should remove the "unimplemented" XFails.

Assisted by: Github Copilot

### DIFF
--- a/test/Feature/HLSLLib/InterlockedOr.resources.int64.test
+++ b/test/Feature/HLSLLib/InterlockedOr.resources.int64.test
@@ -168,7 +168,7 @@ DescriptorSets:
 #--- end
 
 # Bug: https://github.com/llvm/offload-test-suite/issues/1407
-# XFAIL: Intel && Clang && Vulkan
+# XFAIL: (Intel || AMD) && Clang && Vulkan
 
 # Bug: https://github.com/llvm/offload-test-suite/issues/1164
 # XFAIL: Metal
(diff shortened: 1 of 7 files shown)


test/Feature/HLSLLib/InterlockedOr.resources.int64.test
```diff
@@ -168,7 +168,7 @@ DescriptorSets:
 #--- end
 
 # Bug: https://github.com/llvm/offload-test-suite/issues/1407
-# XFAIL: Intel && Clang && Vulkan
+# XFAIL: (Intel || AMD) && Clang && Vulkan
```
bogner: Shouldn't this be `Clang && Vulkan`? As you point out in the PR description this is a compile time error, and it does indeed fail on nvidia and qc: https://github.com/llvm/offload-test-suite/actions/runs/31993778381/job/95283314781, https://github.com/llvm/offload-test-suite/actions/runs/32028148516/job/95386777848

bob80905: Yeah, I think you're right. I seem to recall those other SKUS passing, but maybe there's a separate cause.
I'll make the change and see what happens on this run.

bogner (approved):

Icohedron (approved):

PR #1442, opened 2026-08-14
[LongVec] Add abs signed int test

farzonl:
resolves ##1441

Tests different vector sizes.


test/Feature/HLSLLib/abs.long-vector.32.test
```diff
@@ -0,0 +1,79 @@
...
+      DirectXBinding:
+        Register: 1
+        Space: 0
+      VulkanBinding:
+        Binding: 1
+...
+#--- end
+
+# RUN: split-file %s %t
+# RUN: %dxc_target -T %if Clang %{cs_6_8%} %else %{cs_6_9%} -Fo %t.o %t/source.hlsl
```
llvm-beanz: Should we be testing this with both 6.0 and 6.9 through Clang?

I know DXC requires 6.9, but in theory this works with clang for 6.0.

farzonl: That's a good point.

test/Feature/HLSLLib/abs.long-vector.32.test
```diff
@@ -0,0 +1,79 @@
+#--- source.hlsl
+
+StructuredBuffer<int> In : register(t0);
+RWStructuredBuffer<int> Out : register(u1);
```
farzonl, line 4: We don't have long vector support in clang for buffers yet that's why we are not passing them in here and are using scalar forms instead.

test/Feature/HLSLLib/abs.long-vector.32.test
```diff
@@ -0,0 +1,85 @@
...
+      VulkanBinding:
+        Binding: 1
+...
+#--- end
+
+# Unimplemented https://github.com/llvm/wg-hlsl/issues/467
+# XFAIL: Vulkan
+
+# CI machines don't have SM 6.9 support yet
+# REQUIRES: !DXC || SM_6_9
```
farzonl: Remove this now that offloader can support SM_6_9 enums.
```suggestion
```

test/Feature/HLSLLib/abs.long-vector.32.test
```diff
@@ -0,0 +1,85 @@
+#--- source.hlsl
```
kmpeng, line 1: Should we rename to just `abs.long-vector.test` since there are no 16-bit/64-bit tests?

test/Feature/HLSLLib/abs.long-vector.32.test
```diff
@@ -0,0 +1,85 @@
...
+#--- end
+
+# Unimplemented https://github.com/llvm/wg-hlsl/issues/467
+# XFAIL: Vulkan
+
+# CI machines don't have SM 6.9 support yet
+# REQUIRES: !DXC || SM_6_9
+
+# RUN: split-file %s %t
+# RUN: %dxc_target -T %if Clang %{cs_6_0%} %else %{cs_6_9%} -Fo %t.o %t/source.hlsl
```
kmpeng, line 81: Chris's comment suggested testing both 6.0 and 6.9 through Clang, but this only tests 6.0. Not sure if it's necessary to test both, just pointing it out.

kmpeng (approved):
LGTM assuming everything passes after removing the `REQUIRES`.

test/Feature/HLSLLib/abs.long-vector.32.test
farzonl: We can't test 6.9 through clang. we haven't done vectorized DXIL yet.

test/Feature/HLSLLib/abs.long-vector.32.test
farzonl: There are a few intrinsics that have special 16\64 bit cases and so while this name isn't very distinct for abs could be useful for those other intrinsics.

bob80905 (approved):

PR #1447, opened 2026-08-18
[Matrix] Add comparison operator tests

### LINKED ISSUE #1446: [Matrix] Add elementwise comparison operator tests
This should be very similar to the matrix logical operator tests.

farzonl:
resolves https://github.com/llvm/offload-test-suite/issues/1446

cover cmp operators across matrix sizes and base types

assisted by co-pilot MAI model


test/Basic/Matrix/matrix_comparison_operators.fp16.test
```diff
@@ -0,0 +1,64 @@
...
+  - Name: HalfIn
+    Format: Float16
+    Data: [ 0x4000, 0x4000 ]
+    #       2.0,    2.0
+  - Name: BoolOut
+    Format: Int32
+    FillSize: 48
+  - Name: BoolExpected
+    Format: Int32
+    Data: [ 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 ]
```
kmpeng: (For all tests)
We're only ever testing the true path for these comparisons. Is it worth picking operands that give a mix of true and false so we cover the false path too?

kmpeng (approved):

test/Basic/Matrix/matrix_comparison_operators.fp16.test
```diff
@@ -0,0 +1,72 @@
+#--- source.hlsl
+
+RWStructuredBuffer<half> HalfIn : register(u0);
+RWStructuredBuffer<bool> BoolOut : register(u1);
+
+[numthreads(1, 1, 1)]
+void main() {
+  half4x3 HalfEQLHS = HalfIn[0], HalfEQRHS = HalfIn[1];
```
Icohedron, line 8: [SUGGESTION] You should do the same as commit 8ebdcbd and make all the tests use unique matrix elements to confirm that the compares are actually done elementwise. Otherwise the matrix comparison collapses to a single scalar fcmp instruction.

test/Basic/Matrix/matrix_comparison_operators.fp32.test
```diff
@@ -0,0 +1,105 @@
...
+      Kind: RWStructuredBuffer
+      DirectXBinding:
+        Register: 1
+        Space: 0
+      VulkanBinding:
+        Binding: 1
+...
+#--- end
+
+# Unimplemented https://github.com/microsoft/DirectXShaderCompiler/issues/8796
```
Icohedron, line 100: [QUESTION] I'm wondering why these tests don't carry the same platform exclusion as the bool-matrix tests.

matrix_bool_and_operator.test, matrix_bool_and_operator_thread_group.test and
matrix_bool_or_operator.test all end with:
```
  # Note: Metal does not support boolean matrix types. For
  # Vulkan KosmicKrisp and MoltenVK the output is always False.
  # UNSUPPORTED: Metal || KosmicKrisp || MoltenVK
```
Every test in this PR produces `boolNxM` values from the comparison and indexes them.

test/Basic/Matrix/matrix_comparison_operators.int64.test
```diff
@@ -0,0 +1,71 @@
+#--- source.hlsl
+
+RWBuffer<int64_t> Int64In : register(u0);
```
Icohedron, line 3: [ISSUE] `RWBuffer<int64_t>` has a known issue in Lavapipe.

The existing typed 64-bit test, test/Feature/TypedBuffer/64bit-scalar.test has this XFAIL:
```
  # Driver bug https://gitlab.freedesktop.org/mesa/mesa/-/work_items/15866
  # XFAIL: Lavapipe
```
So either switch to a `RWStructuredBuffer<uint64_t>` or add the same XFAIL.

test/Basic/Matrix/matrix_comparison_operators.uint32.test
```diff
@@ -0,0 +1,104 @@
...
+#--- pipeline.yaml
+
+---
+Shaders:
+  - Stage: Compute
+    Entry: main
+Buffers:
+  - Name: UintIn
+    Format: UInt32
+    Data: [ 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16 ]
```
Icohedron, line 57: [SUGGESTION] The int and uint tests should exercise values where the sign bit is actually set. As-is, there is no observable difference if the compiler used signed/unsigned operations where it should have been unsigned/signed.

Icohedron (commented):
I don't see why the 32-bit tests have to be the only comprehensive ones. 
The 32-bit tests aren't even that much bigger than the other tests in line count either.

e.g. the fp16 test is 72 lines while the fp32 test is 105 lines

test/Basic/Matrix/matrix_comparison_operators.fp32.test
farzonl: The failures don't happen when you produce a boolean matrix only when you try and do an operation on them. At least for metal. I have not tested the mac vulkan runtimes.

test/Basic/Matrix/matrix_comparison_operators.fp16.test
farzonl: We did a group review of this pr on this issue. There is no type specific codegen for comparisons, so the feeling was we didn't need to be as comprehensive.  That said I don't feel strongly about this it was just easier to spot check the cmps were correct if I could look at a smaller set of values to make a determination.

test/Basic/Matrix/matrix_comparison_operators.int64.test
farzonl: Thanks will add an XFAIL was not aware of this Lavapipe limitation.

alsepkow: Looks like uint64 will also need an XFAIL for the same issue.

alsepkow (commented):
LGTM assuming @Icohedron comments are addressed.
And looks like there's a legalization issue causing a crash in the 32 bit tests?

test/Basic/Matrix/matrix_comparison_operators.int64.test
farzonl: So I ran  lavapipe runs and did not see these tests in the failure list

1. [windows-amd, check-hlsl-clang-vk-lavapipe ](https://github.com/llvm/offload-test-suite/actions/runs/33112705334/job/98680751803?pr=1447) Failed Tests (1):
  OffloadTest-clang-vk :: Graphics/SimpleTriangle.test
2. [windows-amd, check-hlsl-vk-lavapipe](https://github.com/llvm/offload-test-suite/actions/runs/33112705334/job/98673774866?pr=1447) Failed Tests (1):
  OffloadTest-vk :: Graphics/SimpleTriangle.test
3. [windows-qc, check-hlsl-vk-lavapipe](https://github.com/llvm/offload-test-suite/actions/runs/33112705334/job/98673290135?pr=1447) Failed Tests (3):
  OffloadTest-vk :: Feature/InlineRT/aabb-procedural.test
  OffloadTest-vk :: Feature/InlineRT/geometry-transform.test
  OffloadTest-vk :: Feature/InlineRT/tlas-array.test

4. [windows-sw-rasterizer-x86_64, check-hlsl-clang-vk-lavapipe](https://github.com/llvm/offload-test-suite/actions/runs/33112705334/job/98669654804?pr=1447) fatal: No target device found! Searched matching regex 'llvmpipe'. Devices reported by api-query: NVIDIA GeForce RTX 3070 Ti, Microsoft Basic Render Driver, NVIDIA GeForce RTX 3070 Ti
5. [windows-sw-rasterizer-x86_64, check-hlsl-vk-lavapipe](https://github.com/llvm/offload-test-suite/actions/runs/33112705334/job/98663063657?pr=1447) fatal: No target device found! Searched matching regex 'llvmpipe'. Devices reported by api-query: NVIDIA GeForce RTX 3070 Ti, Microsoft Basic Render Driver, NVIDIA GeForce RTX 3070 Ti

Icohedron: I suppose if it's not failing then you can just leave out the XFAIL. 
Might be worth replying to the bug report that it doesn't fail on all shaders with `RWBuffer<int64_t>`.

Icohedron (approved):
Would be good to have uint tests exercise extra large integers which have the most significant bit set to be sure it's not interpreted as a negative signed integer instead.

But otherwise LGTM

PR #1448, opened 2026-08-18
Update a few out of date XFAILs

bogner:
These are a few tests that need updates here:

1. These Matrix tests work on DirectX after llvm/llvm-project#211346, since they no longer end up with a constant PHI node.
2. The Matrix groupthread swizzle tests now fail on Vulkan in the same way as the non-groupthread versions.
3. The Textures tests now work after llvm/llvm-project#211256, since the intrinsics are now correctly marked as convergent.


llvm-beanz (approved):

Icohedron (approved):
Seems correct to me

PR #1450, opened 2026-08-18
Add `InterlockedCompareExchangeFloatBitwise` tests

### LINKED ISSUE #851: Add test for `InterlockedCompareExchangeFloatBitwise`
- [ ] Test for 16 bit int types (`InterlockedCompareExchangeFloatBitwise.int16.test`)
- [ ] Test for Half type (`InterlockedCompareExchangeFloatBitwise.fp16.test`)
- [ ] Test for 32 bit types (`InterlockedCompareExchangeFloatBitwise.32.test`)
- [ ] Test for 64 bit int types (`InterlockedCompareExchangeFloatBitwise.int64.test`)
- [ ] Test for Double type (`InterlockedCompareExchangeFloatBitwise.fp64.test`)

A .test file should contain 
```
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# RUN: split-file %s %t
```

bob80905:
This PR adds the `InterlockedCompareExchangeFloatBitwise` tests.
Since the function does not have any 64 bit overloads, there are fewer tests, but they should cover all use cases of this function.

Resolves: https://github.com/llvm/offload-test-suite/issues/851
Assisted by: Github Copilot

### DIFF
--- a/test/Feature/HLSLLib/InterlockedCompareExchangeFloatBitwise.32.test
+++ b/test/Feature/HLSLLib/InterlockedCompareExchangeFloatBitwise.32.test
@@ -2,7 +2,7 @@
 
 // Tests InterlockedCompareExchangeFloatBitwise on groupshared destinations.
 // 256 threads run at the same time. Each test value is exact in float, thus
-// the shader compares floats directly. There are four checks:
+// the shader compares floats directly. There are three checks:
 //
 //   1. Private slots. A compare that does not match must not store. A later
 //      compare that matches must store. Both report the initial value.
@@ -14,24 +14,17 @@
 //   3. Bit patterns. +0.0 against -0.0 must not store. A NaN against the
 //      same NaN bits must store. A buffer supplies the NaN, because HLSL
 //      has no NaN literal.
-//   4. CounterZero. Thread `i` moves the counter from `i` to `i+1` only.
-//      The shader makes 256 attempts, with a barrier between them. Each
-//      thread wins one time, the counter stops at 256, and the counter must
-//      never decrease.
 
-RWStructuredBuffer<uint> OutMono : register(u0);
-RWStructuredBuffer<uint> OutSlots : register(u1);
-RWStructuredBuffer<uint> OutOrig : register(u2);
-RWStructuredBuffer<uint> OutWins : register(u3);
-RWStructuredBuffer<float> OutFinal : register(u4);
-RWStructuredBuffer<float> NanIn : register(u5);
+RWStructuredBuffer<uint> OutSlots : register(u0);
+RWStructuredBuffer<uint> OutOrig : register(u1);
+RWStructuredBuffer<float> OutFinal : register(u2);
+RWStructuredBuffer<float> NanIn : register(u3);
 
 groupshared float SlotsMinusTwoPointFive[256]; // per-thread slots, each -2.5
 groupshared float MatchMinusFive;              // contended, compare matches
 groupshared float NoMatchSeventySeven;         // contended, never matches
 groupshared float ZeroPositive;                // +0.0, compared against -0.0
 groupshared float NanQuiet;                    // NaN, compared against NaN
-groupshared float CounterZero;                 // stepped from 0.0 to 256.0
 groupshared float ClaimZero;                   // 0.0, claimed by the winner
 
 [numthreads(256, 1, 1)]
@@ -42,7 +35,6 @@ void main(uint3 GTID : SV_GroupThreadID) {
     NoMatchSeventySeven = 77.0f;
     ZeroPositive = 0.0f;
     NanQuiet = NanIn[0];
-    CounterZero = 0.0f;
     ClaimZero = 0.0f;
   }
   GroupMemoryBarrierWithGroupSync();
@@ -84,36 +76,14 @@ void main(uint3 GTID : SV_GroupThreadID) {
                      OrigNoMatch == 77.0f && OrigZero == 0.0f &&
                      Claim == 0.0f) ? 1u : 0u;
 
-  // Check 4.
-  float CompareValue = (float)GTID.x;
-  float NewValue = (float)GTID.x + 1.0f;
-  float Prev = 0.0f;
-  uint Mono = 1u;
-  uint Wins = 0u;
-  for (uint I = 0; I < 256; ++I) {
-    float Orig;
-    InterlockedCompareExchangeFloatBitwise(CounterZero, CompareValue, NewValue,
-                                           Orig);
-    if (Orig == CompareValue)
-      ++Wins;
-    GroupMemoryBarrierWithGroupSync();
-    float Cur = CounterZero;
-    if (Cur < Prev)
-      Mono = 0u;
-    Prev = Cur;
-  }
-  OutMono[GTID.x] = Mono;
-  OutWins[GTID.x] = Wins;
-
   GroupMemoryBarrierWithGroupSync();
 
   if (GTID.x == 0) {
-    OutFinal[0] = CounterZero;         // 256.0
-    OutFinal[1] = MatchMinusFive;      // 42.0
-    OutFinal[2] = NoMatchSeventySeven; // 77.0
-    OutFinal[3] = ZeroPositive;        // +0.0, -0.0 must not have matched
-    OutFinal[4] = NanQuiet;            // 7.0, the NaN must have matched
-    OutFinal[5] = ClaimZero;           // 1.0, exactly one thread claimed
+    OutFinal[0] = MatchMinusFive;      // 42.0
+    OutFinal[1] = NoMatchSeventySeven; // 77.0
+    OutFinal[2] = ZeroPositive;        // +0.0, -0.0 must not have matched
+    OutFinal[3] = NanQuiet;            // 7.0, the NaN must have matched
+    OutFinal[4] = ClaimZero;           // 1.0, exactly one thread claimed
   }
 }
 
@@ -124,10 +94,6 @@ Shaders:
   - Stage: Compute
     Entry: main
 Buffers:
-  - Name: OutMono
-    Format: UInt32
-    Stride: 4
-    FillSize: 1024
   - Name: ExpectedOnes
     Format: UInt32
     Stride: 4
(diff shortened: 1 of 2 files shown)


test/Feature/HLSLLib/InterlockedCompareExchangeFloatBitwise.32.test
```diff
@@ -0,0 +1,242 @@
...
+//      matching compare must, and both must report the original value.
+//   2. Contended locations: every thread proposes the same new value, so
+//      exactly one thread may observe the original compare value. A
+//      never-matching compare must leave the destination alone.
+//   3. Bitwise semantics: +0.0 compared against -0.0 must not store even
+//      though the two are numerically equal, and a NaN compared against an
+//      identical NaN bit pattern must store even though NaN != NaN.
+//   4. CounterZero: thread `i` only advances the counter from `i` to `i+1`,
+//      so over 256 barrier-separated attempts each thread succeeds exactly
+//      once and the counter ends at 256. It must never decrease.
```
farzonl: This is very verbose.

test/Feature/HLSLLib/InterlockedCompareExchangeFloatBitwise.32.test
```diff
@@ -0,0 +1,242 @@
...
+    NoMatchSeventySeven = 77.0f;
+    ZeroPositive = 0.0f;
+    NanQuiet = asfloat(NanBits);
+    CounterZero = 0.0f;
+    MatchWinCountZero = 0u;
+    NanWinCountZero = 0u;
+  }
+  GroupMemoryBarrierWithGroupSync();
+
+  // Check 1: both calls report the initial value, only the second stores.
```
farzonl: There is a bunch of stuff we are doing in this file that just seems a bit too extra for testing `InterlockedCompareExchangeFloatBitwise`

bob80905: What specifically are you suggesting I remove?
I think each of the checks have merit.

test/Feature/HLSLLib/InterlockedCompareExchangeFloatBitwise.32.test
bob80905: I think the content justifies the length. The test file is long and its methods aren't very self-evident, so this header comment does help clear things up.

test/Feature/HLSLLib/InterlockedCompareExchangeFloatBitwise.32.test
farzonl: You are supposed to be testing `InterlockedCompareExchangeFloatBitwise` but InterlockedAdd is all over the place. It seems like the thing you really care about for this test is thread synchronization and barriers and you are using `InterlockedCompareExchangeFloatBitwise` a pretext for doing that but the result is you don't have a very independent test for `InterlockedCompareExchangeFloatBitwise`. 

You also are doing a stress test of the GPU's memory model and cache coherency which I don't know is what a `InterlockedCompareExchangeFloatBitwise` should be doing. Different IHVs could do different things and the thing that makes me the most concerned about that is you are doing `WARP && DXC` as an XFAIL when my feeling is that should almost always work for everything we are testing. 

Finally, you have a heavy reliance on `asuint()` and `asfloat()` to bitcast values back and forth and this isn't supposed to be a test for those intrinsics so seeing them here means if we have regressions elsewhere these tests go down with it.

test/Feature/HLSLLib/InterlockedCompareExchangeFloatBitwise.32.test
farzonl: I get the feeling you are trying to test a use case instead of a feature and that is a bit problematic because we aren't able to isolate the behavior of an intrinsic if you do that. We have use case test like particle life, but this shouldn't be that.

test/Feature/HLSLLib/InterlockedCompareExchangeFloatBitwise.32.test
bob80905: Agreed with point 1 and 3. For point 2, I do think *some* stress testing is necessary, as part of proving the interlocked intrinsics function correctly is to allow for opportunities for them to clobber each other.
This requires multiple simultaneously executing threads reading/writing a value.
A correctly implemented interlocked intrinsic will always pass these tests, to my knowledge, regardless of memory model / cache coherency, correct me if I'm wrong.
But I agree on your other points, and have removed the 2 interlockedAdds, and the asuint/asfloat calls.

test/Feature/HLSLLib/InterlockedCompareExchangeFloatBitwise.resources.32.test
```diff
@@ -0,0 +1,300 @@
...
+#--- end
+
+# Unimplemented: https://github.com/llvm/llvm-project/issues/99203
+# XFAIL: Clang
+
+# Unimplemented: https://github.com/microsoft/DirectXShaderCompiler/issues/7782
+# XFAIL: Vulkan && DXC
+
+# Bug: https://github.com/llvm/offload-test-suite/issues/1418
+# UNSUPPORTED: WARP && DXC
```
farzonl: Is it possible to write this test for cs_6_5 to avoid this issue entirely? I feel uncomfortable that we can't pass warp when it is a tier 1 execution environment for us.

bob80905: cs_6_5 also hangs.

test/Feature/HLSLLib/InterlockedCompareExchangeFloatBitwise.resources.32.test
```diff
@@ -0,0 +1,300 @@
...
+      VulkanBinding:
+        Binding: 7
+...
+#--- end
+
+# Unimplemented: https://github.com/llvm/llvm-project/issues/99203
+# XFAIL: Clang
+
+# Unimplemented: https://github.com/microsoft/DirectXShaderCompiler/issues/7782
+# XFAIL: Vulkan && DXC
```
damyanp, line 215: This issue is closed...

damyanp (dismissed):

test/Feature/HLSLLib/InterlockedCompareExchangeFloatBitwise.32.test
damyanp: Agree with Farzon here - we're not trying to write exhaustive driver conformance tests here.

test/Feature/HLSLLib/InterlockedCompareExchangeFloatBitwise.resources.32.test
bob80905: Reopened! My oversight, I thought it was open.

test/Feature/HLSLLib/InterlockedCompareExchangeFloatBitwise.32.test
bob80905: Understandable, I still think an atomicity check belongs in this test, but I suppose we could drop check 4 since it does depend on synchronization and the memory model. 
I'll remove check 4 from both files, which will get rid of the warp hang xfail.

damyanp: Steps 1 - 3 all sound very much like driver conformance tests.  The goal of these tests is to verify that the compiler is generating correct code.  Complex testing like this seems to be starting from an assumption that the compiler generates correct code with the goal of teasing out bugs in the driver's backend compiler.  The cost of this is extremely complicated tests that are very difficult to review and debug, which actually means that these tests aren't really fit for the purpose we need them for.

There may be a place for more adversarial use-case driven tests in the offload test suite, but I'm not sure that this is it, and I'm not sure that we should be focusing on them at this point.

bob80905: I would push back on this:
Steps 1-2 are conventional at this point, and step 3 is new since this is the first bitwise comparison test I am implementing.
We test that the generated code behaves as expected during execution, and to do so we test the basic functionality of any interlocked builtin, regardless of driver implementation: Is it atomic? If it's a compare type, does it become a no-op when the compare is false? And now a new question: If comparison is bitwise, then will it become a no-op when bitwise comparison is unequal even if value comparison is equal?
These don't strike me as driver conformance tests, but tests that prove the builtin does what it says it does.
All previous interlocked tests at the minimum test atomicity, so if you're suggesting to remove atomicity testing here, it would be in stark contrast to *all* previous interlocked offload tests.

test/Feature/HLSLLib/InterlockedCompareExchangeFloatBitwise.resources.32.test
farzonl: Ah ok. I think the issue needs to be rewritten because it implies the code works for sm 6.5

bob80905: You're right. I think I ran the 6.5 repro without the validation layer flag, but it also hangs with this option. I edited the issue.

farzonl (approved):
This does seem to be much more focused on InterlockedCompareExchangeFloatBitwise. Thank you for the changes, LGTM.

PR #1452, opened 2026-08-18
Use the Direct3D 12 AgilitySDK on Windows

damyanp:
The patch arranges it so the offloader and api-query tools run using the D3D12 Agility SDK, as download from a nuget package.  A LKG is specified, but other versions can be requested via cmake configuration options.

Functionality shared with the code that obtains the WARP nuget has been factored out.

Assisted-by: Copilot


cmake/modules/AgilitySDK.cmake
```diff
@@ -0,0 +1,125 @@
...
+  endif()
+
+  if (version STREQUAL "System")
+    set_property(GLOBAL PROPERTY OFFLOADTEST_AGILITY_SDK_SYSTEM TRUE)
+    return()
+  endif()
+
+  guess_nuget_arch(nuget_arch)
+  if (nuget_arch STREQUAL "x86")
+    set(agility_arch "win32")
```
V-FEXrt, line 15: This feels like a silly question, but where do we actually gate that we are only using this on windows and not trying to do it on linux?

V-FEXrt (approved):

cmake/modules/AgilitySDK.cmake
damyanp: The `if (NOT WIN32) return()` on line 4 does that :)

V-FEXrt: oh yep! Sorry that i missed that 😭

alsepkow (approved):

farzonl (approved):

PR #1454, opened 2026-08-19
[ci_results] Work around `gh` stale data issues

bogner:
Calling `gh` has started to return data from a month or so ago fairly frequently. Trying again seems to help, and experimentally it generally only takes a couple of tries to get the correct data.

This is presumably a bug in github's API, but it makes it very hard to use the ci_results script. Implement a workaround that just tries over repeatedly.


Icohedron (approved):

damyanp (approved):
FWIW I've had to add steps to my github graphql queries to spot if the total number of items expected changes while I'm querying it.  Otherwise I can end up with duplicate items in the results.

PR #1456, opened 2026-08-19
[ci_results] Use strptime instead of fromisoformat

bogner:
Apparently [fromisoformat] got smarter in python 3.11. Use strptime for compatibility with older python, since we know the exact format of github's dates anyway.

[fromisoformat]:
  https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat


Icohedron (approved):

kmpeng (approved):

kmpeng (commented):
LGTM, just fix the formatting?

PR #1463, opened 2026-08-20
[LongVec] Add Elementwise execution tests

### LINKED ISSUE #1462: [LongVec] Add elementwise test cases
Much of these are following a format so we can try and knock many of them out at once. This first batch of tests should cover: 

- asdouble
- asfloat
- asin
- asint
- asint16
- asuint
- asuint16
- atan
- atan2
- ceil
- clamp
- clip
- cos
- cosh
- countbits
- ddx-coarse
- ddx-fine
- ddy-coarse
- ddy-fine
- degrees
- exp
- exp2
- f16tof32
- f32tof16
- firstbithigh
- firstbitlow
- floor
- fma
- fmod
- frac
- frexp
- isinf
- isnan
- ldexp
- lerp
- log
- log10
- log2
- mad
- max
- min
- modf
- pow
- radians
- rcp
- reversebits
- round
- rsqrt
- saturate
- select
- sign
- sin
- sinh
- smoothstep
- sqrt
- step
- tan
- tanh
- trunc

farzonl:
resolves https://github.com/llvm/offload-test-suite/issues/1462

Assisted with GPT-5.6 Sol via Copilot


test/Feature/HLSLLib/long-vector/ddy-coarse.long-vector.32.test
```diff
@@ -0,0 +1,106 @@
...
+
+# CI machines don't have SM 6.9 support yet
+# REQUIRES: !DXC || SM_6_9
+
+# RUN: split-file %s %t
+# RUN: %dxc_target -T vs_6_0 -Fo %t-vertex.o %t/vertex.hlsl
+# RUN: %dxc_target -T %if Clang %{ps_6_0%} %else %{ps_6_9%} -Fo %t-pixel.o %t/pixel.hlsl
+# RUN: %offloader %t/pipeline.yaml %t-vertex.o %t-pixel.o | FileCheck %s
+
+# CHECK: Name: Output
```
bob80905: I don't think it is conventional to run filecheck in the offload test suites, and buffer comparisons are preferred. Is there a reason you use filecheck here?

test/Feature/HLSLLib/f32tof16.long-vector.32.test
```diff
@@ -0,0 +1,86 @@
...
+        Space: 0
+      VulkanBinding:
+        Binding: 1
+...
+#--- end
+
+# Unimplemented https://github.com/llvm/wg-hlsl/issues/467
+# XFAIL: Vulkan
+
+# CI machines don't have SM 6.9 support yet
```
bob80905: Isn't this only true for metal at the moment?

bob80905 (commented):
Is this the complete set of elementwise intrinsics, or are there more to come?
If there's going to be many more to come, I wonder if it would be better to find a way to consolidate and parameterize these tests, they are quite repetitive.

test/Feature/HLSLLib/long-vector/ddy-coarse.long-vector.32.test
farzonl: Yes because the ddy\ddx tests are pixel shader specific so we are not writing to a buffer the way our compute shaders do. I could potentially try and use gold images instead and do image compares via imgdiff. Thats how many of the other ddy\ddx testing has been done, but that is a bit more painful to test locally.

test/Feature/HLSLLib/f32tof16.long-vector.32.test
farzonl: These lines don't have to be removed, but can be removed now. essentially if you don't have a new enough agility sdk then this requies allows us to skip the tests. CI has been updated so only value is for folks that don't have latest sdk setup to not get a bunch of test fails. But thats probably not something we want to encourage for users testing their code via this project.

test/Feature/HLSLLib/long-vector/ddy-coarse.long-vector.32.test
farzonl: I thought about this a bit more and my plan to replace these checks with `imgdiff` did not workout. Long vector usage on an image is hard to wrap my brain around. I think you are right it will just be easier to do buffer comparisons and avoid file check all together.

test/Feature/HLSLLib/long-vector/acos.long-vector.32.test
```diff
@@ -0,0 +1,83 @@
...
+RWStructuredBuffer<float> Out : register(u1);
+
+[numthreads(1,1,1)]
+void main() {
+  vector<float, 5> Input0_5;
+  for (uint I = 0; I < 5; ++I)
+    Input0_5[I] = In0[0 + I];
+  vector<float, 5> Result5 = acos(Input0_5);
+  for (uint I = 0; I < 5; ++I)
+    Out[0 + I] = Result5[I];
```
bogner, line 13: We could save a bit of typing here if we use `ByteAddressBuffer` rather than `StructuredBuffer`. Consider:
```hlsl
  vector<float, 5> Input0_5 = In0.Load< vector<float, 5> >(0);
  vector<float, 5> Result5 = acos(Input0_5);
  Out.Store< vector<float, 5> >(0, Result5);
```

or even:
```hlsl
Out.Store< vector<float, 5> >(0, acos(In0.Load< vector<float, 5> >(0)));
```

bogner (commented):
Do we need to test multiple sizes of vector for every single intrinsic? I think a single call to each elementwise intrinsic should be sufficient to prove that they work. Having just one of these tests cover the separate 5/8/16 cases seems sufficient to me.

test/Feature/HLSLLib/long-vector/acos.long-vector.32.test
farzonl: I really like your suggestion, however while less code this generates newer DirectX instructions to handle loads and stores 

If I do:
```hlsl
ByteAddressBuffer In0 : register(t0);
RWByteAddressBuffer Out : register(u1);

[numthreads(1,1,1)]
void main() {
  Out.Store< vector<float, 16> >(0, asfloat(In0.Load< vector<uint, 16> >(0)));
}
```

I generate https://hlsl.godbolt.org/z/hz846Tb4b
```llvm
define void @main() {
  %Out_UAV_rawbuf = call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind { i32 1, i32 1, i32 0, i8 1 }, i32 1, i1 false), !dbg !32 ; line:6 col:45  ; CreateHandleFromBinding(bind,index,nonUniformIndex)
  %In0_texture_rawbuf = call %dx.types.Handle @dx.op.createHandleFromBinding(i32 217, %dx.types.ResBind zeroinitializer, i32 0, i1 false), !dbg !32 ; line:6 col:45  ; CreateHandleFromBinding(bind,index,nonUniformIndex)
  %1 = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle %In0_texture_rawbuf, %dx.types.ResourceProperties { i32 11, i32 0 }), !dbg !32 ; line:6 col:45  ; AnnotateHandle(res,props)  resource: ByteAddressBuffer
  %RawBufferVectorLoad = call %dx.types.ResRet.v16i32 @dx.op.rawBufferVectorLoad.v16i32(i32 303, %dx.types.Handle %1, i32 0, i32 undef, i32 4), !dbg !32 ; line:6 col:45  ; RawBufferVectorLoad(buf,index,elementOffset,alignment)
  %2 = extractvalue %dx.types.ResRet.v16i32 %RawBufferVectorLoad, 0, !dbg !32 ; line:6 col:45
  %3 = bitcast <16 x i32> %2 to <16 x float>, !dbg !33 ; line:6 col:37
  %4 = call %dx.types.Handle @dx.op.annotateHandle(i32 216, %dx.types.Handle %Out_UAV_rawbuf, %dx.types.ResourceProperties { i32 4107, i32 0 }), !dbg !34 ; line:6 col:3  ; AnnotateHandle(res,props)  resource: RWByteAddressBuffer
  call void @dx.op.rawBufferVectorStore.v16f32(i32 304, %dx.types.Handle %4, i32 0, i32 undef, <16 x float> %3, i32 4), !dbg !34 ; line:6 col:3  ; RawBufferVectorStore(uav,index,elementOffset,value0,alignment)
  ret void, !dbg !35 ; line:7 col:1
}
```
The problematic instructions being `@dx.op.rawBufferVectorLoad.v16i32` and `@dx.op.rawBufferVectorStore.v16f32`.

However if I do 
```hlsl
StructuredBuffer<uint> In0 : register(t0);

[numthreads(1,1,1)]
void main() {
  vector<uint, 16> Input0_16;
  for (uint I = 0; I < 16; ++I)
    Input0_16[I] = In0[13 + I];
  vector<float, 16> Result16 = asfloat(Input0_16);
  for (uint I = 0; I < 16; ++I)
    Out[13 + I] = Result16[I];
}
```
we get a vector bitcast and a vector load store https://hlsl.godbolt.org/z/Ydonf11ra

The important part of these tests being the ` %3 = bitcast <16 x i32> %2 to <16 x float>` I wonder if it is worth reducing the hlsl code if it forces us to have to xfail Metal for DXC?

farzonl: We discussed offline. @bogner suggested just xfailing metal. I don't love this as metal is more convenient for me to test this feature on my local machine when using DXC and DXC does not support SPV_EXT_long_vector so we can't do local testing via SPIR-V either.

I do have the ability to remote into a windows machine, which I suppose is a fine way forward.

bogner: FWIW while I do think having some test coverage of the vector rawBufferVectorLoad/Store is important and that using `ByteAddressBuffer` simplifies the tests, I'd be okay with most of these doing the more awkward code if it makes your life better.

That said, the metal shaderconverter definitely doesn't support all of long vector yet - I'm fairly surprised that any of the operations that actually lower to a `dx.op.*` with long vectors work there at all.

test/Feature/HLSLLib/long-vector/abs.long-vector.32.test
```diff
@@ -0,0 +1,65 @@
...
+        Binding: 1
+...
+#--- end
+
+# Unimplemented https://github.com/llvm/wg-hlsl/issues/467
+# XFAIL: Vulkan && DXC
+
+# Bug https://github.com/llvm/offload-test-suite/issues/1506
+# Note We fail for DXC because today Clang scalairzed and does
+# Not hit  IRErrorCodeUnsupportedInstruction on vector @dx.op.*
```
bogner, line 59: I find this sentence a bit confusing (in addition to the typo and the odd whitespace). How about:

```suggestion
# Note that we only pass on clang because Clang scalarizes. We will hit
# IRErrorCodeUnsupportedInstruction once Clang implements SM 6.10.
```

test/Feature/HLSLLib/long-vector/asdouble.long-vector.64.test
```diff
@@ -0,0 +1,83 @@
+#--- source.hlsl
+
+ByteAddressBuffer In0 : register(t0);
+ByteAddressBuffer In1 : register(t1);
+RWByteAddressBuffer Out : register(u2);
+
+[numthreads(1,1,1)]
+void main() {
+  Out.Store<vector<double, 8> >(0, asdouble(In0.Load<vector<uint, 8> >(0),
+      In1.Load<vector<uint, 8> >(0)));
```
bogner, line 10: clang-format (once you convince it to use `Style: c++03`) says:

```suggestion
  Out.Store<vector<double, 8> >(0, asdouble(In0.Load<vector<uint, 8> >(0),
                                            In1.Load<vector<uint, 8> >(0)));
```

test/Feature/HLSLLib/long-vector/asint16.long-vector.16.test
```diff
@@ -0,0 +1,69 @@
...
+---
+Shaders:
+  - Stage: Compute
+    Entry: main
+Buffers:
+  - Name: In0
+    Format: Float16
+    Stride: 2
+    Data: [0x4000, 0x3800, 0x0000, 0x3C00, 0xBC00, 0x4000, 0x3800, 0x0000, 0x3C00, 0xBC00, 0x4000, 0x3800, 0x0000, 0x3C00, 0xBC00, 0x4000]
+    #     [     2,    0.5,      0,      1,     -1,      2,    0.5,      0,      1,     -1,      2,    0.5,      0,      1,     -1,      2]
```
bogner, line 22: aside (as in we don't need to change it here): it's really annoying that we use hex for float16 - it'd be nice to just use literals for exactly representable values like we do for float32

test/Feature/HLSLLib/long-vector/acos.long-vector.32.test
```diff
@@ -0,0 +1,66 @@
...
+    Stride: 4
+    FillSize: 20
+  - Name: ExpectedOut
+    Format: Float32
+    Stride: 4
+    Data: [3.14159265, 2.0943951, 1.57079633, 1.04719755, 0]
+Results:
+  - Result: Test
+    Rule: BufferFloatEpsilon
+    Epsilon: 0.008
```
bogner, line 33: Why choose epsilon over ULPs here (and elsewhere), and why 0.008? Is this how the precision of these ops is specified?

test/Feature/HLSLLib/long-vector/clamp.long-vector.32.test
```diff
@@ -0,0 +1,92 @@
...
+ByteAddressBuffer In0 : register(t0);
+ByteAddressBuffer In1 : register(t1);
+ByteAddressBuffer In2 : register(t2);
+RWByteAddressBuffer Out : register(u3);
+
+[numthreads(1,1,1)]
+void main() {
+  Out.Store<vector<float, 5> >(0, clamp(In0.Load<vector<float, 5> >(0),
+      In1.Load<vector<float, 5> >(0),
+      In2.Load<vector<float, 5> >(0)));
```
bogner, line 12:
```suggestion
  Out.Store<vector<float, 5> >(0, clamp(In0.Load<vector<float, 5> >(0),
                                        In1.Load<vector<float, 5> >(0),
                                        In2.Load<vector<float, 5> >(0)));
```

test/Feature/HLSLLib/long-vector/clip.long-vector.32.test
bogner, line 1: I'm not really following what's happening in this test. I guess if `clip(Input5)` discarded the pixel then `clip(Input8)` and `clip(Input16)` are redundant, but also every value in each input is identical. Can you add some comments about what we're doing here and/or simplify the test?

test/Feature/HLSLLib/long-vector/fma.long-vector.64.test
```diff
@@ -0,0 +1,94 @@
...
+ByteAddressBuffer In0 : register(t0);
+ByteAddressBuffer In1 : register(t1);
+ByteAddressBuffer In2 : register(t2);
+RWByteAddressBuffer Out : register(u3);
+
+[numthreads(1,1,1)]
+void main() {
+  Out.Store<vector<double, 8> >(0, fma(In0.Load<vector<double, 8> >(0),
+      In1.Load<vector<double, 8> >(0),
+      In2.Load<vector<double, 8> >(0)));
```
bogner, line 12: More formatting. I'm going to stop pointing these out now but there are quite a few.

bogner (commented):
Generally looks pretty good. Mostly have a bunch of nitpicks and a question or two.

test/Feature/HLSLLib/long-vector/abs.long-vector.32.test
farzonl: just deleted it. Comment isn't needed captured it instead in the issue.

test/Feature/HLSLLib/long-vector/asdouble.long-vector.64.test
farzonl: How did you do this when I run clang-format it changes up the yaml?

bogner: I have a `clang-format-region` keybinding set up in my editor, otherwise it's a bit of a pain. You could run clang-format on the split files in the build directory but you may run into issues with shaders that use semantics in the arguments to the entry point, as those aren't handled very well by clang-format yet.

farzonl: I don't have that tried doing it with regular clang format. then tried writing a script that ran split file did the format then merged back, eventually I just gave up and updated my generator to just align multi argument intrinsics. I would rather not keep getting caught up on formatting for tests, but you let me know how it looks.

bob80905 (approved):
LGTM

test/Feature/HLSLLib/long-vector/clip.long-vector.32.test
```diff
@@ -0,0 +1,77 @@
...
+//--- pixel.hlsl
+struct PSInput {
+  float4 position : SV_POSITION;
+};
+
+float4 main(PSInput input) : SV_TARGET {
+  vector<float, 5> Input5;
+  for (uint I = 0; I < 5; ++I)
+    Input5[I] = input.position.x - 2.0;
+  clip(Input5);
```
llvm-beanz: Because every vector component is assigned the same value here, the behavior is identical with a vector5 as it would be if the vector were truncated to a vector4 or a scalar.

It is probably worth filling the vector with different values.

llvm-beanz (commented):
I suspect that all of these tests should have:
```
REQUIRES: Clang || SM_6_9
```

Without that these will all fail if they are run on a driver that doesn't support SM 6.9.

Alternatively, it may make more sense to have a "LongVector" feature that is enabled on SM 6.9 or Vulkan based on the Vulkan extension.

farzonl:
> I suspect that all of these tests should have:
> 
> ```
> REQUIRES: Clang || SM_6_9
> ```
> 
> Without that these will all fail if they are run on a driver that doesn't support SM 6.9.
> 

I had `# REQUIRES: !DXC || SM_6_9` before this commit:

 https://github.com/llvm/offload-test-suite/pull/1463/changes/36b8f49d3aa43b86f083f81a7303b409297da734

However that was problematic because it was hiding the fact that the offload test suit wasn't reporting 6_9 as a supported Shader model and just not running these tests for DXC. So I removed this on purpose because it was hiding issues.

> Alternatively, it may make more sense to have a "LongVector" feature that is enabled on SM 6.9 or Vulkan based on the Vulkan extension.

So `SPV_EXT_long_vector` / `VK_EXT_shader_long_vector` is currently enabled by default in clang regardless of shader model not sure what you are suggesting here.

PR #1465, opened 2026-08-20
[Vulkan][DirectX] Refactor texture handling in anticipation of other texture types

Icohedron:
A new TextureDimension enum has been added. Places where a 2D texture-related were expected are now replaced with a helper function and switch over the enum to return the appropriate value.

Places that assumed the number of texture subresources was the number of mip levels are now replaced with a call to a helper to obtain the correct subresource count. The number of subresources is equal to the number of mip levels right now, but that will change with array textures which will add array slices.

Added backend-agnostic helper functions for texture subresource layout arithmetic, and symmetric functions for copying texture contents between tightly-packed host buffers and GPU-mapped memory regions.

Also fixes two latent bugs:
DX copyTextureToBuffer only copied subresource 0, never reading back mip beyond the most detailed one (mip 0).
VK copyTextureToBuffer passed regionCount = 0, making the copy a no-op.

Assisted by: Claude Opus 5

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/include/API/Enums.h
+++ b/include/API/Enums.h
@@ -26,11 +26,11 @@ enum class ResourceKind {
   AccelerationStructure,
 };
 
-enum class TextureShape {
-  Texture1D,
-  Texture2D,
-  Texture3D,
-  TextureCube,
+enum class ResourceDimension {
+  Dim1D,
+  Dim2D,
+  Dim3D,
+  Cube,
 };
 
 enum ShaderContainerType {
--- a/include/API/Enums.h
+++ b/include/API/Enums.h
@@ -26,6 +26,13 @@ enum class ResourceKind {
   AccelerationStructure,
 };
 
+enum class TextureDimension {
+  One,
+  Two,
+  Three,
+  Cube,
+};
+
 enum ShaderContainerType {
   DXIL,
   SPIRV,
--- a/include/API/Enums.h
+++ b/include/API/Enums.h
@@ -26,11 +26,11 @@ enum class ResourceKind {
   AccelerationStructure,
 };
 
-enum class TextureDimension {
-  One,
-  Two,
-  Three,
-  Cube,
+enum class TextureShape {
+  Texture1D,
+  Texture2D,
+  Texture3D,
+  TextureCube,
 };
 
 enum ShaderContainerType {
--- a/include/API/Texture.h
+++ b/include/API/Texture.h
@@ -13,6 +13,7 @@
 #define OFFLOADTEST_API_TEXTURE_H
 
 #include "API/API.h"
+#include "API/Enums.h"
 #include "API/Resources.h"
 
 #include "llvm/ADT/BitmaskEnum.h"
@@ -21,6 +22,7 @@
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/Error.h"
 
+#include <algorithm>
 #include <cstdint>
 #include <optional>
 #include <string>
@@ -73,9 +75,20 @@ struct TextureCreateDesc {
   MemoryBacking Backing = MemoryBacking::Automatic;
   TextureUsage Usage = {};
   Format Fmt = Format::RGBA32Float;
+  TextureDimension Dim = TextureDimension::Two;
   uint32_t Width = 1;
   uint32_t Height = 1;
   uint32_t MipLevels = 1;
+
+  // The total number of subresources: one per mip level today.
+  uint32_t getSubresourceCount() const { return MipLevels; }
+
+  uint32_t getMipWidth(uint32_t Mip) const {
+    return std::max(1u, Width >> Mip);
+  }
+  uint32_t getMipHeight(uint32_t Mip) const {
+    return std::max(1u, Height >> Mip);
+  }
   // Clear value for render target or depth/stencil textures.
   // How and when this is applied depends on the backend:
   // - DX uses it as an optimized clear hint at resource creation time
@@ -89,6 +102,11 @@ inline llvm::Error validateTextureCreateDesc(const TextureCreateDesc &Desc) {
         std::errc::invalid_argument,
         "Format '%s' is not compatible with texture creation.",
         getFormatName(Desc.Fmt).data());
+
+  // Only 2D textures are implemented for now.
+  if (Desc.Dim != TextureDimension::Two)
+    return llvm::createStringError(std::errc::not_supported,
+                                   "Only 2D textures are supported.");
 
   const bool IsDepth = isDepthFormat(Desc.Fmt);
   const bool IsRT = (Desc.Usage & TextureUsage::RenderTarget) != 0;
@@ -159,7 +177,26 @@ struct SubresourceFootprint {
 struct TextureUploadLayout {
   llvm::SmallVector<SubresourceFootprint> Subresources; // One entry per mip.
   uint64_t TotalSizeInBytes = 0;
+
+  // The size of the tightly-packed host-side representation of this texture:
+  // every subresource's rows back to back with no row or subresource padding.
+  uint64_t getPackedSizeInBytes() const {
+    uint64_t Size = 0;
+    for (const SubresourceFootprint &Sub : Subresources)
+      Size += uint64_t(Sub.RowSizeInBytes) * Sub.NumRows;
+    return Size;
+  }
 };
+
+// Copy a texture's contents between a tightly-packed host buffer and a
+// (possibly row/subresource padded) GPU-visible mapping described by `Layout`.
+// These are the single place that knows how to bridge the two layouts, so
+// backends with alignment requirements (e.g. D3D12's 256-byte
+// D3D12_TEXTURE_DATA_PITCH_ALIGNMENT) work for arbitrary widths.
+void copyPackedToTextureLayout(void *Dst, const void *PackedSrc,
+                               const TextureUploadLayout &Layout);
+void copyTextureLayoutToPacked(void *PackedDst, const void *Src,
+                               const TextureUploadLayout &Layout);
 
 // Compute a tightly-packed upload layout (no row or subresource padding) for
 // the given texture description. Suitable for backends whose buffer-to-texture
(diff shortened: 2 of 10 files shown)


include/API/Texture.h
```diff
@@ -73,9 +75,20 @@ struct TextureCreateDesc {
   MemoryBacking Backing = MemoryBacking::Automatic;
   TextureUsage Usage = {};
   Format Fmt = Format::RGBA32Float;
+  TextureDimension Dim = TextureDimension::Two;
```
bob80905: Is there a way we can leave this uninitialized and assign it later, or on construction, to allow for other dim assignments in the future?

include/API/Texture.h
```diff
@@ -159,8 +177,27 @@ struct SubresourceFootprint {
...
+    uint64_t Size = 0;
+    for (const SubresourceFootprint &Sub : Subresources)
+      Size += uint64_t(Sub.RowSizeInBytes) * Sub.NumRows;
+    return Size;
+  }
 };
 
+// Copy a texture's contents between a tightly-packed host buffer and a
+// (possibly row/subresource padded) GPU-visible mapping described by `Layout`.
+// These are the single place that knows how to bridge the two layouts, so
```
bob80905: nit: comment grammar

lib/API/DX/Device.cpp
```diff
@@ -104,6 +104,58 @@ static uint32_t getAlignedTexturePitch(uint32_t Width, uint32_t ElementSize) {
...
+}
+
+static D3D12_SRV_DIMENSION getDXSRVDimension(const TextureCreateDesc &Desc) {
+  switch (Desc.Dim) {
+  case TextureDimension::Two:
+    return D3D12_SRV_DIMENSION_TEXTURE2D;
+  case TextureDimension::One:
+  case TextureDimension::Three:
+  case TextureDimension::Cube:
+    llvm_unreachable("Texture dimension has no SRV mapping yet");
```
bob80905, line 142: [SUGGESTION] Why not complete this function and the below function right now like you did with `getDXResourceDimension`?
I understand that only 2D is supported right now, but aren't the enums defined (e.g., D3D12_SRV_DIMENSION_TEXTURE3D)? Can't we just complete the function now (and it won't be used since tests won't test against unimplemented texture dimensions), instead of leaving it partially complete and needing to add to this in future PRs that implement new dimensions?

include/API/Texture.h
Icohedron: There is existing code that uses `TextureCreateDesc` while leaving some fields at default values. 2D is the default dimension for the most common use-cases like render targets, depth, and stencil buffers.

lib/API/DX/Device.cpp
Icohedron: We generally don't want to add functionality without testing it immediately. Leaving it blank here at least has an expected fail if someone does try to use it before it gets extensively tested to ensure it works properly.

bob80905: I understand your perspective, though I still think I'd follow through / stay consistent with what you did with `getDXResourceDimension` (rather than writing an unreachable for the yet-to-be-implemented dimension types)
Not a blocking comment though.

bob80905 (approved):
only outstanding thing is grammar inside a comment, LGTM

lib/API/DX/Device.cpp
Icohedron: `getDXSRVDimension` also depends on extra variables that are not yet available. 
Subsequent PRs in this stack add the required fields and logic to determine the correct enum to return.

For example, the texture arrays PR modifies this 2D case to return the correct enum for arrays: https://github.com/llvm/offload-test-suite/pull/1468/changes#diff-70fee035163d029347502582085e489e1c1cc30c901b95dbfa7aaf346a0f76afR143-R144

lib/Support/OffloadMigration.cpp
kmpeng, line 293: From my understanding, `calculateLinearSizeInBytes` is sized to only cover mip 0, but the bug fixes you made to `copyTextureToBuffer` now write every subresource into this buffer. For a multi mip texture, the copy would write past the end of the buffer.

lib/API/VK/Device.cpp
```diff
@@ -1174,9 +1242,11 @@ class VKComputeEncoder : public offloadtest::ComputeEncoder {
     insertDebugSignpost(
         llvm::formatv("copyTextureToBuffer {0} -> {1}", VKSrc.Name, VKDst.Name)
             .str());
+    const llvm::SmallVector<VkBufferImageCopy> Regions =
+        getTextureCopyRegions(VKSrc.Desc);
     vkCmdCopyImageToBuffer(CB.CmdBuffer, VKSrc.Image,
                            VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL, VKDst.Buffer,
-                           0, nullptr);
+                           Regions.size(), Regions.data());
```
kmpeng, line 1249: Are there any tests exercising this (and the DX) fix? If not, should we add one with this PR?

Icohedron: There are no tests that exercise these.
The Vulkan code in particular is also currently unreachable, so it can't even be tested yet.
I can probably add a test for the DirectX bug fix.

lib/Support/OffloadMigration.cpp
Icohedron: Ah yea. Good catch. It's actually an artifact that appeared after splitting up the changes into multiple PRs.
The fix is in the next PR in this stack: https://github.com/llvm/offload-test-suite/pull/1468/changes#diff-54c1a908f2bfb7a8485e7bd4307448218e110b8c504003e3e2c6ec04cb6d2674L11-R13

but it should be moved into this PR instead.

include/API/Texture.h
```diff
@@ -90,6 +103,11 @@ inline llvm::Error validateTextureCreateDesc(const TextureCreateDesc &Desc) {
         "Format '%s' is not compatible with texture creation.",
         getFormatName(Desc.Fmt).data());
 
+  // Only 2D textures are implemented for now.
+  if (Desc.Dim != TextureDimension::Two)
+    return llvm::createStringError(std::errc::not_supported,
+                                   "Only 2D textures are supported.");
+
```
alsepkow, line 113: nit: We could add some additional checks here.

```suggestion
  if (Desc.Width == 0 || Desc.Height == 0)
    return llvm::createStringError(
        std::errc::invalid_argument,
        "Texture dimensions must be non-zero, got %ux%u.", Desc.Width,
        Desc.Height);

  if (Desc.MipLevels == 0)
    return llvm::createStringError(std::errc::invalid_argument,
                                   "Texture must have at least one mip level.");

  uint32_t MaxMipLevels = 0;
  for (uint32_t Dimension = std::max(Desc.Width, Desc.Height); Dimension != 0;
       Dimension >>= 1)
    ++MaxMipLevels;

  if (Desc.MipLevels > MaxMipLevels)
    return llvm::createStringError(
        std::errc::invalid_argument,
        "Texture dimensions %ux%u support at most %u mip levels, got %u.",
        Desc.Width, Desc.Height, MaxMipLevels, Desc.MipLevels);
```

lib/Support/OffloadMigration.cpp
alsepkow: Had a pending comment on the same thing

include/API/Enums.h
```diff
@@ -26,6 +26,13 @@ enum class ResourceKind {
   AccelerationStructure,
 };
 
+enum class TextureDimension {
+  One,
+  Two,
+  Three,
+  Cube,
+};
+
```
alsepkow: nit: 'One', 'Two', 'Three' is odd to me. This space has a precedent of '1D', '2D', '3D'
And I don't think 'Cube' is really a dimension?

```suggestion
enum class TextureShape {
  Texture1D,
  Texture2D,
  Texture3D,
  TextureCube,
};
```

include/API/Texture.h
```diff
@@ -73,9 +75,20 @@ struct TextureCreateDesc {
...
+  TextureDimension Dim = TextureDimension::Two;
   uint32_t Width = 1;
   uint32_t Height = 1;
   uint32_t MipLevels = 1;
+
+  // The total number of subresources: one per mip level today.
+  uint32_t getSubresourceCount() const { return MipLevels; }
+
+  uint32_t getMipWidth(uint32_t Mip) const {
+    return std::max(1u, Width >> Mip);
```
alsepkow, line 89:
```suggestion
    assert(Mip < MipLevels && "Mip level index out of bounds.");
    return std::max(1u, Width >> Mip);
```

include/API/Texture.h
```diff
@@ -73,9 +75,20 @@ struct TextureCreateDesc {
...
+  }
+  uint32_t getMipHeight(uint32_t Mip) const {
+    return std::max(1u, Height >> Mip);
```
alsepkow, line 93:
```suggestion
    assert(Mip < MipLevels && "Mip level index out of bounds.");
    return std::max(1u, Height >> Mip);
```

include/API/Texture.h
```diff
@@ -21,6 +22,7 @@
 #include "llvm/Support/Casting.h"
 #include "llvm/Support/Error.h"
 
+#include <algorithm>
 #include <cstdint>
```
alsepkow, line 27:
```suggestion
#include <cassert>
#include <cstdint>
```

Needed for my other assert suggestions, I think.

alsepkow (approved):
Added a few comments. LGTM with those addressed.

kmpeng (approved):

include/API/Enums.h
Icohedron: Actually, we do call these dimensions in the DXIL ABI: https://github.com/llvm/llvm-project/blob/3e8996717946a90a5373c6af465143ccaa0b842b/llvm/include/llvm/Support/DXILABI.h#L34-L40
```c++
enum class ResourceDimension : uint8_t {
  Unknown = 0,
  Dim1D,
  Dim2D,
  Dim3D,
  Cube,
};
```

include/API/Texture.h
Icohedron: I added this extra validation and some tests to exercise it.

PR #1472, opened 2026-08-21
Add `select` matrix tests

### LINKED ISSUE #792: [Matrix] Add matrix test for `select`
> Note: This issue tracks matrix testing completion for `select`. The original test issues are [here](https://github.com/llvm/offload-test-suite/issues/132) (basic types) and [here](https://github.com/llvm/offload-test-suite/issues/213) (struct).  

Matrix requirements:

kmpeng:
Closes #792.

Adds matrix tests for `select` testing `float`. Will remove the XFAIL once the Clang implementation is in.
Only testing `float` because `select` behaves the same at every element type and bit size, and we already did extensive testing on those in the scalar/vector tests.

Assisted-by: Claude Opus 4.8

### DIFF
--- a/test/Feature/HLSLLib/select_mat.test
+++ b/test/Feature/HLSLLib/select_mat.test
@@ -1,0 +1,172 @@
+#--- source.hlsl
+
+// This test tests all the following scenarios for select on matrices:
+//   - Boolean matrix condition, matrix true/false values (3x2, 2x4, 2x2)
+//   - Boolean matrix condition, scalar true value, matrix false value (3x1)
+//   - Boolean matrix condition, matrix true value, scalar false value (4x4)
+//   - Boolean matrix condition, scalar true/false values (2x2)
+//   - Non-boolean matrix condition, matrix true/false values (2x2)
+
+StructuredBuffer<bool4> Cond : register(t0);
+StructuredBuffer<float4> FloatCond : register(t1);
+StructuredBuffer<float4> TrueVal : register(t2);
+StructuredBuffer<float4> FalseVal : register(t3);
+
+RWStructuredBuffer<float4> Out : register(u4);
+
+[numthreads(1,1,1)]
+void main() {
+  float3x2 F32 = select(bool3x2(Cond[0].xyz, Cond[1].xyz),
+                        float3x2(TrueVal[0].xyz, TrueVal[1].xyz),
+                        float3x2(FalseVal[0].xyz, FalseVal[1].xyz));
+  float2x4 F24 = select(bool2x4(Cond[2], Cond[3]),
+                        float2x4(TrueVal[2], TrueVal[3]),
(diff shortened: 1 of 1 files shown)


test/Feature/HLSLLib/select_mat.test
```diff
@@ -0,0 +1,172 @@
...
+//   - Boolean matrix condition, matrix true/false values (3x2, 2x4, 2x2)
+//   - Boolean matrix condition, scalar true value, matrix false value (3x1)
+//   - Boolean matrix condition, matrix true value, scalar false value (4x4)
+//   - Boolean matrix condition, scalar true/false values (2x2)
+//   - Non-boolean matrix condition, matrix true/false values (2x2)
+
+StructuredBuffer<bool4> Cond : register(t0);
+StructuredBuffer<float4> FloatCond : register(t1);
+StructuredBuffer<float4> TrueVal : register(t2);
+StructuredBuffer<float4> FalseVal : register(t3);
```
inbelic, line 13: nit:

```suggestion
StructuredBuffer<uint4> FloatCond : register(t1);
StructuredBuffer<uint4> TrueVal : register(t2);
StructuredBuffer<uint4> FalseVal : register(t3);
```

It might be more clear if a test fails when it is an integer as the dump of the hex representation would be clear. This is more of a thought aloud then an actual request for change

inbelic (approved):

test/Feature/HLSLLib/select_mat.test
kmpeng: Hmm that's a good point, though I would have to use `int4` because I set `FalseVal = -TrueVal`, and about half the output values are negative which are just as unreadable in hex (e.g. `-35` -> `0xffffffdd`). Float32s also still get a readable decimal dump, so I think I'll keep this test as float. But will definitely keep this in mind for future tests

test/Feature/HLSLLib/select_mat.test
```diff
@@ -0,0 +1,172 @@
...
+        Space: 0
+      VulkanBinding:
+        Binding: 4
+#--- end
+
+# Unimplemented https://github.com/llvm/llvm-project/issues/184487
+# XFAIL: Clang
+
+# RUN: split-file %s %t
+# RUN: %dxc_target -HV 202x -T cs_6_5 -Fo %t.o %t/source.hlsl
```
farzonl: why is this SM 6.5?

farzonl (approved):

test/Feature/HLSLLib/select_mat.test
kmpeng: Oh that was just copy and pasted from other tests. I think almost all other tests (matrix or not) use 6.5 - should we normally be using 6.0 instead? But yes I'll change it.

PR #1479, opened 2026-08-25
Add tests for tan()

### LINKED ISSUE #107: Add test for tan
> Note: This test is parented to the matrix implementation issue because there is no original implementation issue.

Matrix tests are also needed:
<br>

A .test file should contain 
```
@@ -0,0 +1,54 @@
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# UNSUPPORTED: Clang-Vulkan
# RUN: split-file %s %t
```

danbrown-amd:
Resolves #107

Assisted-by: Claude Opus 4


test/Feature/HLSLLib/tan.32.test
```diff
@@ -0,0 +1,72 @@
...
+
+StructuredBuffer<float4> In : register(t0);
+
+RWStructuredBuffer<float4> Out : register(u1);
+
+[numthreads(1,1,1)]
+void main() {
+  Out[0] = tan(In[0]);
+  float4 Tmp = {tan(In[1].xyz), tan(In[1].w)};
+  Out[1] = Tmp;
```
kmpeng: (Same for `tan.16.test`)
nit: we shouldn't need `Tmp` variables

```suggestion
  Out[1] = float4(tan(In[1].xyz), tan(In[1].w));
```

test/Feature/HLSLLib/tan.32.test
```diff
@@ -0,0 +1,72 @@
...
+    FillSize: 48
+  - Name: ExpectedOut
+    Format: Float32
+    Stride: 16
+    Data: [ -1.73205, -1.0, -0.57735, -0.26795,
+             0.0, 0.26795, 0.57735, 1.0,
+             1.73205, -1.73205, 0.26795, -0.26795 ]
+    # -sqrt3, -1, -sqrt3/3, sqrt3-2
+    # 0, 2-sqrt3, sqrt3/3, 1
+    # sqrt3, -sqrt3, sqrt3-2, 2-sqrt3
```
kmpeng: (For all files)
Could you double check that all your comments correspond to the right `Data` values? I know this one at least is wrong

```suggestion
    # sqrt3, -sqrt3, 2-sqrt3, sqrt3-2
```

test/Feature/HLSLLib/tan.32.test
```diff
@@ -0,0 +1,72 @@
...
+RWStructuredBuffer<float4> Out : register(u1);
+
+[numthreads(1,1,1)]
+void main() {
+  Out[0] = tan(In[0]);
+  float4 Tmp = {tan(In[1].xyz), tan(In[1].w)};
+  Out[1] = Tmp;
+  float4 Tmp2 = {tan(In[2].xy), tan(In[2].zw)};
+  Out[2] = Tmp2;
+}
```
kmpeng, line 13: (Same for `tan.16.test`)
We should have a constant folding test

test/Feature/HLSLLib/tan.32.test
```diff
@@ -0,0 +1,72 @@
...
+Buffers:
+  - Name: In
+    Format: Float32
+    Stride: 16
+    Data: [ -1.04719755, -0.78539816, -0.52359878, -0.26179939,
+             0.0, 0.26179939, 0.52359878, 0.78539816,
+             1.04719755, -1.04719755, 0.26179939, -0.26179939 ]
+    # -pi/3, -pi/4, -pi/6, -pi/12
+    # 0, pi/12, pi/6, pi/4
+    # pi/3, -pi/3, pi/12, -pi/12
```
kmpeng: (Same for `tan.16.test`)
All the inputs here are small - could we add some larger ones so we also test how `tan` handles big angles? `sin`/`cos` do this with values like +-314.16 and 100. 

Unlike the matrix test, these tests do need to be using a more robust range of values.

test/Feature/HLSLLib/tan.nan-inf-denorm.32.test
```diff
@@ -0,0 +1,64 @@
+#--- source.hlsl
```
kmpeng: Hmmm I'm not sure we need this test? I don't really see it exercising anything important. I think the `sin`/`cos` nan/inf/denorm tests went in before we decided not to test nan/inf/denorm values in the offload-test-suite. 

Anyone else have more insight into why equivalent tests exist for `sin`/`cos` and/or if this is necessary for `tan`? Imo it's not necessary and I would lean towards removing it.

test/Feature/HLSLLib/tan.nan-inf-denorm.32.test
```diff
@@ -0,0 +1,64 @@
...
+      DirectXBinding:
+        Register: 1
+        Space: 0
+      VulkanBinding:
+        Binding: 1
+...
+#--- end
+
+# Bug https://github.com/KhronosGroup/SPIRV-Cross/issues/2525
+# XFAIL: Vulkan && MoltenVK
```
kmpeng: If we do keep this test, the linked bug is closed. Is this actually still failing?

test/Feature/HLSLLib/tan.32.test
danbrown-amd: Good catch.  Same line was wrong for 16 bits.

test/Feature/HLSLLib/tan.nan-inf-denorm.32.test
danbrown-amd: Removed XFAIL for now.  The remaining failure (Windows / Intel / dxc / 16-bit) is an output sign flip for +/- 314.16.  XFAIL or change to a non-triggering large value?

test/Feature/HLSLLib/tan.32.test
```diff
@@ -0,0 +1,73 @@
...
+StructuredBuffer<float4> In : register(t0);
+
+RWStructuredBuffer<float4> Out : register(u1);
+
+[numthreads(1,1,1)]
+void main() {
+  Out[0] = tan(In[0]);
+  Out[1] = float4(tan(In[1].xyz), tan(In[1].w));
+  Out[2] = float4(tan(In[2].xy), tan(In[2].zw));
+  Out[3] = float4(tan(-314.16), tan(-100), tan(100), tan(314.16));
```
kmpeng: (Same for `tan.16.test`)
Large values should also (probably primarily) be tested through the buffer inputs. We should also have a mix of values in the constant folding cases rather than only using large ones. 
We usually try to reuse input/output values from the buffer tests so we can verify the constant folded outputs match.

test/Feature/HLSLLib/tan_mat.test
```diff
@@ -0,0 +1,101 @@
...
+RWStructuredBuffer<float4> Out : register(u1);
+
+[numthreads(1,1,1)]
+void main() {
+  float3x2 R32 = tan(float3x2(In[0].xyz, In[1].xyz));
+  float2x4 R24 = tan(float2x4(In[2], In[3]));
+  float3x1 R31 = tan(float3x1(In[4].xyz));
+  float4x4 R44 = tan(float4x4(In[5], In[6], In[7], In[8]));
+  float2x2 R22 = tan(float2x2(0.0, 0.52359878, 0.78539816, 1.04719755));
+  // 0, pi/6, pi/4, pi/3
```
kmpeng: nit: put this comment on the same line?

```suggestion
  float2x2 R22 = tan(float2x2(0.0, 0.52359878, 0.78539816, 1.04719755)); // 0, pi/6, pi/4, pi/3
```

test/Feature/HLSLLib/tan.16.test
```diff
@@ -0,0 +1,74 @@
...
+StructuredBuffer<half4> In : register(t0);
+
+RWStructuredBuffer<half4> Out : register(u1);
+
+[numthreads(1,1,1)]
+void main() {
+  Out[0] = tan(In[0]);
+  Out[1] = half4(tan(In[1].xyz), tan(In[1].w));
+  Out[2] = half4(tan(In[2].xy), tan(In[2].zw));
+  Out[3] = half4(tan(-314.16), tan(-100), tan(100), tan(314.16));
```
kmpeng: (Same for `tan.32.test`)
This should be
```
  Out[3] = tan(half4(...));
```
Currently `tan` runs before the inputs are converted to `half`, so it's not actually testing fp16 constant folding.

test/Feature/HLSLLib/tan.nan-inf-denorm.32.test
kmpeng: I suspect this might be failing because you're doing `half4(tan...))` instead of `tan(half4(...))` for the constant folding test. See my other comment about this.

test/Feature/HLSLLib/tan.16.test
kmpeng: This 16-bit test still doesn't test large values through the input buffer. The inputs/outputs between the 16-bit and 32-bit tests also diverge now, which is fine but it is usually easier to verify correctness when they match.

test/Feature/HLSLLib/tan.16.test
```diff
@@ -38,11 +38,11 @@ Buffers:
     Data: [ 0xBEEE, 0xBC00, 0xB89E, 0xB44A,
             0x0000, 0x344A, 0x389E, 0x3C00,
             0x3EEE, 0xBEEE, 0x344A, 0xB44A,
-            0x9205, 0x38B3, 0xB8B3, 0x1205 ]
+            0x0000, 0xBEEE, 0x3EEE, 0xB8B3 ]
     # -sqrt3, -1, -1/sqrt3, sqrt3-2
     # 0, 2-sqrt3, sqrt3/3, 1
     # sqrt3, -sqrt3, 2-sqrt3, sqrt3-2
```
kmpeng: (For all files)
nit: maybe we should write `sqrt3` as `sqrt(3)` because without parentheses, some of these values are a little ambiguous (e.g. `sqrt3-2` can mean `sqrt(3)-2` or `sqrt(3-2)`)

test/Feature/HLSLLib/tan.16.test
```diff
@@ -38,11 +38,11 @@ Buffers:
     Data: [ 0xBEEE, 0xBC00, 0xB89E, 0xB44A,
             0x0000, 0x344A, 0x389E, 0x3C00,
             0x3EEE, 0xBEEE, 0x344A, 0xB44A,
-            0x9205, 0x38B3, 0xB8B3, 0x1205 ]
+            0x0000, 0xBEEE, 0x3EEE, 0xB8B3 ]
     # -sqrt3, -1, -1/sqrt3, sqrt3-2
     # 0, 2-sqrt3, sqrt3/3, 1
```
kmpeng: (For all files)
nit: 1/sqrt3 and sqrt3/3 are the same value - probably worth picking one to use throughout all the files for consistency

test/Feature/HLSLLib/tan_mat.test
```diff
@@ -0,0 +1,100 @@
...
+            -1.73205, -1.0, -0.57735, -0.26795,
+            0.0, 0.26795, 0.57735, 1.0,
+            1.73205, -1.73205, 0.26795, -0.26795,
+            -1.0, 1.0, -1.73205, -0.57735,
+            0.0, 0.57735, 1.0, 1.73205 ]
+    # 3x2: -sqrt3, -1, -sqrt3/3, sqrt3-2, 0, 2-sqrt3, _, _
+    # 2x4: 0, 2-sqrt3, sqrt3/3, 1, sqrt3, -sqrt3, 2-sqrt3, sqrt3-2
+    # 3x1: -sqrt3, 0, sqrt3, _
+    # 4x4: -sqrt3, -1, -sqrt3/3, sqrt3-2, 0, 2-sqrt3, sqrt3/3, 1,
+    #      sqrt3, -sqrt3, 2-sqrt3, sqrt3-2, -1, 1, -sqrt3, -sqrt3/3
```
kmpeng: nit: these comments still use `sqrt3` and `sqrt3/3`. Can we update them to `sqrt(3)` and `1/sqrt(3)`?

kmpeng (approved):
LGTM other than 1 nit and deciding whether to delete the nan/inf/denorm test. I'll see if I can get an answer for you.

Also ideally let's get the matrix implementation of `tan` in, remove the XFAIL here, then merge this.

test/Feature/HLSLLib/tan.nan-inf-denorm.32.test
bob80905: Yes, IIRC, theres some UB with special values that we want to avoid testing, since drivers may implement behavior differently.
I agree that this test should be removed along with the sin / cos examples.

bob80905 (approved):
Approving because I have no preference between merging this in and then a follow-up PR removing the nan / inf / denorm set of tests for the trig functions, or removing it here right now.

test/Feature/HLSLLib/tan.nan-inf-denorm.32.test
kmpeng: Confirmed with the person who wrote the `sin`/`cos` tests that the nan/inf/denorm tests were added before the decision to avoid these values. I think we can safely remove this test.

danbrown-amd: Done.

test/Feature/HLSLLib/tan_mat.test
```diff
@@ -0,0 +1,100 @@
...
+      Kind: RWStructuredBuffer
+      DirectXBinding:
+        Register: 1
+        Space: 0
+      VulkanBinding:
+        Binding: 1
+#--- end
+
+# Unimplemented https://github.com/llvm/llvm-project/issues/184520
+# XFAIL: Clang
```
kmpeng:
```suggestion
# Bug https://github.com/llvm/llvm-project/issues/220723
# XFAIL: Clang && Vulkan
```

kmpeng (commented):
Will merge this in for you after this change is made, assuming all tests pass.

PR #1480, opened 2026-08-26
[SPIRV] Remove stale matrix swizzle XFAILs

alsepkow:
llvm/llvm-project#191909 exposed llvm/llvm-project#180258 and caused these tests to fail. That LLVM change was reverted by llvm/llvm-project#218804, and the scheduled Intel Clang Vulkan runs now report both tests as XPASS.

### DIFF
--- a/test/Basic/Matrix/matrix_swizzle_one_based.test
+++ b/test/Basic/Matrix/matrix_swizzle_one_based.test
@@ -112,9 +112,6 @@ DescriptorSets:
 ...
 #--- end
 
-# BUG https://github.com/llvm/llvm-project/issues/180258
-# XFAIL: Vulkan && Clang
-
 # RUN: split-file %s %t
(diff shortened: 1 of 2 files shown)


bob80905 (approved):

bogner (approved):
I worry we're kind of playing whack-a-mole with these tests, as they seem to keep hitting slight variants of these issues here. That said, it's fine to update them, so this LGTM.

A couple of minor comments on the commit message:

> Remove stale Clang Vulkan XFAILs from the zero- and one-based matrix
> swizzle tests

This is essentially just repeating the title - probably don't need it

> ## Rationale

It is very rarely worthwhile to use headings in a commit message - you can just drop this. It's clear that the explanation is rationale without labeling it as such.

> llvm/llvm-project#191909 exposed llvm/llvm-project#180258 and caused
> these tests to fail. That LLVM change was reverted by
> llvm/llvm-project#218804, and the scheduled Intel Clang Vulkan runs now
> report both tests as XPASS.

PR #1483, opened 2026-08-27
[SPIR-V] Add support for dynamic resources to the Vulkan backend

### LINKED ISSUE #1413: [SPIR-V] Add support for dynamic (directly indexed) resources to offload test suite
In https://github.com/llvm/offload-test-suite/issues/1360 a new `HeapIndex` parameter is going to be added to the `Resource` yaml element allow specifying where a resource can be found on the CVB/SRV/UAV or Sampler heap, for the purpose of direct resource indexing. We need to add support for this to the Vulkan's `Device.cpp` implementation.

hekota:
Implements the `HeapIndex` resource property (added for DirectX in #1415) in the Vulkan backend, so `ResourceDescriptorHeap[i]` / `SamplerDescriptorHeap[i]` tests run on SPIR-V.

DXC's SPIR-V backend lowers heap accesses into unbounded runtime arrays in descriptor set 0, so the runtime has to reconstruct the same layout: Each heap takes a single binding, picked from the bindings left unused by the explicitly
bound resources. If more than one resource type is read from the resource heap, DXC emits one runtime array per type, all aliasing that same binding. See the "ResourceDescriptorHeaps & SamplerDescriptorHeaps" section of DXC's
[docs/SPIR-V.rst](https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst#resourcedescriptorheaps-samplerdescriptorheaps).

Tests:
- removed `XFAIL: Vulkan` from two tests that are now passing
- added a test covering heap-indexed resource with counter

Fixes #1413

Assisted by Claude Opus 5.


lib/API/VK/Device.cpp
```diff
@@ -1687,6 +1797,17 @@ class VulkanDevice : public offloadtest::Device {
       EnabledDeviceExtensions.push_back(VK_EXT_MESH_SHADER_EXTENSION_NAME);
     }
 
+    // ????
```
bob80905: An odd comment, is this intentional?

test/Feature/DynamicResources/dyn-res-uav-counter.test
```diff
@@ -0,0 +1,98 @@
...
+
+Buffers:
+  - Name: DynResIndices
+    Format: Hex32
+    Data:
+      [ 0x4 ]
+
+  - Name: UavCounterBuf
+    Format: Float32
+    Stride: 16
```
alsepkow:
```suggestion
    Stride: 4
```

alsepkow: Wondering why this didn't fail?

test/Feature/DynamicResources/dyn-res-cvb-srv-uav.test
```diff
@@ -116,9 +116,6 @@ DescriptorSets:
 # Unimplemented: https://github.com/llvm/llvm-project/issues/133831
 # XFAIL: Clang
 
-# Unimplemented: https://github.com/llvm/offload-test-suite/issues/1413
-# XFAIL: Vulkan
```
alsepkow, line 120: Copilot pointed out that all three tests in this PR depend on the optional features: 
runtimeDescriptorArray and descriptorBindingPartiallyBound

Do we want to add a capability check for that?

test/Feature/DynamicResources/dyn-res-uav-counter.test
```diff
@@ -0,0 +1,98 @@
...
+        Binding: 1
+
+    - Name: UavCounterBuf
+      Kind: RWStructuredBuffer
+      HasCounter: true
+      HeapIndex: 4
+
+...
+#--- end
+
```
alsepkow, line 89: Does this test also require VK_EXT_mutable_descriptor_type support on vulkan?

test/Feature/DynamicResources/dyn-res-cvb-srv-uav.test
```diff
@@ -116,9 +116,6 @@ DescriptorSets:
...
-
```
alsepkow, line 121: Copilot is also saying DX should be gated on SM 6.6 and ResourceBindingTier == 3.

Do we care to check for that?

lib/API/VK/Device.cpp
```diff
@@ -1687,6 +1797,17 @@ class VulkanDevice : public offloadtest::Device {
...
+    if (HasMutableDescriptorTypeExt) {
+      if (!MutableDescriptorFeatures.mutableDescriptorType)
+        return llvm::createStringError(
+            std::errc::not_supported,
```
alsepkow, line 1920: Is this going to error out on general VK device creation when we don't have mutable descript types? But that sounds like an optional feature?

test/Feature/DynamicResources/dyn-res-uav-counter.test
```diff
@@ -0,0 +1,98 @@
+#--- source.hlsl
+
+// This test verifies handling of dynamic resource for
+// a constant buffer, SRV and UAV resources.
```
alsepkow:
```suggestion
// This test verifies a dynamically indexed RWStructuredBuffer
// with an associated UAV counter.
```

lib/API/VK/Device.cpp
```diff
@@ -1944,10 +2080,77 @@ class VulkanDevice : public offloadtest::Device {
           Binds.push_back(CB);
         }
       }
+
+      // The descriptor heaps always live in set 0. Their entries are sparse,
+      // so the unwritten slots are declared as partially bound.
+      const bool AddHeaps = (SetIdx == 0) && !HeapLayout.empty();
```
alsepkow, line 2224: nit: could factor the new block here and below into a helper.

```
llvm::Expected<VkDescriptorSetLayout>
createDescriptorSetLayout(const DescriptorSetLayoutDesc &SetDesc,
                          VkShaderStageFlags StageFlags,
                          const DescriptorHeapLayout *HeapLayout);

**Review by alsepkow — APPROVED** [review]
Adding some comments.

Someone with more experience in the area should still review.

File: lib/API/VK/Device.cpp (hunk 1 above)
Comment (hekota) [reply lib/API/VK/Device.cpp]: Oops :)

File: test/Feature/DynamicResources/dyn-res-uav-counter.test (hunk 4 above)
Comment (hekota) [reply test/Feature/DynamicResources/dyn-res-uav-counter.test:89]: I'll add it.

File: test/Feature/DynamicResources/dyn-res-cvb-srv-uav.test (hunk 3 above)
Comment (hekota) [reply test/Feature/DynamicResources/dyn-res-cvb-srv-uav.test:120]: I'll add it.

File: lib/API/VK/Device.cpp (hunk 6 above)
Comment (hekota) [reply lib/API/VK/Device.cpp:1920]: I've added a feature check for this to the test.

File: lib/API/VK/Device.cpp [hunk 9]
```diff
@@ -1914,40 +2041,117 @@ class VulkanDevice : public offloadtest::Device {
...
+              *VKBinding.CounterBinding, VK_DESCRIPTOR_TYPE_STORAGE_BUFFER,
+              RB.DescriptorCount, StageFlags);
           Binds.push_back(CB);
         }
       }
+
+      // The descriptor heaps always live in set 0. Their entries are sparse,
+      // so the unwritten slots are declared as partially bound.
+      const bool AddHeaps = (SetIdx == 0) && !HeapLayout.empty();
+      llvm::SmallVector<VkDescriptorBindingFlags, 8> BindingFlags;
```
bogner: Probably better to use the default size for these `SmallVector`s - it'll pick something reasonable for the stack based on the size of the elements.

```suggestion
      llvm::SmallVector<VkDescriptorBindingFlags> BindingFlags;
```

lib/API/VK/Device.cpp
```diff
@@ -3795,8 +4032,17 @@ class VulkanDevice : public offloadtest::Device {
 
         VkWriteDescriptorSet WDS = {};
         WDS.sType = VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET;
-        WDS.dstSet = IS.DescriptorSets[SetIdx];
-        WDS.dstBinding = R.VKBinding->Binding;
+        if (R.HeapIndex) {
+          // The heaps live in set 0 regardless of which set declared the
+          // resource; the heap index selects the slot within the binding.
+          WDS.dstSet = IS.DescriptorSets[0];
```
bogner, line 4188: Should we assert that `IS.DescriptorSets` isn't empty here, or is that accounted for?

bogner (approved):
LGTM

lib/API/VK/Device.cpp
```diff
@@ -106,6 +107,100 @@ static VkDescriptorType getDescriptorType(const ResourceKind RK) {
...
+        // A counter shares the heap index of the resource it belongs to.
+        if (RB.HasCounter) {
+          UsesCounterHeap = true;
+          Layout.CounterHeapSize = std::max(Layout.CounterHeapSize, End);
+        }
+      }
+    }
+  }
+
+  uint32_t NextBinding = 0;
```
bob80905, line 193: Question: I see this function called in 2 separate places.
Is it possible we want to declare this variable as static so its value persists between function calls?

bob80905 (approved):
Other than my question, LGTM

lib/API/VK/Device.cpp
bob80905: Talked with Helena offline, the two call sites to computeDescriptorHeapLayout should return identical layouts, and shouldn't use different binding indices, since the goal is recomputation on the 2nd call, not handling a different set of bindings.

PR #1486, opened 2026-08-27
Migrate Intel SKU to frequent build method

bob80905:
Like https://github.com/llvm/offload-test-suite/pull/1436, this PR migrates the intel runners to pull from the frequent builder's binaries.
#1436 built the offload test suite with MSVC, because the standalone configure in  test-callable.yaml  set no compiler. This PR makes that build use `clang-cl` from the downloaded distribution, matching the legacy in-tree build. This applies to the intel and the nvidia cells.
The intel and nvidia cells are now one `Exec-Tests-Windows` matrix job, which replaces the separate `Exec-Tests-Windows-NVIDIA` job. This renames the nvidia checks.

Assisted by: Github Copilot
Addresses: https://github.com/llvm/offload-test-suite/issues/1388

### DIFF
--- a/.github/workflows/test-callable.yaml
+++ b/.github/workflows/test-callable.yaml
@@ -148,8 +148,6 @@ jobs:
               echo "::error::clang-tidy missing from the downloaded distribution ($clang_tidy). The resolved frequent-build-llvm run predates clang-tidy being added to StandaloneDistribution.cmake; rebuild or resolve a newer run."
               exit 1
             fi
-            # The legacy build used clang-cl for these sources. Use clang-cl
-            # from the distribution to keep the same compiler.
             clang_cl="$GITHUB_WORKSPACE/install/bin/clang-cl"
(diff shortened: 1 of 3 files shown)


.github/workflows/test-callable.yaml
```diff
@@ -148,9 +148,19 @@ jobs:
               echo "::error::clang-tidy missing from the downloaded distribution ($clang_tidy). The resolved frequent-build-llvm run predates clang-tidy being added to StandaloneDistribution.cmake; rebuild or resolve a newer run."
               exit 1
             fi
+            # The legacy build used clang-cl for these sources. Use clang-cl
+            # from the distribution to keep the same compiler.
```
farzonl: this comment isn't going to make sense after this pr.  just delete it.

farzonl (approved):

bogner (approved):

PR #1490, opened 2026-08-31
Add `InterlockedCompareStoreFloatBitwise` tests

### LINKED ISSUE #853: Add test for `InterlockedCompareStoreFloatBitwise`
- [ ] Test for 16 bit int types (`InterlockedCompareStoreFloatBitwise.int16.test`)
- [ ] Test for Half type (`InterlockedCompareStoreFloatBitwise.fp16.test`)
- [ ] Test for 32 bit types (`InterlockedCompareStoreFloatBitwise.32.test`)
- [ ] Test for 64 bit int types (`InterlockedCompareStoreFloatBitwise.int64.test`)
- [ ] Test for Double type (`InterlockedCompareStoreFloatBitwise.fp64.test`)

A .test file should contain 
```
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# RUN: split-file %s %t
```

bob80905:
This PR adds tests for `InterlockedCompareStoreFloatBitwise`.
It mirrors the tests added for the exchange variant here: https://github.com/llvm/offload-test-suite/pull/1450

Fixes https://github.com/llvm/offload-test-suite/issues/853
Assisted by: Github Copilot

### DIFF
--- a/test/Feature/HLSLLib/InterlockedCompareStoreFloatBitwise.32.test
+++ b/test/Feature/HLSLLib/InterlockedCompareStoreFloatBitwise.32.test
@@ -1,10 +1,10 @@
 #--- source.hlsl
 
 // Tests InterlockedCompareStoreFloatBitwise on groupshared destinations.
-// 256 threads run at the same time. Each test value is exact in float, thus
-// the shader compares floats directly. The intrinsic does not report the
-// original value, thus each check reads the destination. There are three
-// checks:
+// 256 threads run at the same time. Each test value is exactly representable
+// as a 32-bit float, thus the shader compares floats directly. The intrinsic
+// does not report the original value, thus each check reads the destination.
+// There are three checks:
 //
 //   1. Private slots. A compare that does not match must not store. A later
 //      compare that matches must store.
@@ -22,8 +22,8 @@ RWStructuredBuffer<float> NanIn : register(u2);
 groupshared float SlotsMinusTwoPointFive[256]; // per-thread slots, each -2.5
 groupshared float MatchMinusFive;              // contended, compare matches
 groupshared float NoMatchSeventySeven;         // contended, never matches
-groupshared float ZeroPositive;                // +0.0, compared against -0.0
-groupshared float NanQuiet;                    // NaN, compared against NaN
+groupshared float PositiveZero;                // +0.0, compared against -0.0
+groupshared float QuietNaN;                    // NaN, compared against NaN
 
 [numthreads(256, 1, 1)]
 void main(uint3 GTID : SV_GroupThreadID) {
@@ -31,13 +31,15 @@ void main(uint3 GTID : SV_GroupThreadID) {
   if (GTID.x == 0) {
     MatchMinusFive = -5.0f;
     NoMatchSeventySeven = 77.0f;
-    ZeroPositive = 0.0f;
-    NanQuiet = NanIn[0];
+    PositiveZero = 0.0f;
+    QuietNaN = NanIn[0];
   }
   GroupMemoryBarrierWithGroupSync();
 
-  // Check 1. Only the second call stores. If the first call stored, the
-  // second compare cannot match and the slot keeps the wrong value.
+  // Check 1. The compare against 1.0f must not store, leaving the slot at
+  // -2.5f. The following compare against -2.5f must store the per-thread
+  // value. If the first call stored, the second compare cannot match and the
+  // slot keeps the wrong value.
   InterlockedCompareStoreFloatBitwise(SlotsMinusTwoPointFive[GTID.x], 1.0f,
                                       99.0f);
   InterlockedCompareStoreFloatBitwise(SlotsMinusTwoPointFive[GTID.x], -2.5f,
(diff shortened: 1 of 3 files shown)


test/Feature/HLSLLib/InterlockedCompareStoreFloatBitwise.32.test
```diff
@@ -0,0 +1,152 @@
+#--- source.hlsl
+
+// Tests InterlockedCompareStoreFloatBitwise on groupshared destinations.
+// 256 threads run at the same time. Each test value is exact in float, thus
```
alsepkow: nit:
Is it common to say 'exact in float' instead of something like 'exactly representable as a 32-bit float' or 'exactly representable as a float'?

test/Feature/HLSLLib/InterlockedCompareStoreFloatBitwise.32.test
```diff
@@ -0,0 +1,152 @@
...
+  SlotsMinusTwoPointFive[GTID.x] = -2.5f;
+  if (GTID.x == 0) {
+    MatchMinusFive = -5.0f;
+    NoMatchSeventySeven = 77.0f;
+    ZeroPositive = 0.0f;
+    NanQuiet = NanIn[0];
+  }
+  GroupMemoryBarrierWithGroupSync();
+
+  // Check 1. Only the second call stores. If the first call stored, the
```
alsepkow: nit: 'Only the second call stores' is ambiguous. The 'second call' could also refer the second call of this function on any thread?

```
// The compare against 1.0f must not store, leaving the slot at -2.5f.
// The following compare against -2.5f must store the per-thread value.

File: test/Feature/HLSLLib/InterlockedCompareStoreFloatBitwise.32.test [hunk 3]
```diff
@@ -0,0 +1,152 @@
...
+// checks:
+//
+//   1. Private slots. A compare that does not match must not store. A later
+//      compare that matches must store.
+//   2. Contended locations. All threads propose the same value, thus the
+//      final value is the same whichever thread stores it. A compare that
+//      never matches must not change the destination.
+//   3. Bit patterns. +0.0 against -0.0 must not store. A NaN against the
+//      same NaN bits must store. A buffer supplies the NaN, because HLSL
+//      has no NaN literal.
```
alsepkow, line 16: nit: Suggest moving the explanatory comments directly above the test case locations.

test/Feature/HLSLLib/InterlockedCompareStoreFloatBitwise.32.test
```diff
@@ -0,0 +1,152 @@
...
+
+RWStructuredBuffer<uint> OutSlots : register(u0);
+RWStructuredBuffer<float> OutFinal : register(u1);
+RWStructuredBuffer<float> NanIn : register(u2);
+
+groupshared float SlotsMinusTwoPointFive[256]; // per-thread slots, each -2.5
+groupshared float MatchMinusFive;              // contended, compare matches
+groupshared float NoMatchSeventySeven;         // contended, never matches
+groupshared float ZeroPositive;                // +0.0, compared against -0.0
+groupshared float NanQuiet;                    // NaN, compared against NaN
```
alsepkow: nit: Do we have a precedent for the flipped adjective-non order?
I only see this naming style in the other InterlockedCompareExchangeFloatBitwise.32.test you added.

'PositiveZero' and 'QuietNaN' read easier to me 😛

alsepkow (approved):
LGTM. Just some nits.

test/Feature/HLSLLib/InterlockedCompareStoreFloatBitwise.32.test
bob80905: I don't think so, your alternative phrasing sounds better to me.

test/Feature/HLSLLib/InterlockedCompareStoreFloatBitwise.32.test
bob80905: True, modified the language.

test/Feature/HLSLLib/InterlockedCompareStoreFloatBitwise.32.test
bob80905: For consistency with previous tests and to serve as a header comment / table of contents, I would prefer to keep this as is.

test/Feature/HLSLLib/InterlockedCompareStoreFloatBitwise.32.test
bob80905: Totally reads easier, yeah.
I've changed the var names.

farzonl (approved):

PR #1492, opened 2026-09-01
Add `abs` matrix tests

### LINKED ISSUE #796: [Matrix] Add matrix test for `abs`
> Note: This issue tracks matrix testing completion for `abs`. The original test issue is [here](https://github.com/llvm/offload-test-suite/issues/157).  

Matrix requirements:

kmpeng:
Closes #796.

Adds matrix tests for `abs` testing `float`, `int`, and `uint`. Will remove the XFAILs once the Clang implementation is in.
Only testing 32-bit types because `float`/`int`/`uint` exercise all of the different `abs` paths on the DirectX and SPIR-V backends, and those paths behave the same at every bit size.

Assisted-by: Claude Opus 4.8


llvm-beanz (approved):

test/Feature/HLSLLib/abs_mat.uint.test
llvm-beanz, line 1: One question here, do we generally need abs on uint? I assume these always get optimized away to nothing right?

kmpeng: Yeah, I was actually debating whether to add this test for a while. I ended up adding it because `uint` goes through its own path (I believe a separate lowering in DXC and a different overload in Clang), so I thought it might be worth testing that it really behaves as an identity across targets. But definitely open to removing it.

test/Feature/HLSLLib/abs_mat.fp.test
```diff
@@ -0,0 +1,90 @@
...
+            1024.25, 1.0, 42.75, 7.5,
+            60000.0, 21.001, 999.5, 0.0,
+            12.5, 6.0, 34.75, 4096.0,
+            11.5, 2.25, 3.125, 4.0625,
+            100.0, 200.5, 300.25, 400.75,
+            8192.0, 16384.0, 32768.0, 25.25,
+            12.5, 6.0, 34.75, 4096.0 ]
+Results:
+  - Result: Test
+    Rule: BufferFloatULP
```
bob80905: Since this op at most should just change the sign bit, and ULPT = 0, do we want to use buffer exact instead?

bob80905 (approved):

PR #1500, opened 2026-09-04
Update passing InterlockedXYZ tests

bogner:
These are generally passing given the following changes:
- PR llvm/llvm-project#219277, fixing llvm/llvm-project#217196
- PR llvm/llvm-project#219317, fixing llvm/llvm-project#99125

There are two new issues that are still causing some of these to fail:
- llvm/llvm-project#221366 - an issue with TypedBuffer
- llvm/llvm-project#221370 - an issue with ByteAddressBuffer

Finally, the `InterlockedExchange.resources.32` test looks like it was erroneously marked XFAIL on Metal - it has never failed.

### DIFF
--- a/test/Feature/HLSLLib/InterlockedAdd.resources.int64.test
+++ b/test/Feature/HLSLLib/InterlockedAdd.resources.int64.test
@@ -222,7 +222,7 @@ DescriptorSets:
 # Bug: https://github.com/llvm/offload-test-suite/issues/1418
 # UNSUPPORTED: WARP && DXC
 
-# Bug: https://github.com/llvm/offload-test-suite/issues/1435
+# TODO: Bug: Unsupported ptrcast user. Please fix.
 # XFAIL: Clang && Vulkan
 
 # RUN: split-file %s %t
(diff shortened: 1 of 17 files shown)


test/Feature/HLSLLib/InterlockedAdd.resources.int64.test
```diff
@@ -222,7 +222,7 @@ DescriptorSets:
 # Bug: https://github.com/llvm/offload-test-suite/issues/1418
 # UNSUPPORTED: WARP && DXC
 
-# Bug: https://github.com/llvm/offload-test-suite/issues/1435
+# TODO: Bug: Unsupported ptrcast user. Please fix.
```
Icohedron: No issue filed for this yet?

Icohedron (approved):
LGTM, aside from one XFAIL missing an issue link, but that can be fixed in a follow-up.

test/Feature/HLSLLib/InterlockedAdd.resources.int64.test
bogner: This is https://github.com/llvm/llvm-project/issues/221370 - guess I missed one when I went back over the tests to add the link.

alsepkow (approved):

damyanp (approved):

PR #1512, opened 2026-09-10
Use values with smaller error for tan.16

bogner:
A few of the values here were a bit too likely to differ across APIs and hardware.

- In practice I would expect most uses of tan (especially in half precision) to be on inputs between -2pi and 2pi, and values like 100.0h that are 10pi or more could easily hit a lot of intermediate rounding.
- The 100pi input here is pretty far off from being representable in half (0x5ce9 is 314.25), so we're just testing another arbitrary value here that is also susceptible to rounding errors. Using the well-known 355/113 approximation of pi we can test something that's more likely to be consistent.
- We didn't have a test for pi itself, which seems pretty useful.
- An arbitrary number in the likely range is also a good idea, so I added cases for 2.0h.

### DIFF
--- a/test/Feature/HLSLLib/tan.16.test
+++ b/test/Feature/HLSLLib/tan.16.test
@@ -42,7 +42,7 @@ Buffers:
             0x0000, 0xBEEC, 0x3EEC, 0xc05f ]
     # -sqrt(3), -1, -1/sqrt(3), sqrt(3)-2
     # 2-sqrt(3), 1/sqrt(3), 1, sqrt(3)
-    # 0, tan(pi) == 0, tan(355.0) == 0, tan(2.0) ~= -2.185
+    # 0, tan(pi) == 0, tan(355.0) ~= 0.000, tan(2.0) ~= -2.185
     # 0, -sqrt(3), sqrt(3), tan(2.0) ~= -2.185
 Results:
   - Result: Test1
--- a/test/Feature/HLSLLib/tan.16.test
+++ b/test/Feature/HLSLLib/tan.16.test
@@ -41,8 +41,8 @@ Buffers:
             0x0000, 0xBEEC, 0x3EEC, 0xc05f ]
     # -sqrt(3), -1, -1/sqrt(3), sqrt(3)-2
     # 2-sqrt(3), 1/sqrt(3), 1, sqrt(3)
-    # 0, tan(-100pi), tan(100), tan(100pi)
-    # 0, -sqrt(3), sqrt(3), tan(2.0)
+    # 0, tan(pi) == 0, tan(355.0) == 0, tan(2.0) == 0
+    # 0, -sqrt(3), sqrt(3), tan(2.0) == 0
 Results:
   - Result: Test1
     Rule: BufferFloatEpsilon
--- a/test/Feature/HLSLLib/tan.16.test
+++ b/test/Feature/HLSLLib/tan.16.test
@@ -9,7 +9,7 @@ void main() {
   Out[0] = tan(In[0]);
   Out[1] = half4(tan(In[1].xyz), tan(In[1].w));
   Out[2] = half4(tan(In[2].xy), tan(In[2].zw));
-  Out[3] = tan(half4(0, -1.04719755, 1.04719755, 100)); // 0, -pi/3, pi/3, 100
+  Out[3] = tan(half4(0., -1.04719755, 1.04719755, 2.)); // 0, -pi/3, pi/3, 2.0
 }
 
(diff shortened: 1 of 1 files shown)


test/Feature/HLSLLib/tan.16.test
kmpeng: Change these too?

kmpeng (approved):
LGTM

test/Feature/HLSLLib/tan.16.test
```diff
@@ -9,7 +9,7 @@ void main() {
   Out[0] = tan(In[0]);
   Out[1] = half4(tan(In[1].xyz), tan(In[1].w));
   Out[2] = half4(tan(In[2].xy), tan(In[2].zw));
```
bob80905, line 11: nit: 0.0

test/Feature/HLSLLib/tan.16.test
```diff
@@ -24,10 +24,10 @@ Buffers:
     Stride: 8
     Data: [ 0xBC30, 0xBA48, 0xB830, 0xB430,
             0x3430, 0x3830, 0x3A48, 0x3C30,
-            0x0000, 0xDCE9, 0x5640, 0x5CE9 ]
+            0x0000, 0x4248, 0x5D8C, 0x4000 ]
     # -pi/3, -pi/4, -pi/6, -pi/12
     # pi/12, pi/6, pi/4, pi/3
-    # 0, -100pi, 100, 100pi
+    # 0, pi (rounded 3.140625), 113pi (rounded 355.0), 2.0
```
bob80905, line 31: So we can use a value much larger than 2pi here because it is a common well-tested value and we don't expect intermediate rounding issues?

bob80905 (approved):
LGTM

test/Feature/HLSLLib/tan.16.test
bogner: I don't understand what this is referring to - did this comment attach to the right line?

bob80905: Sorry no, it was attached to the line above the one I wanted, line 12.

test/Feature/HLSLLib/tan.16.test
bogner: We can get away with this since 355/113 is a surprisingly good rational approximation of pi (good to 6 decimal places!) and 355.0 is exactly representable as a half. I'm not super familiar with the specifics of the algorithms we'd need to worry about intermediate rounding in here, but certainly across the GPUs we're testing on this seems to be reliable enough.

I'd be okay with dropping the large value test entirely as well if you think this isn't all that valuable.

bob80905: Sounds good to me, we can remove it if it causes problems later.

bob80905 (approved):

test/Feature/HLSLLib/tan.16.test
bogner: I've updated it to explicitly spell out `0.0h` instead of `0.0`, and updated the other literals along with it

test/Feature/HLSLLib/tan.16.test
```diff
@@ -37,12 +38,12 @@ Buffers:
...
-            0x0000, 0xADD3, 0xB8B3, 0x2DD3,
-            0x0000, 0xBEEC, 0x3EEC, 0xB8B3 ]
+            0x0000, 0x0000, 0x0000, 0xc05f,
+            0x0000, 0xBEEC, 0x3EEC, 0xc05f ]
     # -sqrt(3), -1, -1/sqrt(3), sqrt(3)-2
     # 2-sqrt(3), 1/sqrt(3), 1, sqrt(3)
-    # 0, tan(-100pi), tan(100), tan(100pi)
-    # 0, -sqrt(3), sqrt(3), tan(100)
+    # 0, tan(pi) == 0, tan(355.0) == 0, tan(2.0) == 0
+    # 0, -sqrt(3), sqrt(3), tan(2.0) == 0
```
kmpeng: I don't think tan(2.0) = 0

bogner: lol, oops. It most certainly does not, and 0xc05f isn't zero either. Updated to say approximately -2.185

bob80905: NVM I'm wrong

bogner: I updated the comment about `tan(355)` as well since it's only approximately 0 (albeit to five decimal places)

PR #1516, opened 2026-09-14
Over-allocate raw buffers if they won't be a multiple of 4 bytes

### LINKED ISSUE #1515: [DirectX] Truncated buffers for raw access of odd numbers of 16-bit types
When using (RW)ByteAddressBuffer with an odd number of 16 bit types, we're truncating the last element of the buffer. Consider:

```
#--- source.hlsl

ByteAddressBuffer In0 : register(t0);
RWByteAddressBuffer Out : register(u1);

[numthreads(1,1,1)]
void main() {
  Out.Store<uint16_t>(8, In0.Load<uint16_t>(8));
}

//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
Buffers:
  - Name: In0
    Format: UInt16
    Stride: 2
    Data: [0, 0, 0, 0, 0x1234]
  - Name: Out
    Format: UInt16
    Stride: 2
    FillSize: 10
  - Name: ExpectedOut
    Format: UInt16
    Stride: 2
    Data: [0, 0, 0, 0, 0x1234]
Results:
  - Result: Test
    Rule: BufferExact
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In0
      Kind: ByteAddressBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWByteAddressBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end

# REQUIRES: Int16

# RUN: split-file %s %t
# RUN: %dxc_target -enable-16bit-types -T cs_6_5 -Fo %t.o %t/source.hlsl
```

This test fails, as `Data` is all zeros - the odd-numbered element at offset 8 is never read or written.

More helpfully, if we use a 1-element buffer we get an error:
```
# | D3D: ID3D12Device::CreateShaderResourceView: The Dimensions of the View are invalid due to at least one of the following conditions. Assuming this Format (0x27, R32_TYPELESS), FirstElement (value = 0) must be between 0 and the maximum offset of the Buffer, 18446744073709551615, inclusively. With the current FirstElement, NumElements (value = 0) must be between 1 and 0, inclusively, in order that the View fit on the Buffer.
# | Exception Code: 0x0000087A
```

The issue is with how we set up `NumElements` for raw buffers around [DX/Device.cpp#L2025](https://github.com/llvm/offload-test-suite/blob/main/lib/API/DX/Device.cpp#L2025) and [L2059](https://github.com/llvm/offload-test-suite/blob/main/lib/API/DX/Device.cpp#L2059). Something like the following resolves the issue, but may not be sound:

```diff
--- a/lib/API/DX/Device.cpp
+++ b/lib/API/DX/Device.cpp
@@ -1974,6 +1974,9 @@ public:
       BufferSizeInBytes = CounterOffsetInBytes + sizeof(uint32_t);
     }

+    if (UINT64 Rem = BufferSizeInBytes % 4)
+      BufferSizeInBytes += Rem;
+
     const D3D12_HEAP_PROPERTIES HeapProps = CD3DX12_HEAP_PROPERTIES(HeapType);
     const D3D12_RESOURCE_DESC BufferDesc =
         CD3DX12_RESOURCE_DESC::Buffer(BufferSizeInBytes, Flags);
@@ -2022,7 +2025,7 @@ public:
       switch (Desc.AccessType) {
       case BufferShaderAccessType::Raw:
         SRVDesc.Format = DXGI_FORMAT_R32_TYPELESS;
-        SRVDesc.Buffer.NumElements = static_cast<uint32_t>(SizeInBytes / 4);
+        SRVDesc.Buffer.NumElements = static_cast<uint32_t>(BufferSizeInBytes / 4);
         SRVDesc.Buffer.Flags = D3D12_BUFFER_SRV_FLAG_RAW;
         break;
       case BufferShaderAccessType::Typed:
@@ -2056,7 +2059,7 @@ public:
       switch (Desc.AccessType) {
       case BufferShaderAccessType::Raw:
         UAVDesc.Format = DXGI_FORMAT_R32_TYPELESS;
-        UAVDesc.Buffer.NumElements = static_cast<uint32_t>(SizeInBytes / 4);
+        UAVDesc.Buffer.NumElements = static_cast<uint32_t>(BufferSizeInBytes / 4);
         UAVDesc.Buffer.Flags = D3D12_BUFFER_UAV_FLAG_RAW;
         break;
       case BufferShaderAccessType::Typed:
```

bogner:
we need to make sure that the underlying storage doesn't truncate the last value if we have an odd number of 16-bit values accessed through a ByteAddressBuffer,

Resolves #1515


lib/API/DX/Device.cpp
```diff
@@ -1974,6 +1974,13 @@ class DXDevice : public offloadtest::Device {
       BufferSizeInBytes = CounterOffsetInBytes + sizeof(uint32_t);
     }
 
+    if (Desc.AccessType == BufferShaderAccessType::Raw) {
+      // If we have a raw buffer we need the allocation to be a multiple of 4
+      // bytes. Overallocate if we need to.
+      if (UINT64 Rem = BufferSizeInBytes % 4)
+        BufferSizeInBytes += Rem;
```
farzonl, line 1981: I am seeing a similar failure of trailing 0s on a fma long vector test which requires doubles. would this change fix that bug as well? I don't see anything specific here to 16 bit in the offloader changes. The fma tests were passing when I was using StructuredBuffer so this isn't an issue with the intrinsic.

```
 Format:          Float64
# | Stride:          8
# | Data:            [ 0, 0.5, 0, -0.5, 0.5, 1.5, 2, 1.5 ]
# | ...
# | Got:
# | ---
# | Name:            Out
# | Format:          Float64
# | Stride:          8
# | Data:            [ 0, 0.5, 0, 0, 0.5, 1.5, 0, 0 ]
```

bogner: No, this looks different.
- It's impossible for a buffer of doubles to have a size that isn't a multiple of 4 bytes (it will always be a multiple of 8 bytes)
- This output looks like we're only writing ~every other value - the first and third look fine but the second and fourth are missing~ every other pair of values? This is different than truncating the last one.

bogner: Looks like a bug in clang: https://hlsl.godbolt.org/z/sP6qboW1j

The indices we store to look suspect:
```llvm
  call void @dx.op.rawBufferStore.i32(i32 140, %dx.types.Handle %4, i32 0, i32 undef, i32 %101, i32 %105, i32 %102, i32 %106, i8 15, i32 4), !dbg !250
  call void @dx.op.rawBufferStore.i32(i32 140, %dx.types.Handle %4, i32 0, i32 undef, i32 %103, i32 %107, i32 %104, i32 %108, i8 15, i32 4), !dbg !250
  ...
  call void @dx.op.rawBufferStore.i32(i32 140, %dx.types.Handle %4, i32 32, i32 undef, i32 %113, i32 %117, i32 %114, i32 %118, i8 15, i32 4), !dbg !250
  call void @dx.op.rawBufferStore.i32(i32 140, %dx.types.Handle %4, i32 32, i32 undef, i32 %115, i32 %119, i32 %116, i32 %120, i8 15, i32 4), !dbg !250
```

We're storing two sets of two doubles at offset 0, clobbering each other, and then two sets at offset 32, clobbering each other again. These offsets should be 0, 16, 32, and 48.

farzonl (approved):

joaosaffran (approved):

PR #1518, opened 2026-09-14
Add *MemoryBarrier tests

### LINKED ISSUE #829: Add test for `AllMemoryBarrier`
- [ ] Test for 16 bit int types (`AllMemoryBarrier.int16.test`)
- [ ] Test for Half type (`AllMemoryBarrier.fp16.test`)
- [ ] Test for 32 bit types (`AllMemoryBarrier.32.test`)
- [ ] Test for 64 bit int types (`AllMemoryBarrier.int64.test`)
- [ ] Test for Double type (`AllMemoryBarrier.fp64.test`)

A .test file should contain 
```
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# RUN: split-file %s %t
```

### LINKED ISSUE #833: Add test for `DeviceMemoryBarrier`
- [ ] Test for 16 bit int types (`DeviceMemoryBarrier.int16.test`)
- [ ] Test for Half type (`DeviceMemoryBarrier.fp16.test`)
- [ ] Test for 32 bit types (`DeviceMemoryBarrier.32.test`)
- [ ] Test for 64 bit int types (`DeviceMemoryBarrier.int64.test`)
- [ ] Test for Double type (`DeviceMemoryBarrier.fp64.test`)

A .test file should contain 
```
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# RUN: split-file %s %t
```

### LINKED ISSUE #844: Add test for `GroupMemoryBarrier`
- [ ] Test for 16 bit int types (`GroupMemoryBarrier.int16.test`)
- [ ] Test for Half type (`GroupMemoryBarrier.fp16.test`)
- [ ] Test for 32 bit types (`GroupMemoryBarrier.32.test`)
- [ ] Test for 64 bit int types (`GroupMemoryBarrier.int64.test`)
- [ ] Test for Double type (`GroupMemoryBarrier.fp64.test`)

A .test file should contain 
```
#--- source.hlsl

[numthreads(1,1,1)]
void main() {
}
```
```
//--- pipeline.yaml

---
Shaders:
  - Stage: Compute
    Entry: main
    DispatchSize: [1, 1, 1]
Buffers:
  - Name: In
    Format: Float32
    Stride: 16
  - Name: Out
    Format: Float32
    Stride: 4
    ZeroInitSize: 12
    Format: Float32
    Stride: 4
    Data: [ 8, 3.14159, 13, 5 ]
  - Result: Test1
    ULPT: 1
    Actual: Out
    Expected: ExpectedOut
DescriptorSets:
  - Resources:
    - Name: In
      Kind: StructuredBuffer
      DirectXBinding:
        Register: 0
        Space: 0
      VulkanBinding:
        Binding: 0
    - Name: Out
      Kind: RWStructuredBuffer
      DirectXBinding:
        Register: 1
        Space: 0
      VulkanBinding:
        Binding: 1
...
#--- end
```
```
# RUN: split-file %s %t
```

bob80905:
This PR adds offload tests for:
`AllMemoryBarrier`
`DeviceMemoryBarrier`
`GroupMemoryBarrier`
builtins.

Fixes https://github.com/llvm/offload-test-suite/issues/829
Fixes https://github.com/llvm/offload-test-suite/issues/833
Fixes https://github.com/llvm/offload-test-suite/issues/844

Assisted by: Github Copilot


test/Feature/HLSLLib/AllMemoryBarrier.test
```diff
@@ -0,0 +1,67 @@
...
+
+groupshared uint GroupData[4];
+
+[numthreads(4, 1, 1)]
+void main(uint3 GTID : SV_GroupThreadID) {
+  uint Index = GTID.x;
+  GroupData[Index] = Index + 1;
+  UavData[Index] = Index + 5;
+
+  AllMemoryBarrier();
```
kmpeng: Hmm I'm kind of confused what these tests are actually testing. Wouldn't all these tests pass without the `*MemoryBarrier()` calls? Seems like every thread is just reading/writing to its own index. And from my understanding barriers only have observable effects across threads.

bob80905: You're right that these tests may pass without the calls. I don't think it's possible to test these are correctly implemented without using the `*Sync()` builtins, and we want to avoid using other irrelevant builtins in these tests.
However, the tests still provide end-to-end coverage that the builtin compiles, passes shader validation, survives backend translation, creates a pipeline, and executes well on each supported API.
Though, the tests would not catch a compiler that silently removes the barrier or emits the wrong memory semantics.
I think given what we set out to achieve for offload tests, this is sufficient.

kmpeng: I see, I agree. 

Maybe we should change the comment at the top of the files to clarify exactly what we're testing?

farzonl: survives backend translation seems like a test for the lit tests in the DirectX backend I do think Kaitlin has a point here.
I think you want to do something like this:
```
if Thread 1:
    Write: X = 1
    AllMemoryBarrier();
    Read: r1 = Y

else Thread 2:
    Write: Y = 1
    AllMemoryBarrier();
    Read: r2 = X
```

farzonl: OK so this is what copilot came up with from my example. Maybe something like this but without GroupMemory mixed in
```hlsl
// StoreBufferTest.compute
// Dispatches: Dispatch(NUM_TESTS / 64, 1, 1)

#define THREADS_PER_GROUP 128
#define HALF_GROUP 64 // Ensures Thread A and Thread B are in different waves/warps

// UAVs for the test variables. 
// Note: We do not use 'globallycoherent' because we are testing visibility 
// within the same Thread Group (same SM/Compute Unit L1 Cache).
RWStructuredBuffer<uint> TestData_X : register(u0);
RWStructuredBuffer<uint> TestData_Y : register(u1);

// Stores the read results for validation
RWStructuredBuffer<uint2> Results : register(u2);

// Atomic counter to track how many times the barrier failed
RWStructuredBuffer<uint> ViolationCount : register(u3);

[numthreads(THREADS_PER_GROUP, 1, 1)]
void CSMain(
    uint3 DTid : SV_DispatchThreadID, 
    uint GI : SV_GroupIndex,
    uint3 Gid : SV_GroupID)
{
    // Determine which half of the thread group this thread belongs to
    bool isThreadA = GI < HALF_GROUP;
    
    // The shared memory index this pair will operate on
    // Thread 0 and Thread 64 will operate on the same index, 
    // Thread 1 and Thread 65 will operate on the same index, etc.
    uint pairIndex = Gid.x * HALF_GROUP + (GI % HALF_GROUP);

    uint r = 0;

    if (isThreadA)
    {
        // THREAD A
        TestData_X[pairIndex] = 1;
        
        // The barrier under test. 
        // This should force the write to X to be visible before the read of Y.
        AllMemoryBarrier(); 
        
        r = TestData_Y[pairIndex];
        Results[pairIndex].x = r; // Store r1
    }
    else
    {
        // THREAD B
        TestData_Y[pairIndex] = 1;
        
        // The barrier under test.
        // This should force the write to Y to be visible before the read of X.
        AllMemoryBarrier(); 
        
        r = TestData_X[pairIndex];
        Results[pairIndex].y = r; // Store r2
    }

    // Synchronize the group so we can safely evaluate the results in the same shader pass
    GroupMemoryBarrierWithGroupSync();

    // Have only one thread from the pair check the outcome to prevent double-counting
    if (isThreadA)
    {
        uint r1 = Results[pairIndex].x;
        uint r2 = Results[pairIndex].y;
        
        // The Litmus Check:
        // If both threads read 0, the CPU/GPU buffered the writes and reordered them 
        // after the reads. The memory barrier completely failed.
        if (r1 == 0 && r2 == 0)
        {
            InterlockedAdd(ViolationCount[0], 1);
        }
    }
}
```

farzonl: > You're right that these tests may pass without the calls. 

Then we aren't testing the feature.

> I don't think it's possible to test these are correctly implemented without using the `*Sync()` builtins, and we want to avoid using other irrelevant builtins in these tests. 

If a feature depends on another then we should use it. 


> However, the tests still provide end-to-end coverage that the builtin compiles, passes shader validation, survives backend translation.

Thats a test of the compiler not the execution behavior. We should already have tests for the compiler in llvm.

> Though, the tests would not catch a compiler that silently removes the barrier or emits the wrong memory semantics. 

That should never be the purpose of the offloader. if the compiler is doing something wrong then we should add tests to the DirectX and SPIR-V backends to catch that.

> I think given what we set out to achieve for offload tests, this is sufficient.

I don't agree

PR #1520, opened 2026-09-16
[Metal][Vulkan][DirectX] Add MSAA render target support

### LINKED ISSUE #1043: [SPIRV][DirectX] Add MSAA render target support
Required for testing `SV_SampleIndex`, `SV_Coverage`, and future `Texture2DMS` support.

Current status: DX and Vulkan hardcode raster sample counts to `1`. Render-target readback also uses a direct texture-to-buffer copy, which is invalid for multisampled textures.

_Note_: The scope is raster sample-count configuration, color resolve, and semantic/image validation on DX and Vulkan. Texture2DMS binding is deferred to https://github.com/llvm/offload-test-suite/issues/1428

AC:
- [ ] Allow YAML config to specify an optional valid sample count
- [ ] Add `SampleCount` to texture, render-pass, traditional raster PSO, and mesh raster PSO descriptors
- [ ] Thread the count through color target, default depth target, render pass, and PSO creation; reject attachment-count mismatches.
- [ ] DX: Query multisample quality/format support, configure PSO and texture sample descriptions, and use `TEXTURE2DMS` for multisampled SRVs.
- [ ] Vulkan: Query supported sample counts and format properties, then configure image samples, attachment samples, and `rasterizationSamples`.
- [ ] Add a `ComputeEncoder::resolveTexture` api as follows:
- [ ] DX: Implement resolve with `ResolveSubresource` and `RESOLVE_SOURCE`/`RESOLVE_DEST` transitions.
- [ ] Vulkan: Implement resolve with `vkCmdResolveImage` and the required transfer layouts/transitions.
- [ ] Create a single-sample resolve target for MSAA raster pipelines and read back from it.
- [ ] Add a MSAA golden-image test on DX/WARP and Vulkan.

Icohedron:
Fixes https://github.com/llvm/offload-test-suite/issues/1043

This PR:

- Adds an optional `Bindings.SampleCount` and propagates it through texture, render-pass, traditional raster, and mesh raster descriptors. Validate attachment compatibility and query device and format support on DirectX and Vulkan. Metal reports multisampling as unsupported.

- Adds `ComputeEncoder::resolveTexture`, implemented with `ResolveSubresource` on DirectX and `vkCmdResolveImage` on Vulkan. Multisampled raster pipelines now resolve into a single-sample target before readback.

- Adds traditional and mesh raster tests using a dedicated 4x MSAA golden image (https://github.com/llvm/offload-golden-images/pull/12), plus validation tests for unsupported sample-count configurations.

Assisted by: GPT 5.6 Sol


Icohedron:
Tests `test/Graphics/MSAA/MeshShaderTriangle.test` and `test/Graphics/MSAA/ShaderTriangle.test` will fail until https://github.com/llvm/offload-golden-images/pull/12 is merged

docs/MipMappedTextures.md
```diff
@@ -5,6 +5,9 @@ verifying that the compiler correctly generates code for
 `CalculateLevelOfDetail`, `SampleGrad` (explicit gradients), and direct access
 via `mips[level][coords]`.
 
+Multisampled render targets have one mip level; see
+[MSAA.md](MSAA.md).
```
inbelic, line 9:
```suggestion
Multisampled render targets have one mip level; see [MSAA.md](MSAA.md).
```
nit

test/Graphics/MSAA/SimpleTriangle.test
```diff
@@ -0,0 +1,77 @@
...
+      Channels: 4
+      Offset: 12
+      Name: COLOR
+  RenderTarget: Output
+  SampleCount: 4
+DescriptorSets: []
+...
+#--- end
+
+# XFAIL: Clang && !Vulkan
```
inbelic, line 70: Missing issue link

test/Tools/Offloader/invalid-sample-count.test
inbelic, line 1: I'm not sure we really have a precedent for adding error tests of offloader error messages, so not really sure these need to be added. Otherwise should we be adding them for all the other error messages? Are these special that warrant them?

lib/API/MTL/MTLDevice.cpp
```diff
@@ -55,6 +55,19 @@ static llvm::Error toError(NS::Error *Err) {
...
 
+static llvm::Error validateMetalSampleCount(MTL::Device *Device,
+                                            uint32_t SampleCount,
+                                            llvm::StringRef FieldName) {
+  if (auto Err = validateSampleCount(SampleCount, FieldName))
+    return Err;
+  if (!Device->supportsTextureSampleCount(SampleCount))
+    return llvm::createStringError(
+        std::errc::not_supported,
+        "%s %u is not supported on this Metal device.", FieldName.str().c_str(),
```
bob80905, line 66: nit: I think this diagnostic is missing something, maybe:
"%s with sample count %u is not supported on this Metal device" ?

lib/API/VK/Device.cpp
```diff
@@ -1387,6 +1387,61 @@ class VKComputeEncoder : public offloadtest::ComputeEncoder {
...
+      return llvm::createStringError(
+          std::errc::not_supported,
+          "vkCmdResolveImage only supports color images.");
+
+    auto &VKSrc = llvm::cast<VulkanTexture>(Src);
+    auto &VKDst = llvm::cast<VulkanTexture>(Dst);
+
+    CB.addImageTransition(CB.PendingSrcAccess,                /*SrcAccessMask*/
+                          VK_ACCESS_TRANSFER_READ_BIT,        /*DstAccessMask*/
+                          VKSrc.preferredLayoutOrUndefined(), /*OldLayout*/
```
bob80905, line 1404: nit: these comments would slightly clear up the DX implementation too, old+new layout I mean.

bob80905 (commented):
2 nits, the first one I think is worth addressing before merging in.
Agreed with Finn about the unprecedented diagnostic testing for the offloader.

test/Tools/Offloader/invalid-sample-count.test
Icohedron: I don't see why more testing would be bad. There were already some existing offloader error message tests, so I may as well add more.

test/Graphics/MSAA/SimpleTriangle.test
Icohedron: This is intentional because the non-MSAA SimpleTriangle test also does not have an issue link. So this was simply copied over.

test/Tools/Offloader/invalid-sample-count.test
Icohedron: Some existing offloader error message tests added by others:
- https://github.com/llvm/offload-test-suite/blob/main/test/Tools/Offloader/missing-counter-binding.test
- https://github.com/llvm/offload-test-suite/blob/main/test/Tools/Offloader/array-in-root-constant.test

inbelic: I won't block on this and we can remove it in the future if we decide on not wanting them

inbelic (approved):

lib/API/MTL/MTLDevice.cpp
Icohedron: `%s` is `SampleCount`, so your error message would read: "SampleCount with sample count N is not supported on this Metal device"

Icohedron: But yes, I think the `%s` can be replaced with a hard-coded `SampleCount` string since that is the only field applicable for this error message.

bob80905: Ah, I had thought it would be the resource name, mb.

bob80905 (approved):

lib/API/VK/Device.cpp
Icohedron: If this is done, then it should be in a separate PR that edits all the places where DX adds resource transitions.

Icohedron: But I also think DX's version of this function is more easy to read and don't require the comments.

PR #1526, opened 2026-09-18
[DirectX][Metal][Vulkan] Add support for multiple viewports and SV_ViewportArrayIndex

### LINKED ISSUE #1051: [SPIRV][DirectX] Add multiple viewports support
Required for testing of `SV_ViewportArrayIndex` written from a vertex shader. A vertex shader may select a viewport per triangle from a vertex attribute.

Current Status: Both backends hardcode a single viewport derived from the render target dimensions, and `Encoder::setViewport`/`setScissor` take a single value.

Vulkan already queries `multiViewport` and `shaderOutputViewportIndex` but enables neither.

AC:
- [ ] Allow YAML test configuration to define multiple viewports (and matching scissor rects) under `Bindings`
- [ ] API: extend `Encoder::setViewports`/`setScissors` to take a list
- [ ] DX: issue `RSSetViewports`/`RSSetScissorRects` with the configured count
- [ ] DX: cabapility check of `VPAndRTArrayIndexFromAnyShaderFeedingRasterizerSupportedWithoutGSEmulation`
- [ ] VK: enable `multiViewport` and `shaderOutputViewportIndex` device features
- [ ] VK: issue `vkCmdSetViewport`/`vkCmdSetScissor` with the configured count
- [ ] VK: update `VkPipelineViewportStateCreateInfo` with viewport/scissor count
- [ ] Add lit feature for multi-viewport capability detection, set per backend as support lands
- [ ] Add testing by checking computed position of pixel shader through a UAV record

Icohedron:
Fixes https://github.com/llvm/offload-test-suite/issues/1051

This PR adds support for multiple viewports (and scissor rectangles) to the DirectX, Metal, and Vulkan backends, as well as `SV_ViewportArrayIndex`.

Assisted by: GPT 5.6 Sol


docs/Viewports.md
```diff
@@ -0,0 +1,124 @@
...
+
+This divides an 800-by-600 render target into two side-by-side viewports.
+Viewport 0 uses scissor 0, viewport 1 uses scissor 1, and so on.
+
+`Viewports` and `Scissors` must be specified together and contain the same
+number of entries. Up to 16 entries are supported.
+
+Viewport fields:
+
+| Field | Required | Default | Constraint |
```
inbelic: nit: align tables? not sure if that is standard in the repo or not

include/API/Device.h
```diff
@@ -111,6 +111,7 @@ struct TraditionalRasterPipelineCreateDesc {
   // Set if Topology == PatchList. Validated in
   // Pipeline.cpp::validatePipelineKind.
   std::optional<uint32_t> PatchControlPoints;
+  uint32_t ViewportCount = 1;
```
inbelic: worth documenting the invariant: ViewportCount must equal the number of viewports/scissors later passed to setViewports/setScissors, because Vulkan bakes it into the pipeline while DX/Metal take it purely from the encoder call. right now the two are only kept in sync because both derive from P.Bindings.

include/API/Encoder.h
```diff
@@ -129,23 +130,15 @@ class ComputeEncoder : public CommandEncoder {
...
-};
-
 class RenderEncoder : public CommandEncoder {
 public:
   using CommandEncoder::CommandEncoder;
 
-  virtual void setViewport(const Viewport &VP) = 0;
-  virtual void setScissor(const ScissorRect &Rect) = 0;
+  /// Bind viewports to consecutive slots starting at zero.
+  virtual void setViewports(llvm::ArrayRef<Viewport> Viewports) = 0;
```
inbelic: nothing at this layer enforces that setViewports and setScissors get the same count, yet Vulkan requires both to match the pipeline's viewportCount. consider documenting that here (or taking both arrays in one call) so backend implementors don't have to infer it.

include/API/Viewport.h
```diff
@@ -0,0 +1,32 @@
...
+//===----------------------------------------------------------------------===//
+
+#ifndef OFFLOADTEST_API_VIEWPORT_H
+#define OFFLOADTEST_API_VIEWPORT_H
+
+#include <cstdint>
+
+namespace offloadtest {
+
+// D3D12's limit; Vulkan multiViewport and Metal 2 support at least 16.
```
inbelic: Recheck: this comment already qualifies the Vulkan guarantee with multiViewport, so my earlier wording nit was unnecessary. No change requested here; the separate runtime device-capability concern is in VK/Device.cpp.

lib/API/DX/Device.cpp
```diff
@@ -1137,23 +1137,31 @@ class DXRenderEncoder : public offloadtest::RenderEncoder {
...
-  void setViewport(const offloadtest::Viewport &VP) override {
-    D3D12_VIEWPORT DXVP = {};
-    DXVP.TopLeftX = VP.X;
-    DXVP.TopLeftY = VP.Y;
-    DXVP.Width = VP.Width;
-    DXVP.Height = VP.Height;
-    DXVP.MinDepth = VP.MinDepth;
-    DXVP.MaxDepth = VP.MaxDepth;
-    CB.CmdList->RSSetViewports(1, &DXVP);
+  void setViewports(llvm::ArrayRef<offloadtest::Viewport> VPs) override {
```
inbelic: Recheck: YAML parsing is only one caller of the public encoder API, so retaining these backend precondition assertions is reasonable. I withdraw the earlier suggestion to remove them merely because the YAML path validates the arrays. No change required.

lib/API/DX/Device.cpp
```diff
@@ -3173,6 +3172,7 @@ class DXDevice : public offloadtest::Device {
         PipelineDesc.Topology = P.Bindings.Topology;
         PipelineDesc.PatchControlPoints = P.Bindings.PatchControlPoints;
         PipelineDesc.DSFormat = Format::D32FloatS8Uint;
+        PipelineDesc.ViewportCount = P.Bindings.getViewportCount();
```
inbelic: DX never reads Desc.ViewportCount (viewport count is dynamic state on RSSetViewports), same for Metal. these four assignments are dead stores today. either drop them and document in Device.h that ViewportCount is Vulkan-only, or leave them but add that note so nobody assumes DX honours it.

lib/API/VK/Device.cpp
```diff
@@ -2400,6 +2404,9 @@ class VulkanDevice : public offloadtest::Device {
   createTraditionalRasterPipeline(
       llvm::StringRef Name, const BindingsDesc &BindingsDesc,
       const TraditionalRasterPipelineCreateDesc &Desc) override {
+    assert(Desc.ViewportCount > 0 &&
```
inbelic: Concern: this assertion only checks the framework's 1..16 limit, not whether this device supports multiViewport, so a two-viewport YAML reaches vkCreateGraphicsPipelines on an unsupported device in both debug and release builds. The lit feature protects tagged tests but not direct offloader invocations. Could both traditional and mesh pipeline creation return an llvm::Error when ViewportCount > 1 without multiViewport, or when the count exceeds the device's maxViewports? Correction to my earlier wording: replacing the assert alone is not enough; the device-capability check is missing.

lib/Support/Pipeline.cpp
```diff
@@ -106,6 +106,28 @@ uint32_t PushConstantBlock::size() const {
...
+llvm::SmallVector<ScissorRect> IOBindings::getScissors() const {
+  if (!Scissors.empty())
+    return Scissors;
+
+  assert(RTargetBufferPtr && "Raster pipeline has no render target");
+  ScissorRect Rect;
+  Rect.Width = static_cast<uint32_t>(RTargetBufferPtr->OutputProps.Width);
+  Rect.Height = static_cast<uint32_t>(RTargetBufferPtr->OutputProps.Height);
+  return {Rect};
+}
```
inbelic: Concern, narrowed to Metal scissors: explicit rectangles can extend beyond the render attachment, while Metal's setScissorRect contract requires them to lie completely inside it. Please either diagnose this for Metal or clamp to the attachment before binding. Do not impose render-target bounds on viewports generally: oversized/offset viewports are legitimate. Correction to the earlier comment: RTargetBufferPtr is resolved after MappingTraits<IOBindings>::mapping, so attachment-dependent validation belongs after resolution in the outer pipeline mapping or in the backend, not in the bindings mapping. The earlier D3D12 claim was not established.

lib/Support/Pipeline.cpp
```diff
@@ -569,6 +591,55 @@ void MappingTraits<offloadtest::IOBindings>::mapping(
...
+               " Scissors may be specified, found " +
+               std::to_string(B.Scissors.size()) + ".");
+  if (B.Viewports.empty() != B.Scissors.empty())
+    I.setError(
+        "Bindings: 'Viewports' and 'Scissors' must be specified together.");
+  else if (B.Viewports.size() != B.Scissors.size())
+    I.setError(Twine("Bindings: 'Scissors' has ") +
+               std::to_string(B.Scissors.size()) + " entries but there are " +
+               std::to_string(B.Viewports.size()) +
+               " viewports; they must match.");
```
inbelic: simplification: the empty()!=empty() case is subsumed by the size comparison, so this can be one check. e.g. `if (B.Viewports.size() != B.Scissors.size()) I.setError(..."'Viewports' and 'Scissors' must be specified together and have the same number of entries (got N vs M)")`. one branch, one message, same coverage.

lib/Support/Pipeline.cpp
```diff
@@ -569,6 +591,55 @@ void MappingTraits<offloadtest::IOBindings>::mapping(
...
+  I.mapRequired("Width", V.Width);
+  I.mapRequired("Height", V.Height);
+  I.mapOptional("MinDepth", V.MinDepth, 0.0f);
+  I.mapOptional("MaxDepth", V.MaxDepth, 1.0f);
+
+  if (V.Width <= 0 || V.Height <= 0)
+    I.setError("Viewport width and height must be positive.");
+  if (V.MinDepth < 0 || V.MaxDepth > 1 || V.MinDepth > V.MaxDepth)
+    I.setError(
+        "Viewport depth range must satisfy 0 <= MinDepth <= MaxDepth <= 1.");
```
inbelic, line 634: nit: copilot pointed out that you could technically specify `nan` or `inf` as a width or height and it would be parsed and reach this check. Maybe worth adding a `std::isfinite` check?

test/lit.cfg.py
```diff
@@ -219,6 +225,7 @@ def setDeviceFeatures(config, device, compiler):
         config.available_features.add("Int16")
         config.available_features.add("Int64")
         config.available_features.add("Half")
+        config.available_features.add("multi-viewport")
```
inbelic: Recheck: the docs explicitly make supported Metal targets the basis for advertising this feature, and I have not identified a supported target violating that assumption. The earlier unconditional-gating comment is therefore not an established bug. A capability check would only be needed if the supported hardware set includes devices without multiple viewports; please do not add a Mac2-only restriction just for symmetry with DX/Vulkan.

test/Graphics/MultipleViewports.test
```diff
@@ -0,0 +1,134 @@
+# Two full-screen quads select side-by-side viewports. The pixel shader records
+# the viewport index and transformed SV_POSITION for every covered pixel.
+
+#--- vertex.hlsl
+struct VSOutput {
+    float4 position : SV_POSITION;
+    uint viewport : SV_ViewportArrayIndex;
+    nointerpolation uint vpIndex : VPINDEX;
```
inbelic: Is it possible for the `PSInput` to read the index from `SV_ViewPortArrayIndex` rather than this user semantic? I would prefer that if possible to test both semantics

test/Graphics/MultipleViewports.test
```diff
@@ -0,0 +1,134 @@
...
+      Width: 4
+      Height: 1
+  Scissors:
+    - X: 0
+      Y: 0
+      Width: 4
+      Height: 1
+    - X: 4
+      Y: 0
+      Width: 4
```
inbelic, line 103: Should the scissor rect be a subset of the viewport rect? Would this testcase catch the case when the scissor rect is ignored or incorrect? Although maybe this is addressed anyway since both are written to the out buffer?

lib/API/DX/Device.cpp
```diff
@@ -3006,17 +3014,8 @@ class DXDevice : public offloadtest::Device {
...
-    VP.Height =
-        static_cast<float>(P.Bindings.RTargetBufferPtr->OutputProps.Height);
-    Encoder.setViewport(VP);
-
-    ScissorRect Scissor;
-    Scissor.Width = static_cast<uint32_t>(VP.Width);
-    Scissor.Height = static_cast<uint32_t>(VP.Height);
-    Encoder.setScissor(Scissor);
+    Encoder.setViewports(P.Bindings.getViewports());
+    Encoder.setScissors(P.Bindings.getScissors());
```
inbelic: Should there be an assertion that the viewportcount is equivalent to the bindings sizes?

lib/Support/Pipeline.cpp
inbelic: nit: It seems like this could just be the second check and error message

inbelic (commented):
Mostly LGTM, just some small requests for change

test/Graphics/MultipleViewports.test
Icohedron: It is not a requirement that a scissor rect is contained by or is a subset of a viewport.

inbelic (approved):

docs/Viewports.md
```diff
@@ -0,0 +1,127 @@
...
+| `MinDepth` | No       | `0`     | Finite and between 0 and 1          |
+| `MaxDepth` | No       | `1`     | Finite and between `MinDepth` and 1 |
+
+Scissor fields:
+
+| Field    | Required | Default | Constraint        |
+|----------|----------|---------|-------------------|
+| `X`      | No       | `0`     | Non-negative      |
+| `Y`      | No       | `0`     | Non-negative      |
+| `Width`  | Yes      | —       | Greater than zero |
```
bob80905: nit: can these be infinite?

include/API/Encoder.h
bob80905: nit: should the function parameter be set to const?

lib/API/DX/Device.cpp
```diff
@@ -1137,23 +1137,31 @@ class DXRenderEncoder : public offloadtest::RenderEncoder {
...
-    DXVP.Height = VP.Height;
-    DXVP.MinDepth = VP.MinDepth;
-    DXVP.MaxDepth = VP.MaxDepth;
-    CB.CmdList->RSSetViewports(1, &DXVP);
+  void setViewports(llvm::ArrayRef<offloadtest::Viewport> VPs) override {
+    assert(!VPs.empty() && "At least one viewport is required.");
+    assert(VPs.size() <= offloadtest::MaxViewports &&
+           "Viewport count exceeds D3D12's per-pipeline limit.");
+    llvm::SmallVector<D3D12_VIEWPORT, offloadtest::MaxViewports> DXVPs;
+    DXVPs.reserve(VPs.size());
```
bob80905: nit: I believe both `reserve`s are dead code, since small vector is set with capacity MaxViewports.
What about this definition?
`llvm::SmallVector<D3D12_VIEWPORT, VPs.size()> DXVPs;`

Same applies with the setScissors function.

lib/API/MTL/MTLDevice.cpp
```diff
@@ -3071,6 +3084,7 @@ class MTLDevice : public offloadtest::Device {
         Device->supportsFamily(MTL::GPUFamilyMetal3);
     Caps.insert(std::make_pair(
         "MeshShader", makeCapability<bool>("MeshShader", MeshShaderSupported)));
+
```
bob80905: nit: is this new line a clang-format artifact?

lib/Support/Pipeline.cpp
```diff
@@ -569,6 +593,58 @@ void MappingTraits<offloadtest::IOBindings>::mapping(
...
+  I.mapOptional("MaxDepth", V.MaxDepth, 1.0f);
+
+  if (!std::isfinite(V.X) || !std::isfinite(V.Y) || !std::isfinite(V.Width) ||
+      !std::isfinite(V.Height) || !std::isfinite(V.MinDepth) ||
+      !std::isfinite(V.MaxDepth)) {
+    I.setError("Viewport values must be finite.");
+    return;
+  }
+  if (V.Width <= 0 || V.Height <= 0)
+    I.setError("Viewport width and height must be positive.");
```
bob80905: question, I noticed the -Height thing in Vulkan, is it semantically sound for Vulkan to specify a negative height?

test/Graphics/MultipleViewports.test
```diff
@@ -0,0 +1,132 @@
...
+Shaders:
+  - Stage: Vertex
+    Entry: main
+  - Stage: Pixel
+    Entry: main
+Buffers:
+  - Name: RenderTarget
+    Format: Float32
+    Channels: 4
+    FillSize: 128 # 8x1 @ 16 bytes per pixel
```
bob80905: nit: with 2 quads wouldn't this be 4x2?

